PR #3, opened 2026-02-24
Add Identity option 3, where agent can create a wallet by himself

danielmbirochi:


0xsarvesh (approved):

agent-onboarding/SKILL.md
```diff
@@ -45,25 +47,133 @@ Follow every step in order. Do not skip steps.
...
+  # See Error Handling section below
+  exit 1
+fi
+```
+
+Persist the wallet immediately (before continuing onboarding) — this key controls funds and must not be lost:
+
+```bash
+mkdir -p ~/.config/senpi
+cat > ~/.config/senpi/wallet.json << EOF
```
garg-prashant, line 141: little worried about this one because some other plugin on users openclaw setup might be able to read this. May be not for this release, but we can instruct user to clear this file once they have backed up

garg-prashant (commented):
mostly looks good.

PR #47, opened 2026-03-06
adding skill attribution

Hrishikesh-Thakkar:
> [!NOTE]
> **Medium Risk**
> Updates the shared `call_mcp()` wrapper used by all LION scripts to always pass `skill_name`/`skill_version`, which could affect every Senpi MCP invocation if those parameters are unsupported or conflict with existing args. Otherwise the change is small and mostly documentation.
> 
> **Overview**
> Adds *skill attribution* to the LION strategy by documenting required `skill_name`/`skill_version` fields and introducing a new `references/skill-attribution.md` reference.
> 
> Updates `lion_config.call_mcp()` to automatically inject `skill_name="lion-strategy"` and `skill_version="1.0"` into every Senpi MCP tool call (unless already provided), making all executions traceable without changing individual call sites.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 54caf9b5caecc23e8f79c57494a00666053f654a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lion/SKILL.md
+++ b/lion/SKILL.md
@@ -26,6 +26,33 @@ Wait for the market to break. Trade the repair.
 
 ---
 
+## Skill Attribution
+
+**Skill ID:** `lion-strategy`
+
+When calling `strategy_create` or `strategy_create_custom_strategy`, always include:
+
+```json
+"skill_name": "lion-strategy",
+"skill_version": "1.0"
+```
+
+This is required for attribution and tracking. Example:
+
+```json
+{
+  "tool": "strategy_create_custom_strategy",
+  "args": {
+    "initialBudget": 5000,
+    "positions": [],
+    "skill_name": "lion-strategy",
+    "skill_version": "1.0"
+  }
+}
+```
+
+---
+
 ## Architecture
 
 ```
--- a/lion/SKILL.md
+++ b/lion/SKILL.md
@@ -28,28 +28,7 @@ Wait for the market to break. Trade the repair.
 
 ## Skill Attribution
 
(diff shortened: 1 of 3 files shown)


lion/SKILL.md
```diff
@@ -26,6 +26,33 @@ Wait for the market to break. Trade the repair.
 
 ---
 
+## Skill Attribution
+
+**Skill ID:** `lion-strategy`
```
danielmbirochi: Why do we need this ? I don't see its being used in other places, also it seems to be the same value of the skill name in the frontmatter.

lion/SKILL.md
```diff
@@ -26,6 +26,33 @@ Wait for the market to break. Trade the repair.
...
+{
+  "tool": "strategy_create_custom_strategy",
+  "args": {
+    "initialBudget": 5000,
+    "positions": [],
+    "skill_name": "lion-strategy",
+    "skill_version": "1.0"
+  }
+}
+```
```
danielmbirochi: What is inside the `SKILL.md` file is passed as context to LLM in every call, I'm not sure if we do need this here, because the llm would choose to create a strategy only in a few subset of interactions w/ the skill - if that's the case, then this should be moved to a reference file, then you can maintain in the `SKILL.md` only something like: `for creating a strategy, check details in <point to respective reference file>

danielmbirochi (changes requested):

danielmbirochi (approved):

PR #55, opened 2026-03-10
Replacing deprecated gemini models 2.0 with 2.5

bhaskarhocane:
It is breaking the wolf cron.
image


betashop:
I'm not sure this is the right approach. Models should not be coded into skills. Skills should be able to adjust to the appropriate available models.

PR #208, opened 2026-05-08
Add senpi_runtime_helpers Python wrapper + migrate Pangolin to it

rachinkapoor:
## Summary

- Adds `_helpers/senpi_runtime_helpers/` — a stdlib-only Python wrapper that gives senpi-skills producers a persistent HTTPS client to MCP, a direct POST path to the runtime's `/signals` endpoint, a stale-PID-recovering `scanner_lock`, bounded `parallel(...)` fan-out, a per-tick TTL cache, and a `producer_daemon` scheduler that replaces openclaw cron + `agentTurn`.
- Migrates Pangolin to the wrapper (the only producer migrated in this PR — others stay on the legacy `mcporter`/`external-scanner ingest` path until their own migrations).
- Adds a wrapper-discoverable `SKILL.md` and reference docs so agents authoring future skills find the right path.

## Why

Per `senpi-trading-runtime/docs/runtime-v2-fixes/runtime-2-performance-findings.md`, the legacy producer architecture was paying for:

- 5–8 s CLI cold-start per `openclaw senpi external-scanner ingest`.
- 6-process `mcp-remote` spawn tree (250–300 MB transient RSS, 2.5–5 s per MCP call).
- Cron + `agentTurn` LLM tax on every tick — 740 LLM agent runs in 3.5 days just to dispatch a Python script.
- Fork-storm pressure under overlapping crons (kernel `EAGAIN` from `fork()`).

The wrapper closes all four. On the wrapper-test deploy:

| Metric | Legacy | Wrapper |
|---|---|---|
| MCP call latency (median) | 2.5–5 s | 273 ms |
| Producer tick wall-clock | 30–60 s | ~4 s |
| `mcp_initialized` events / process | once per call | 1 lifetime |
| Plugin re-registrations / 3.5 days | 605 | ~12 |
| Zombie/h from cron pipeline | 2.4 | 0 (single one-shot startup zombie) |

## What's in this PR

| Path | What |
|---|---|
| `_helpers/senpi_runtime_helpers/` | Wrapper package: `client.py` (MCP + `/signals` + `/state` probes), `daemon.py` (scheduler + boot probe + periodic `alive_check`), `lock.py` (PID-aliveness recovery), `cache.py` (LRU+TTL+thundering-herd coalesce), `parallel.py` (bounded fan-out, hard `tick_timeout` enforcement), `_config.py`, `_logging.py`. 50 stdlib `unittest` tests. |
| `_helpers/senpi_runtime_helpers/SKILL.md` | Agent-facing skill: decision tree, recipes, errors→fixes table, footgun callout (`asset`/`direction` go top-level, not in `data`). |
| `_helpers/senpi_runtime_helpers/references/` | `architecture.md` (rationale + measured impact), `migration-cookbook.md` (legacy → wrapper step-by-step), `runtime-deployment.md` (3-step Railway recipe), `signal-schema.md` (full `SignalItem` reference). |
| `GUIDE.md` | New Section 3.0 directs new producer skills at the wrapper. Sections 3.1–3.7 stay published as a reference for un-migrated skills. |
| `senpi-trading-runtime/SKILL.md` | Companion-skill callout pointing producers at the wrapper. |
| `pangolin/scripts/pangolin-producer.py` | Migrated to `producer_daemon` + `client.push_signal`. Keeps Pangolin's funding-fade thesis verbatim. |
| `pangolin/runtime.yaml` | `external_scanner.config.fields` cleaned up so `asset`/`direction` are top-level routing only (matches the runtime's `SignalItem` schema). |

## Compatibility

Requires `senpi-trading-runtime >= 2.0.0`. The wrapper does not parse the legacy `{ results: [...] }` envelope from 1.x, and the daemon's default `alive_check` requires the new `GET /state` endpoint (2.x only) to detect runtime-delete and scanner-rename. Currently verified against `@senpi/runtime@2.0.0-dev.runtime-phase-2.20260508074726` — the dev tag pin is in `_helpers/senpi_runtime_helpers/references/runtime-deployment.md` and updates as new dev tags publish.

## Test plan

- [ ] CI: `npm test` across the 50-test wrapper suite (stdlib `unittest`, no creds).
- [ ] Manual: deploy the matched pair (`@senpi/runtime@2.0.0-dev.runtime-phase-2.<TIMESTAMP>` + this branch) on Railway. Confirm:
  - [ ] `[senpi_helpers] event=daemon_started, alive_check_enabled=true` at boot.
  - [ ] `[senpi_helpers] event=mcp_initialized` exactly once per process lifetime.
  - [ ] Per-tick MCP latency 250–500 ms typical.
  - [ ] `daemon_aborted_no_runtime` fires when an unregistered wallet is passed (boot probe self-terminate).
  - [ ] `is_scanner_registered` returns False if the scanner is dropped from the runtime config; daemon self-terminates within ~5 min.
- [ ] Pangolin tick on the wrapper-test box reaches `daemon_tick_finished status=ok` end-to-end against the 2.0 runtime.

## Out of scope

- Migrating the other ~89 producers. Each is its own follow-up PR; `migration-cookbook.md` documents the per-skill recipe.
- Scanner-level liveness check requires runtime 2.0 (already shipped on `runtime-phase-2`). Pre-2.0 deploys will fail the `alive_check` boot probe — opt out via `alive_check=None` if running against an older runtime is intentional.
- Per-skill telemetry counters (`runtime.signals_ingested`) — TODO'd on the runtime side under the S11 telemetry work.

---

> [!NOTE]
> **Medium Risk**
> Adds a sizable new stdlib-only producer runtime wrapper (direct MCP + `/signals` HTTP, daemon scheduler, locking/cache/parallelism) plus an operator CLI; risk is mainly correctness/regression when producers migrate to this new execution and ingestion path.
> 
> **Overview**
> Introduces `senpi_runtime_helpers`, a new stdlib-only Python wrapper for producers that replaces the legacy `mcporter`/OpenClaw CLI subprocess approach with a persistent HTTPS MCP client, direct `/signals` posting, per-tick caching, bounded parallel fan-out, and a long-lived `producer_daemon` scheduler with runtime/scanner liveness checks.
> 
> Adds an operator-facing `senpi-helpers` CLI (list/health/stats/stop/restart) backed by daemon state files (`pid.json`/`boot.json`/`heartbeat.json`) and log aggregation, and updates `GUIDE.md` to direct new skills to the wrapper while keeping legacy docs as reference.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a58588582c1afee1efec58d0e06e9e9b36b1baf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


_helpers/senpi_runtime_helpers/cache.py
```diff
@@ -0,0 +1,238 @@
...
+
+    with cache.lock:
+        cache.store[key] = _Entry(time.time(), value)
+        cache.store.move_to_end(key)
+        while len(cache.store) > cfg.TICK_CACHE_MAX_ENTRIES:
+            cache.store.popitem(last=False)
+            cache.evictions += 1
+        cache.misses += 1
+        cache.pending.pop(key, None)
+    pending.set()
```
danielmbirochi: **Cache coalesce has a wakeup race that defeats the point of the cache.**

Owner inserts the entry under `cache.lock` (lines 168–174) and then calls `pending.set()` *outside* the lock (line 176). A waiter that returns from `pending.wait()` between those two steps re-acquires `cache.lock`, finds no entry (it isn't published yet), falls through to `client.mcp_call(...)` at line 153, and re-issues the call. The "last writer wins for the cache entry" comment papers over the race, but the wrapper's whole purpose is to collapse N concurrent callers into one MCP call.

Fix: set the event while still holding `cache.lock` (move line 176 inside the `with cache.lock:` block at line 168), or have waiters re-check under lock with a brief retry.

`test_thundering_herd_coalesced_to_one_call` only covers the happy path with a sleep — this race never surfaces in CI.

_helpers/senpi_runtime_helpers/client.py
```diff
@@ -0,0 +1,784 @@
...
+        conn = self._pool.get(scheme, host, port, timeout)
+        try:
+            conn.request("GET", path, headers={"Host": parts.netloc, "Accept": "application/json"})
+            resp = conn.getresponse()
+        except (http.client.HTTPException, ConnectionError, OSError, TimeoutError) as e:
+            self._pool.reset(scheme, host, port)
+            raise SenpiClientError(f"state probe transport error: {e}") from e
+
+        try:
+            raw = resp.read()
```
danielmbirochi: **`_fetch_state` never closes the `HTTPResponse`, leaking its keep-alive socket.**

`raw = resp.read()` (line 430) and the non-200 raise (line 435) never call `resp.close()` — same on the success path. Compare to `_post_json`, which uses `with` semantics so the response is closed deterministically. Under fan-out and frequent `/state` probes from `alive_check`, this leans on GC for connection cleanup.

Fix: wrap the response in `try/finally: resp.close()` or restructure to `with resp:`.

_helpers/senpi_runtime_helpers/daemon.py
```diff
@@ -0,0 +1,313 @@
...
+                    after_tick=tick_count,
+                    reason="alive_check returned False between ticks",
+                )
+                break
+
+            tick_count += 1
+            tick_started_at = time.time()
+            log_event("daemon_tick_started", name=name, tick=tick_count)
+
+            armed = _arm_tick_alarm(tick_timeout)
```
danielmbirochi: **SIGALRM is armed before `scanner_lock` enters and disarmed after it exits — the lock ceremony is inside the timeout budget.**

`_arm_tick_alarm(tick_timeout)` (line 262) runs before `with scanner_lock(...)` (line 266), and `_disarm_tick_alarm()` (line 279) runs in the outer `finally`. The alarm can fire during `scanner_lock.__enter__` (inside `_write_metadata_inplace` in `lock.py`) and during `__exit__` while it's releasing the flock and closing the fd. The producer's `fn()` body is what should be timed, not the lock acquisition/release.

Fix: arm inside the `with scanner_lock(...)` block (right before `fn()`), disarm right after.

_helpers/senpi_runtime_helpers/SKILL.md
```diff
@@ -0,0 +1,390 @@
...
+| `signal_post: HTTP 400 INVALID_REQUEST: Exceeded api.maxItemsPerSignalsRequest=10` | Batch larger than runtime cap | Split the batch (default cap is 10) |
+| `signal_post: HTTP 4xx/5xx ENVELOPE_CODE: …` | Runtime rejected the envelope itself | Read the human message in the exception — it's the runtime's diagnostic |
+| `MCP error: …` from `mcp_call` | MCP server reported a tool-side error | Check tool name + arguments against `senpi-hyperliquid-mcp` schema |
+| `urllib.error.URLError` from `mcp_call` | Network / connect / TLS failure | Verify `SENPI_MCP_URL`, `SENPI_AUTH_TOKEN`. Retry transient errors at next tick |
+| `socket.timeout` from `mcp_call` or `signal_post` | Wall-clock timeout exceeded | Override per-call: `client.mcp_call(tool, timeout=60.0, …)` |
+| `lock_stale_recovered` (info, not error) | Previous holder crashed; lock auto-recovered | No action; this is the intended self-healing behaviour |
```
danielmbirochi: **`lock_stale_recovered` is documented but never emitted by the code.**

This errors→fixes table references `lock_stale_recovered`; `references/architecture.md:129` makes the same claim ("logs `lock_stale_recovered` and proceeds"). The actual emission in `lock.py:157` is `lock_recovered_after_crash`. Operators grepping logs for the documented event find nothing.

Pick one and align the other (rename the code event or rename the doc references).

_helpers/senpi_runtime_helpers/SKILL.md
```diff
@@ -0,0 +1,390 @@
...
+daemon (tick lifecycle + signal handling).
+
+---
+
+## See also
+
+- [`references/migration-cookbook.md`](references/migration-cookbook.md) — long-form migration with before/after snippets.
+- [`references/architecture.md`](references/architecture.md) — why this exists, performance numbers, incident background.
+- [`references/signal-schema.md`](references/signal-schema.md) — full `SignalItem` shape, validation rules, `data` block conventions.
+- [`pangolin/scripts/pangolin-producer.py`](../../pangolin/scripts/pangolin-producer.py) — reference wrapper-based producer.
```
danielmbirochi, line 485: **Test count is stale.**

Docs say "35 tests" here and in `_helpers/senpi_runtime_helpers/README.md:34`. `tests/test_*.py` defines 51 `def test_*`. PR description says 50. Pick the right number and update both docs.

_helpers/senpi_runtime_helpers/references/signal-schema.md
```diff
@@ -0,0 +1,198 @@
...
+
+Full description of the `SignalItem` shape consumed by
+`senpi-trading-runtime`'s `POST /signals` endpoint, as exposed by
+`SenpiClient.push_signal(...)` and `push_signals([...])`.
+
+For usage recipes, read [`../SKILL.md`](../SKILL.md). This doc is the
+authoritative field reference.
+
+Source of truth on the runtime side:
+[`senpi-trading-runtime/src/runtime-api/routes/signals.schema.ts`](../../../../senpi-trading-runtime/src/runtime-api/routes/signals.schema.ts)
```
danielmbirochi: **Cross-repo doc links don't resolve in `senpi-skills`.**

This line and `_helpers/senpi_runtime_helpers/client.py:739` link to `senpi-trading-runtime/src/runtime-api/routes/signals.schema.ts` via `../../../../`. `_helpers/senpi_runtime_helpers/references/architecture.md:22,76` and `GUIDE.md:217` similarly link to `senpi-trading-runtime/docs/runtime-v2-fixes/runtime-2-performance-findings.md` and `wrapper-plan.md`. Those paths live in the runtime repo, not here — `senpi-trading-runtime/` in this repo only contains `SKILL.md` + `references/`.

Fix: qualify as "in the senpi-trading-runtime repo" or replace with the published doc URL.

danielmbirochi (commented):
Some comments on wrapper correctness and doc/code drift. None are blocking but all are concrete fixes before this carries the fleet.

_helpers/senpi_runtime_helpers/SKILL.md
rachinkapoor: Both docs say 54 now (matches the 54 actual tests across 5 files). Fixed in 5d0552b.

_helpers/senpi_runtime_helpers/daemon.py
```diff
@@ -0,0 +1,318 @@
...
+    # wallet; uses scanner when provided); explicit None → opt out;
+    # callable → use as-is.
+    if alive_check is _DEFAULT_ALIVE_CHECK:
+        if wallet is None:
+            raise ValueError(
+                "producer_daemon requires `wallet=...` for the default runtime "
+                "liveness check. Pass `wallet=<your strategy wallet>` to enable "
+                "auto-self-termination on runtime delete, or `alive_check=None` "
+                "to opt out."
+            )
```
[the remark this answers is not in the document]
rachinkapoor: Good catch. Daemon now rejects empty / non-0x wallets up front with a clear ValueError, instead of bypassing the None-check and silently turning every alive probe into a swallowed exception. Test added. Fixed in 2366efa.

_helpers/senpi_runtime_helpers/cache.py
```diff
@@ -0,0 +1,255 @@
...
+    consistent because `_Cache.__init__` allocates everything before the
+    attribute is bound."""
+    cache = getattr(client, _CLIENT_CACHE_ATTR, None)
+    if isinstance(cache, _Cache):
+        return cache
+    new_cache = _Cache()
+    # If client._cache was previously set to a plain dict (legacy SenpiClient
+    # __init__ path before the upgrade), replace it with the typed cache.
+    setattr(client, _CLIENT_CACHE_ATTR, new_cache)
+    return new_cache
```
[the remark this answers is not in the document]
rachinkapoor: Fixed by initializing `_cache` directly as a `_Cache()` in `SenpiClient.__init__`. The dict-to-_Cache transition is gone, so the race window is gone. Also dropped the misleading 'thread-safe via the GIL' comment — the GIL does not make check-then-act atomic. Fixed in 8fa7f32.

_helpers/senpi_runtime_helpers/client.py
```diff
@@ -0,0 +1,788 @@
...
+        # Drain so the connection can be reused for the next request.
+        body_bytes = resp.read()
+        # If the server signalled connection-close, drop our reference.
+        if resp.getheader("Connection", "").lower() == "close":
+            pool.reset(scheme, host, port)
+        raise urllib.error.HTTPError(
+            url, resp.status, resp.reason, dict(resp.getheaders()), io.BytesIO(body_bytes)
+        )
+
+    return resp
```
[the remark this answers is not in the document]
rachinkapoor: Fixed. Converted `_post_json` to a `@contextmanager` so `Connection: close` is honored in `finally` on every exit path — success, 4xx/5xx, and exceptions inside the `with` block. Mirrors what `_fetch_state` already does on the GET side. Caller `with _post_json(...) as resp:` syntax is unchanged. Test added. Fixed in 2214e95.

_helpers/senpi_runtime_helpers/PR-208-COMMENT-RESPONSES.md
```diff
@@ -0,0 +1,37 @@
...
+appreciates above ~$168, which is the bracket the test deployment is
+intentionally exercising.
+
+The fleet baseline will revert to `1000.0` (or be made wallet-relative) when
+the Pangolin producer migrates beyond the test deployment. That work happens
+in a follow-up PR alongside the next round of fleet tuning; treating it as a
+gate on this PR would block the wrapper migration on unrelated tuning work.
+
+The bugbot comment is left **unresolved** in GitHub (rather than resolved) so
+the issue remains visible until the post-test rebase lands.
```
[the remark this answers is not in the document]
rachinkapoor: Removed. The rationale (rebase `STARTING_BUDGET` back to 1000.0 in a follow-up PR after the test deployment graduates) stays in this bugbot thread, which is the canonical record. Fixed in 19dc1ed.

pangolin/scripts/pangolin_config.py
```diff
@@ -39,6 +42,51 @@
...
+    """Module-level attribute that defers SenpiClient construction until first
+    attribute access. Preserves the legacy `cfg._wrapper_client.mcp_call(...)`
+    call shape used by the producer + scanner without forcing eager
+    instantiation at import time."""
+
+    def __getattr__(self, name: str):
+        return getattr(_get_wrapper_client(), name)
+
+
+_wrapper_client = _WrapperClientProxy()
```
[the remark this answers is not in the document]
rachinkapoor: Fixed in e83d7d4. `_WrapperClientProxy` now delegates `__setattr__` AND `__delattr__` to the wrapped `SenpiClient` in addition to `__getattr__`. The hazard called out here is real: `__getattr__` is a fallback (only invoked when normal attribute lookup fails), but `__setattr__` and `__delattr__` are always called for assignment/deletion. Without delegation, a stray `cfg._wrapper_client.X = Y` would land on the proxy's own `__dict__` and a subsequent read would find it via normal lookup, bypassing the wrapped client entirely. The canonical `cache.py` setattr path the bugbot called out is already unreachable after the eager `_cache` init in 8fa7f32, but the proxy is now defensively correct for any future writer. Proxy holds no state of its own, so blanket delegation is right. Docstring expanded to record why all three dunders are required.

_helpers/senpi_runtime_helpers/references/architecture.md
```diff
@@ -0,0 +1,141 @@
+# Architecture & rationale
```
garg-prashant: Its a public repo, do you think we should keep our architecture docs here ?

_helpers/senpi_runtime_helpers/daemon.py
```diff
@@ -0,0 +1,382 @@
...
+            except Exception as e:  # noqa: BLE001 — log and keep looping
+                tick_status = "error"
+                # Bare exception type name as the code — keeps the `errors_by_code`
+                # histogram in `senpi-helpers stats` readable. The full
+                # exception message remains in `error` for debugging.
+                tick_code = type(e).__name__
+                tick_err = f"{type(e).__name__}: {e}"
+
+            if tick_status != "ok":
+                error_count += 1
```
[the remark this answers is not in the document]
rachinkapoor: Fixed in a82b40b. `skipped_locked` is now excluded from `error_count` (daemon.py), from `_compute_health` failure status (cli.py treats it as healthy), and from `ticks_errors_by_code` in the stats aggregator (stats.py — it still shows up in `ticks_by_status.skipped_locked` since operators want to see the count, just not as an error). Three tests added across test_daemon / test_cli / test_stats.

_helpers/senpi_runtime_helpers/client.py
```diff
@@ -0,0 +1,812 @@
...
+        if data is not None:
+            item["data"] = data
+        if asset is not None:
+            item["asset"] = asset
+        if direction is not None:
+            item["direction"] = direction
+        if score is not None:
+            item["score"] = score
+        if signal_type is not None:
+            item["signal_type"] = signal_type
```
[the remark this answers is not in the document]
rachinkapoor: Bugbot is mistaken on this one. The runtime schema at `senpi-trading-runtime/src/runtime-api/routes/signals.schema.ts:41,57` declares `signal_type: Type.Optional(Type.String())` — snake_case, not camelCase. The wire key the wrapper sends matches exactly. `signal-schema.md` in this PR documents `signal_type` as the canonical name, also matching. No fix needed; `additionalProperties` is not raising on this field because the field is in fact declared. Closing as not-a-bug.

_helpers/senpi_runtime_helpers/manage.py
```diff
@@ -0,0 +1,261 @@
...
+        return False
+    try:
+        os.kill(pid, 0)
+        return True
+    except ProcessLookupError:
+        return False
+    except PermissionError:
+        return True
+    except OSError:
+        return False
```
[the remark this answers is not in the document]
rachinkapoor: Fixed in 8ed58e5. Canonical `pid_alive(pid)` now lives in `state.py`. `lock.py`, `cli.py`, and `manage.py` each `from .state import pid_alive` under their existing local name (`_process_alive` / `_is_pid_alive` / `is_pid_alive`) so call sites are unchanged. Bug fixes now land once. Six tests in test_state.PidAliveTests pin the invariant — three of them use `assertIs` to verify each module's local name IS the same function object as `state.pid_alive`, so a future re-divergence fails the suite.

_helpers/senpi_runtime_helpers/references/cli-reference.md
```diff
@@ -0,0 +1,401 @@
...
+
+# Via the Python package (works anywhere senpi_runtime_helpers is importable)
+python3 -m senpi_runtime_helpers.cli <subcommand> [args]
+```
+
+The wrapper script is at `<workspace>/skills/_helpers/senpi-helpers` and is
+executable. Alias it for convenience:
+
+```bash
+alias senpi-helpers=/data/workspace/skills/_helpers/senpi-helpers
```
garg-prashant: btw, there are several paths the skill can be installed in. Here we are assuming that it would always be in the hardcoded path.

_helpers/senpi_runtime_helpers/references/architecture.md
rachinkapoor: Yes? 
Please explain your concern with doc. It explains the architecture of runtime and helper.

_helpers/senpi_runtime_helpers/references/cli-reference.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+## Global options
+
+| Flag                  | Default                                | Notes                                                                                |
+|-----------------------|----------------------------------------|--------------------------------------------------------------------------------------|
+| `--state-dir <path>`  | `$SENPI_HELPERS_STATE_DIR`             | Override the state directory for this invocation.                                    |
```
garg-prashant, line 36: Too much responsibility on LLM to figure out things. Lets see how this goes.

rachinkapoor: TBH, it's for me and you to be able to help debug users

_helpers/senpi_runtime_helpers/references/cli-reference.md
rachinkapoor: Good catch, thanks. Fixed in e2a9305. The doc now uses `${OPENCLAW_WORKSPACE:-/data/workspace}/skills/_helpers/senpi-helpers` so it adapts to any install path. The `python3 -m senpi_runtime_helpers.cli` form is already install-path agnostic.

_helpers/senpi_runtime_helpers/references/runtime-deployment.md
```diff
@@ -0,0 +1,169 @@
...
+
+Open https://railway.app/template/openclaw and deploy. The template provisions an OpenClaw gateway service with the right Dockerfile + a 22 GB volume mounted at `/data`.
+
+This step gives you a service ID, a project ID, and an environment ID — keep them; you'll need them in step 2.
+
+### 2. Set these environment variables
+
+```bash
+# Pin the runtime + skills to the verified-working pair
+SENPI_RUNTIME_NPM_SPEC=@senpi/runtime@2.0.0-dev.runtime-phase-2.20260511075330
```
garg-prashant: No, this is no more valid. The version won't have `dev` in its name.

_helpers/senpi_runtime_helpers/references/architecture.md
rachinkapoor: Removed

_helpers/senpi_runtime_helpers/references/runtime-deployment.md
```diff
@@ -0,0 +1,169 @@
...
+| MCP keep-alive working | `grep mcp_initialized /tmp/pangolin-daemon.log \| wc -l` | `1` (per process lifetime) |
+| Tick latency | `grep daemon_tick_finished /tmp/pangolin-daemon.log \| tail` | 3-5 s (not 30+ s) |
+| MCP call latency | `grep mcp_call /tmp/pangolin-daemon.log \| tail` | 250-500 ms typical |
+| Runtime registered | `openclaw senpi runtime list` | one row, status=running |
+| Plugin version | gateway log | matches `SENPI_RUNTIME_NPM_SPEC` |
+
+## Pin rationale
+
+The two pins above (`SENPI_RUNTIME_NPM_SPEC` + `SENPI_SKILLS_BRANCH`) are a matched pair. The `2.0.0-dev.runtime-phase-2.20260511075330` build is a 2.0 dev tag with the senpi-stack `/signals` + `/audit` envelope, the `/state` endpoint that `helper-mcp-envelope-aligned`'s daemon needs for scanner-level liveness, the restored `senpi.{listActions,getActionState,getActionHistory}` gateway methods, and the 90 s LLM-decision abort cap. Pinning a 1.x runtime would produce silent envelope-parse failures and missing `/state` 404s; pinning a 2.x runtime against the older `wrapped-skills` branch (which doesn't expect the new envelope) goes the other way. Keep both halves on 2.x.
```
garg-prashant: ahh, too coupled to a specific version 2.0.0-dev.runtime-phase-2.20260511075330 which is not even the prod version

_helpers/senpi_runtime_helpers/references/runtime-deployment.md
rachinkapoor: Yes, once we release the runtime, the correct version will be in place.

_helpers/senpi_runtime_helpers/references/runtime-deployment.md
```diff
@@ -0,0 +1,169 @@
+# Runtime deployment
```
garg-prashant: remove this file

PR #305, opened 2026-05-22
docs: add DSL-preset selection to strategy-authoring instructions

betashop:
**For team review.** Follow-on to #300 (now merged). Integrates the new `dsl-presets.yaml` into how agents create strategies. Docs only.

## What it does
When creating a strategy, the agent now prompts the user to pick one of the four named DSL presets — **`balanced`** (default), **`let_winners_run`**, **`mean_reversion`**, **`scalp`** — or hand-author a custom one. **If the user gave the agent full autonomy to design the strategy, the agent picks the preset itself** by matching the producer archetype (trend/single-asset/striker/follower → `let_winners_run`; faders/contrarian → `mean_reversion`; high-frequency → `scalp`; else → `balanced`).

## Changes (4 files)
- **`senpi-trading-runtime/SKILL.md`** — preset table + the prompt-or-autopick instruction + archetype→preset mapping, in the DSL section.
- **`senpi-entrypoint` + `senpi-onboard` post-onboarding.md** — extended the "Always recommend DSL" line with the preset-selection step.
- **`dsl-configuration.md`** — fixed a stale "24h outer bound" → "72h" in the balanced table row (it contradicted the preset block + the ⭐ callout, both 72h).

Points at the `dsl-presets.yaml` + `dsl-configuration.md` (already on main).

## Note for review
This documents *copy-paste-the-block* preset selection. Separately proposing that the runtime resolve a preset **by name** (`dsl_preset: balanced`) so agents don't paste blocks at all — see the message to the runtime team. If that lands, this doc updates to "reference the preset by name."

## Test plan
- [ ] Authoring flow now reads: pick archetype → build producer → configure yaml → **pick DSL preset** → verify
- [ ] No code touched

PR #300, opened 2026-05-21
docs: strategy-authoring onboarding — route to runtime refs, two paths, recommend DSL

betashop:
**For team review — not to be merged by me.** This is the consolidated re-open of #297/#298/#299 (admin-merged in error; backed out of main in 43999f1). Docs only, no code.

## Problem

New agents — ours and users' — aren't clear how to write strategies. Every surface an agent hits when building one (`CLAUDE.md`, the runtime `SKILL.md`, the senpi-entrypoint "build a strategy" flow) failed to route them to the canonical references. The build-a-strategy onboarding flow was a freeform brainstorm that never mentioned the runtime, producer-patterns, or the SDK — and said "do NOT require any skill install." Result: agents improvising off-pattern strategies, hand-rolling daemons, and shipping positions with no exit protection.

## Solution (three changes, one PR)

1. **Route to `producer-patterns.md` first** — one ordered read-path (producer-patterns → SDK → yaml → risk-gates/dsl → verify), repeated in all three surfaces.
2. **Two on-ramps, one structure** — Path A (clone a template) and Path B (bring your own scanner) are equals converging on the same runtime + DSL + yaml + risk.
3. **Recommend DSL on every strategy** by default (user can tune/opt out), with a config pointer.

## Important notes

- **Docs only** — no code touched.
- Two load-bearing principles now stated explicitly everywhere: **(1)** the producer ONLY emits signals — never `create_position`, never its own exit logic, never a hand-rolled daemon; the runtime owns execution/DSL/risk. **(2)** DSL on by default — the protection that would've prevented the naked positions from last week's producer-death + late-fill incidents.
- Touches **senpi-entrypoint**, which the team owns — please review the build-flow framing especially.
- Consolidated into one PR because the three changes are layered edits to the same 3 files (couldn't be separate non-conflicting PRs).

## Test plan
- [ ] CLAUDE.md / runtime SKILL.md / entrypoint build-flow all route to producer-patterns.md first
- [ ] Both authoring paths present + the producer-only-emits-signals invariant explicit
- [ ] DSL recommended by default with config pointer
- [ ] No code touched

PR #306, opened 2026-05-22
docs: strategy-creation.md — the self-contained fast path (cut Opus authoring time)

betashop:
**For team review.** Directly targets Ignas's dogfood finding: a fresh Opus agent built a correct bundle from #300's guidance, but took **~6 minutes — almost all discovery roundtrips.** Target is **<3 min on Opus** before we tune for open-source agents.

## Root cause (from Ignas's trace)

The 6 min wasn't authoring — it was discovery: browse repo → find example → fetch 3 example files → **reverse-engineer the runtime.yaml schema from pangolin** → cross-check funding semantics. ~7 fetch-and-reason roundtrips before writing code. The authoring path was spread across 5 files; each was a separate fetch.

## Fix: one self-contained doc

`senpi-trading-runtime/references/strategy-creation.md` collapses the path into a **single fetch**:
- 5-step path + the producer-only-emits-signals invariant
- archetype → example-agent → DSL-preset table (pick all three at once)
- **inline** producer skeleton (real SDK shape)
- **inline** complete `runtime.yaml` template
- the 4 DSL presets, balanced inlined
- verify steps
- **gotchas inlined** — asset/direction top-level, bare model name, `version: 1`, `@senpi-ai/runtime`, and **HL funding is per-hour** (the doc bug Ignas's agent caught). Each of these was a roundtrip otherwise.

`SKILL.md` + `CLAUDE.md` now route "writing a strategy?" to this one doc first; the ordered references become the deep-dive. **Goal: ~2 fetches, not ~7.**

## Not in this PR (deliberate)
- **Onboarding build-flows** repointed to this doc = fast follow after #305 merges (avoids file overlap).
- **The biggest single lever is outside this repo:** the `senpi-overview` MCP guide routes "build a strategy" to the repo *root*, not here — so agents never enter our front door. Separate ask to the MCP-guide owner (+ fix the per-8h funding bug there). That + this doc together are what gets us under 3 min.

## Test plan
- [ ] Re-run Ignas's exact `author_funding_harvest` prompt against this doc; measure roundtrip + time drop
- [ ] Confirm inline schema/skeleton match yaml-schema.md / python-producer-sdk.md
- [ ] No code touched

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,14 +6,9 @@ This repo is almost entirely AI-generated. The rules below exist because they ha
 
 ## ▶ Writing or editing a trading strategy? Start here
 
-Do **not** improvise a strategy from scratch. The canonical authoring path lives in `senpi-trading-runtime/`. Read in this order before writing code:
+Do **not** improvise a strategy from scratch. **Read one self-contained doc first: [`senpi-trading-runtime/references/strategy-creation.md`](senpi-trading-runtime/references/strategy-creation.md).** It has the whole path, an inline producer skeleton, a complete `runtime.yaml`, the DSL presets, the archetype→example map, and the gotchas — enough to build in a single fetch.
 
-1. **`senpi-trading-runtime/references/producer-patterns.md`** — the archetype catalog. Pick the closest pattern; it links to a working example agent. Copy that example.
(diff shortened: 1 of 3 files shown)


betashop:
Superseded by #317 — consolidated into one PR that ships the canonical `strategy-creation.md` and routes every authoring entry point to it (this PR's content is folded in there). Closing in favor of #317.

PR #370, opened 2026-06-18
fix(strategy-discover): belief-first entry-point flow, fund routing, de-bias, stack-by-default

betashop:
## What this fixes

The new `senpi-strategy-discover` skill is the right move (decision tree as code), but the **conversation flow** didn't match the producer-patterns decision-tree intent, and it was biased toward crypto. This PR rewrites the flow (SKILL.md only).

### Problems with the current flow
1. **Wrong ordering.** It was a flat list of five peer questions (belief/assets/risk/horizon/budget). The tree is **belief-FIRST → a follow-up that *depends on the belief* (Layer 2A–2H) → budget/risk only to size & lock in (Layer 3).** The second question must branch on the belief, not be a fixed "assets."
2. **Layer 0 backwards.** Intent: **always start with belief; drop to Layer 0 only if the user can't answer it.** The current version made "I don't know" reopen the belief question and omitted the Layer-0 fallback paths (express lane / plain-language quiz / show-by-vibe / opt-in contextual suggestion) entirely.
3. **Funds unreachable.** `--belief` is `trend|contrarian|copy|breakout|structural|single_market` — no thesis-fund or hedge-fund route, so the funds (the products most users want) can't be surfaced by belief.
4. **No "stack a hedge" step** — it lands on one pick and hands off; never encourages pairing.
5. **Crypto bias** — "when a coin's running," BTC/ETH-first examples — even though ~half of HL volume is now non-crypto (stocks, commodities, pre-IPO).

## What changed (SKILL.md)
- **Belief-first → branch-dependent drill-down → size** (Layer 1 → 2 → 3), with **Layer 0 as the explicit fallback** (express / quiz / vibe / opt-in contextual suggestion) only when the user can't answer belief.
- **Restructured into an entry-point map** — help-me-pick, browse/orient, read-the-market, already-decided, build-custom, and "I don't know" all interconnect; *read the market*, *browse*, and *build custom* are first-class moves available at any point. Covers every start permutation (browse→deploy, browse→build, browse→market→fit, market-first→pick, …).
- **Funds are first-class Layer-1 branches** and lead the reordered question (managed → gut-feel → specific → copy → hot → advanced): `thesis_fund` (a view) / `hedge_fund` (a return style).
- **De-biased from crypto** — asset-agnostic phrasing; stock/pre-IPO/commodity-forward examples (NVDA, SpaceX, gold); a golden rule that ~half the volume isn't crypto.
- **"Stack, don't isolate"** — offer a complementary hedge on any single-wallet pick (`funding_split` absent), à la Spider + Dog; skip it for funds (already internally hedged).
- **Market read** moved to the present-picks step; **never pre-fetch on entry**.

## ⛔ Companion engine changes required before this merges (`scripts/discover.py`)
This PR is the playbook; three small engine additions make it real. Left to @sarvesh to avoid conflicting with active WIP:

- [ ] **Add fund beliefs** — `thesis_fund` + `hedge_fund` to `BELIEF_TO_ARCHETYPE` / `BELIEF_SYN` (syns: "view"/"opinion"/"macro" → thesis_fund; "hedge fund"/"market-neutral"/"income" → hedge_fund), routing to catalog `group == "hedge-fund"` (+ thesis-fund presets). Without this the flow can route to funds but the engine returns none.
- [ ] **Tag funds so they're matchable** — fund packages' `strategy.yaml` `catalog:` block needs the corresponding declared `archetype` (`hedge_fund` / `thesis_fund`) so the generated catalog carries it.
- [ ] **`--hedge-for <id>`** mode → return complements. Cheap: **`ARCH_OPPOSITE` already exists** in the engine (trend↔contrarian, etc.) — reuse it to surface the opposite-archetype top pick, excluding funds.
- [ ] `funding_split` is **already** returned per candidate — that's the stack-vs-don't signal the agent uses. No change; just confirming.

Until those land, the SKILL notes the works-today fallback (route fund-intent via `--assets`; surface a hedge by re-running for the opposite belief).

## Scope
SKILL.md only — I deliberately did **not** touch `discover.py` (it's actively being written). Targets `feat/strategy-discover-engine`, not `main`.

### DIFF
--- a/senpi-strategy-discover/SKILL.md
+++ b/senpi-strategy-discover/SKILL.md
@@ -28,8 +28,14 @@ and recommend in a natural voice. It must never feel like a form or a filter.
   catalog or filter strategies yourself.
 - **Only ever name strategies the engine returned** (in `MatchResult.candidates`). Copy the `id`/`name`
   verbatim from its JSON. If it's not in the JSON, don't say it. This is the anti-hallucination rule.
-- **Lead with the live market read** — the `market_facts` (price move, funding, trend) are what make
-  you sound like an analyst, not a search box.
+- **Lead with the live market read _when you present picks_** — the `market_facts` (price move, funding,
+  trend) are what make you sound like an analyst. But **never pre-fetch on entry**: the opener is a
+  question, not a scan (see Conversation flow).
+- **Not just crypto.** Senpi trades **stocks, commodities, indices, and pre-IPO names 24/7** — about half
+  the volume here isn't crypto. Keep every question, example, and default **asset-agnostic**; never
+  assume "a coin."
+- **Stack, don't isolate.** One strategy is one bet. On any pick that isn't already a multi-wallet fund,
+  offer a complementary hedge (Conversation flow → *Stack, don't isolate*).
 - **Echo your understanding in one line** before showing picks ("got it — cautious, BTC/ETH, ~$300").
 - **Don't re-ask what they've already told you** — if "I'm new" → set `--experience new` and move on;
   if they named an asset/risk, use it. Only ask for genuine gaps.
@@ -45,11 +51,16 @@ Invoke via the `exec` tool with discrete flags (all optional; pass what you've l
 python3 scripts/discover.py [--risk conservative|moderate|aggressive]
   [--assets <csv of class-tags btc_eth,major_alts,universe_crypto,xyz_equities,commodities,indices,pre_ipo
              and/or named tickers BTC,SOL,NVDA>]
-  [--belief trend|contrarian|copy|breakout|structural|single_market]
+  [--belief trend|contrarian|copy|breakout|structural|single_market|thesis_fund|hedge_fund]
   [--horizon scalp|swing|position|hodl] [--direction long_only|short_only|any]
   [--budget <number>] [--exclude <csv>] [--experience new|experienced]
+  [--hedge-for <strategy id>]   # return complements to an already-chosen pick
   [--limit 8] [--offset 0]
 ```
+
+> `thesis_fund` / `hedge_fund` beliefs and `--hedge-for` require the companion engine
+> changes in this PR's checklist; until those land, route fund-intent via `--assets`
+> and surface a hedge by re-running for the opposite belief.
 
 - Values can be loose ("pretty cautious", "btc and eth") — the engine canonicalizes; unknown → ignored.
 - **You hold the flags across turns** and re-run with the FULL set each time (the script is stateless).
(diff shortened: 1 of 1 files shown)


0xsarvesh:
Thanks for this @betashop — the belief-first flow, entry-point map, Layer-0 fallback, crypto de-bias, and "stack, don't isolate" were exactly the right calls, and they're now in the skill.

**Heads-up on how it landed, because it changes the companion-engine checklist.** While this PR was open I simplified `discover.py` to a **filter + return-all** matcher on `feat/strategy-discover-engine`. The matcher now hard-filters only on *concrete* constraints (asset domain, named asset, strict direction, explicit exclusions) and returns the **whole eligible set**; the LLM does all soft ranking (risk / belief / worldview) over the records, which carry author-declared free-text **`thesis` + `tags`**. That removed `--belief`, the relevance score, and `ARCH_OPPOSITE`.

So the three companion engine items don't apply as written — but each is covered a different way:

- **Fund beliefs (`thesis_fund`/`hedge_fund` on `--belief`)** → there's no `--belief` flag anymore. Funds are matched by the LLM ranking the free-text `thesis`/`tags` on each record (`hedge-fund`, `tail-risk`, `all-weather`, `thesis-fund`, `war`, …), generated from producer-patterns §17–27. The reordered Layer-1 question leads with funds, and a broad run surfaces `rhino` / `ox` / `spider` / `thesis-war-escalation` / `thesis-risk-off` — verified reachable.
- **Tag funds so they're matchable** → done via `thesis` + `tags` (free text, no controlled vocab) rather than an `archetype` enum, so open-ended worldviews ("there'll be a war", "all-weather") match without a maintained taxonomy.
- **`--hedge-for <id>` reusing `ARCH_OPPOSITE`** → not needed: since the engine returns *all* eligible, the agent surfaces a hedge by **re-running the engine broadly** and picking a complement from the set (SKILL documents this). `funding_split` is still the stack-vs-skip signal, as you noted.

Your conversation playbook is ported wholesale into `SKILL.md` (now v2.2.0) on the simplified engine — belief-first → belief-dependent follow-up → size, the entry-point map, Layer-0, the asset-agnostic de-bias, and stack-don't-isolate. The only thing deliberately dropped is the belief-flag *mechanism*; the *structure* is all there.

**Recommendation: close this as superseded.** Merging the SKILL.md here would reintroduce `--belief` / `--hedge-for` / `relevance` against the new engine. The work lands on `feat/strategy-discover-engine` (matcher simplification + the ported flow). Happy to walk through the before/after mapping if you want to confirm nothing was lost.

PR #376, opened 2026-06-23
New skill: senpi-portfolio — real-time all-wallet portfolio analysis

betashop:
## What

A new guide/analysis skill, **`senpi-portfolio`**, that answers *"analyze my portfolio across all wallets"* with a precise, real-time wallet/balance taxonomy and genuine analysis — not a flat data dump. Same hidden-engine pattern as `senpi-strategy-discover` / `senpi-market-pulse` / `senpi-smart-money`.

## The problems it solves (all from real agent failures)

1. **Wallet clarity** — main embedded wallet vs. per-strategy sub-wallets vs. deployed-in-positions vs. idle, computed explicitly and labeled by *location*.
2. **The idle-cash trap** — `account_get_portfolio.total_withdrawable` is free margin sitting **inside strategy wallets** (bucket 2), **not** idle cash in the embedded wallet (bucket 1). A user who moved everything into strategies has a **$0 embedded wallet** even when `total_withdrawable` is four figures. The engine computes these as two structurally separate fields so the agent can't conflate them. *(This is the exact bug where an agent reported $1,157 of strategy margin as "idle in the embedded wallet" when it held $0.)* A test guards it.
3. **No cached data** — `account_get_portfolio` caches HL data **12h** unless `forceFetch: true`; the engine always forces a fresh fetch and reads each strategy's live `strategy_get_clearinghouse_state`.
4. **Analysis, not a dump** — every position is compared to the market (24h move, with/against the tape, leveraged return); the portfolio gets net-exposure, concentration, and idle-drag reads.
5. **Funding-ledger correctness** — surfaces `totalFunded`/`totalWithdrawn` so a small balance from withdrawn profits isn't misread as a loss (`netFunded` can be negative — normal).

## The three buckets (reconcile by construction)

```
grand_total = idle_in_embedded      (HL USDC + EVM USDC in the main wallet — truly free)
            + idle_in_strategies     (free margin inside strategy wallets — == total_withdrawable)
            + deployed_in_positions  (equity tied up in positions = account_value − withdrawable)
```

`total_withdrawable` is **bucket 2, not bucket 1** — the whole point of the skill.

## How it works

- `user_get_me` → embedded address; `account_get_portfolio(forceFetch)` → embedded idle (HL USDC + EVM `token_balances`) + aggregate for a reconciliation cross-check.
- `strategy_list` → per-wallet `strategy_get_clearinghouse_state` (live, both DEXes). Each wallet's `withdrawable` = idle-in-strategy; `account_value − withdrawable` = deployed.
- Capped per-asset `market_get_asset_data` → `vs_market` (with/against the tape) on every position.
- Computes exposure (net long/short, by asset), concentration, idle drag, and a `reconciles` flag.

Fails open end-to-end. **Offline fixture test: 7/7**, no network — guards `idle_in_embedded ≠ total_withdrawable` and that the three buckets sum to the total.

## CTAs

> 1. Want me to rebalance or adjust any of these positions?
> 2. Want me to put the idle capital to work in a new strategy?

CTA 1 → execution tools; CTA 2 → strategy-discover / strategy-author / `strategy_top_up`. Propose, never act.

## Review notes

- Field names verified against the Senpi MCP tool schemas + the `senpi-overview` guide. Defensive alias fallbacks; `--dry` dumps raw responses if the live shape differs.
- **USER-scoped `SENPI_AUTH_TOKEN` required** — app-scoped returns no wallet data and the engine sets `meta.degraded`.


betashop:
## Fixed: cross-DEX double-count (`5293a1e`)

Live raw clearinghouse data settled the reconciliation. A strategy wallet's `main` and `xyz` sections are **two views of one wallet, not separate pools** — `withdrawable` is the *shared* idle collateral, reported identically in both, and each view's `accountValue` = shared idle + only that DEX's position equity.

The engine was summing **both** `accountValue` and `withdrawable` across the two views → double-counted the shared collateral (inflated a real ~$3.1K account to ~$5.6K). A naive `max()` fix under-counts the other way when a wallet holds positions on **both** DEXes.

**Correct rule, now implemented per wallet:**
```
shared_idle  = max(main.withdrawable, xyz.withdrawable)   # mirrored — count once
deployed     = Σ max(0, dex.accountValue − shared_idle)   # position equity across both dexes
wallet_value = shared_idle + deployed = main.av + xyz.av − shared_idle
```

Verified on the real cub-short wallet: `main.av 1149.42 / xyz.av 970.67 / shared withdrawable 740.32` → **$1,379.77** (not $2,120.09 summed, not $1,149.42 max). Grand total across the three Cub wallets reconciles to **~$3,102.94**.

- Test fixture rewritten to the **real mirrored-withdrawable shape** (the old one didn't mirror it, which is why it missed this).
- New regression test `test_shared_dex_collateral_not_double_counted` locks the exact case.
- SKILL.md documents the "one wallet, two views" rule so the agent never re-sums them by hand.

Tests **8/8**. This is the "inspect the MCP before writing" lesson — I'd inferred the clearinghouse shape instead of verifying it; raw data fixed both the value and the rule.

betashop:
Superseded by #383 — rebuilt cleanly on strategy-v2 (original was main-based; main is being replaced by strategy-v2).

PR #378, opened 2026-06-24
Discovery-pilot prep: senpi-audit (read-bucket) + senpi-help (capabilities directory)

betashop:
Prep for the **next set of pilots** — testing how agents find less-frequent / denied capabilities so nothing is lost when tools leave the eager list.

## `senpi-audit` — the clean READ bucket
Answers *"what happened?"* — recent activity / a strategy's history / failure investigation. Collapses `audit_get_recent_actions`, `audit_get_strategy_history`, `audit_query`. Hidden-engine pattern, fails open, `--dry`, **offline test 4/4**.

> Note: the DEFER *mutation* tools (cancel/pause/withdraw/transfer/stop-edits) stay first-class — they need the confirmation UX — so only the audit **reads** are skilled. The skillable long tail is the rare reads.

## `senpi-help` — the safety net
An always-available **capabilities directory** (pure-markdown skill, no engine) mapping every Senpi intent → the skill or tool that handles it, *including tools intentionally kept out of the eager list*. Core line: **"a missing tool is never a missing capability."** When the description-scan misses, the agent consults the directory and routes — no dead ends. This is what makes aggressive tool-denial safe.

## Why these two
Jason's concern: as we deny the rarely-used tail, one big catch-all bucket has poor discoverability — a description can't precisely signal each distinct rare capability, and with no lazy loading a missed skill = a lost capability. The answer is **coherent buckets (not one catch-all) + a directory backstop**. These two artifacts + the discovery-test let us expand denials *safely* — proven findable, with a fallback.

Both descriptions are spec-clean (≤1024 chars); `senpi-help` has no engine (Level-2 directory only).


betashop:
Superseded by 385 — rebuilt cleanly on strategy-v2 (original was main-based; main is being replaced).

PR #380, opened 2026-06-24
Runtime 3.0 — interactive strategy-author skill (build one decision at a time)

betashop:
**For @Sarvesh — runtime-3 only. The diff is a single file: `senpi-strategy-author/SKILL.md`.**

Turns the strategy-authoring guide into a skill whose **default behavior is the interactive build** — and aligns the skill to the **Runtime 3.0 supervised contract** (`scan(inputs, ctx)`, pure `scoring.py`, `marginPct` *intent*, `ctx.state`, named DSL presets), sourced entirely from your `references/creating-a-strategy.md`.

## What the skill now does out of the box
- **Asks the 7 design decisions ONE at a time** (ask → reflect → next). Never dumps all 7 or recites the guide — that wall-of-questions was the failure mode.
- **Mines the opening ask for constraints already stated** so they aren't dropped — e.g. "rotate the cohort every 3 days" → the **Memory** decision (a 3-day cohort cache in `ctx.state`). (From a live run where the agent lost exactly that detail.)
- **Replays the full captured spec for a "yes" before writing code.**
- Then **assembles** (`scoring.py` / `scan.py` / `runtime.yaml` / `strategy.yaml`) → unit-tests math → **smoke-tests** (confirm the runtime *accepted* a signal, not just ticked).
- **Wallets & concurrency:** a new strategy gets its **own new wallet and runs alongside** existing ones — never "shut down X first." Multiple wallets/instances concurrently is normal (long book + short hedge = two instances). (From a live run where the agent wrongly framed concurrency as a blocker.)

## Scope — runtime 3.0 only
The SKILL.md teaches only the supervised `scan(inputs, ctx)` model and references only the runtime-3 docs (`creating-a-strategy.md`, `dsl-presets.yaml`, `glossary.yaml`, `validate_strategy.py`). It **supersedes** the stale `producer_daemon` / `push_signal` / `load_params` / `scanner.py` framing the old SKILL.md carried. SKILL.md is ~2.4k tokens (Level-2 < 5k); description 598 chars (spec-clean).

## Flag for you (your branch — your call, not in this diff)
5 references in `senpi-strategy-author/references/` still describe the **retired daemon model** and are no longer referenced by the runtime-3 skill or `creating-a-strategy.md` — recommend removing/migrating as part of the runtime-3 cleanup:
- `producer-patterns.md`, `python-producer-sdk.md`, `strategy-creation.md`, `momentum-guarded-strategy.md`, `signal-schema.md`
- Note: `yaml-schema.md` links `momentum-guarded-strategy.md` — fix that link if you remove it.

I left these out of this PR so the diff stays just the runtime-3 skill; happy to do the prune in a follow-up if you'd rather.

## Testing notes
- The **authoring conversation + package generation** is host-agnostic — testable on any agent today.
- The **deploy/smoke-test half needs a `@senpi-ai/runtime` 3.x host** (an old-runtime agent will fall back to the daemon path). Suggest validating the full `scan → signal → runtime-accepted` loop on a v3 host before fleet rollout.
- Follow-up worth adding to `senpi-strategy-ops` deploy: a **runtime-version guard** (`requires.runtime` vs the host's `@senpi-ai/runtime`) so a v1/v3 mismatch fails loud instead of silently daemon-falling-back.

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -54,6 +54,9 @@ For each: ask the question, offer the options as plain choices, then map the ans
 1. **Universe — "What should it watch and trade?"**
    A) one asset · B) a fixed basket you name · C) dynamic (scan everything, filter by volume) ·
    D) derived (trade what the best traders / a cohort hold). → sets how `scan()` builds its list.
+   **Verify every ticker the user names (A/B) against `market_list_instruments` before it enters the
+   package — a ticker that isn't a live instrument silently no-trades. The broad index is `xyz:XYZ100`,
+   not `xyz:NASDAQ`; check, don't assume.**
 2. **Data — "What does it read to decide?"**
    candles (`market_get_asset_data`) · funding/OI (`market_get_funding_*`) · smart-money
    (`leaderboard_*` / `discovery_*`) · cross-asset flow. → the `call_tool`s in `scan()`.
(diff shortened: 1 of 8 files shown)


betashop:
Closing in favor of #386 — they are **not** equivalent, and merging this one would regress `strategy-v2`:

- **Reverts the cooldown-units fix.** `strategy-v2` already uses `cooldown_minutes` across the risk reference files; this branch still carries `cooldown_seconds`, which would reintroduce the known silent-live-bug (wrong unit).
- **Stale strategy-ops scripts.** `deploy.py` + `validate_universe.py` here have diverged from the recent strategy-ops work already on `strategy-v2`.

The valuable author content (`SKILL.md` + `references/creating-a-strategy.md`) is byte-identical to #386, which is based on current `strategy-v2` and carries none of the regressions. **#386 supersedes this.**

PR #385, opened 2026-06-25
Discovery-pilot prep: senpi-audit (read-bucket) + senpi-help (capabilities directory)

betashop:
> ## ⚠️ Assumption — DO NOT MERGE until @Sarvesh enables Tool Search
> This pilot is designed for the **Tool-Search-enabled** configuration. It assumes OpenClaw
> **`tools.toolSearch`** is turned on in **`directory` mode** so the catalog is *deferred, not denied*
> — the model sees compact descriptors + a hot set of likely schemas, and hydrates the rest on demand.
> **No tools are denied by this PR.** Two prerequisites are Sarvesh's / infra's call:
> 1. **Version** — the deployed OpenClaw build must include `tools.toolSearch`. Our last-checked build
>    (2026.5.7) showed no tool-search symbols in source, so this likely needs a **Railway-template
>    version bump**, not just a flag.
> 2. **Config** — `{ tools: { toolSearch: { mode: "directory" } } }` (per-host for a pilot, template for fleet).
>
> Until both are in place, these skills still work (tools load eagerly, no context savings yet) — so the
> PR is safe to hold, but the *context win* only lands once Sarvesh enables Tool Search. **Holding for Sarvesh.**

---

Prep for the **next set of pilots** — testing how agents discover the less-frequent capabilities once the catalog moves behind Tool Search, so nothing is lost when full schemas leave the eager prompt.

## `senpi-audit` — the clean READ bucket
Answers *"what happened?"* — recent activity / a strategy's history / failure investigation. Collapses `audit_get_recent_actions`, `audit_get_strategy_history`, `audit_query`. Hidden-engine pattern, fails open, `--dry`, **offline test 4/4**.

> Note: the *mutation* tools (cancel/pause/withdraw/transfer/stop-edits) stay first-class — they need the confirmation UX — so only the audit **reads** are skilled. The skillable long tail is the rare reads.

## `senpi-help` + always-loaded router — the discovery layer
An always-available **capabilities directory** (pure-markdown skill, no engine) mapping every Senpi intent → the skill or tool that handles it. Core line: **"a missing tool is never a missing capability."** Paired with the always-loaded `AGENTS.md` router block (intent→skill map, in context every turn) and the find-then-stop ladder:

1. Senpi skill first → 2. then `senpi-help` → 3. then **Tool Search** (`tool_search`/`tool_describe`/`tool_call`, or `tool_search_code`) → 4. cap at ~5 attempts, then stop and suggest another path (no 10–20-min grinds).

## Why these two (under Tool Search)
With Tool Search in `directory` mode, full schemas leave the prompt but tool **names + descriptions stay visible** in the bounded directory — and anything beyond it is one `tool_search` away. The skills act as the **hot set** (the common reads, pre-loaded), `senpi-help` + the router give **human-curated discovery** on top of the raw catalog, and the ladder guarantees the agent **routes before giving up**. Net: the context win of deferral, with discovery *better* than raw Tool Search alone — and **zero denials**, so nothing is ever unreachable.

Both descriptions are spec-clean (≤1024 chars); `senpi-help` has no engine (Level-2 directory only).

## Measuring it (once enabled)
Tool Search ships built-in telemetry — serialized prompt bytes, catalog size, search/describe/call counts — and a gateway E2E proving the payload shrinks. So we get a **real before/after number**, not an estimate.

---
_Rebuilt cleanly on `strategy-v2` — supersedes #378 (original was main-based; strategy-v2 replaces main)._

### DIFF
--- a/senpi-audit/README.md
+++ b/senpi-audit/README.md
@@ -1,0 +1,28 @@
+# senpi-audit
+
+A Senpi skill that answers **"what happened?"** — recent account activity, a single strategy's change
+history, or failure investigation. The clean read-bucket for the audit tools.
+
+Same hidden-engine pattern as the other read skills. Collapses: `audit_get_recent_actions`,
+`audit_get_strategy_history`, `audit_query`.
+
+## Modes
+
+```sh
+python3 scripts/audit.py                  # recent activity across the account
+python3 scripts/audit.py --strategy <id>  # one strategy's full mutation history
+python3 scripts/audit.py --failures       # only failed operations (debugging)
+python3 scripts/audit.py --tool <name>    # filter to one tool
+python3 scripts/audit.py --dry            # raw schema dump
+python3 scripts/audit.py --fixture tests/fixtures/audit_fixture.json   # offline (tests)
+```
+
+Returns `{entries, summary, meta}` — each entry carries `time`, `action_type`, `tool`, `success`,
+`resource`, and the agent's `reason` for the action. `summary` rolls up counts + surfaces failures.
+
+USER-scoped token required (defaults to the authenticated user's log). Offline test 4/4.
+
+> **Role in the context-reduction plan:** this is the **read** half of the DEFER bucket — a coherent,
+> findable skill so "what happened?" doesn't require the 3 raw `audit_*` tools in the eager list. The
+> *mutation* DEFER tools (cancel/pause/withdraw/transfer/stops-edit) stay first-class (they need the
+> confirmation UX), so they are not collapsed here.
--- a/senpi-audit/SKILL.md
+++ b/senpi-audit/SKILL.md
@@ -1,0 +1,80 @@
+---
+name: senpi-audit
+description: >-
+  Answer "what happened?" — review the user's recent Senpi activity, a single
+  strategy's change history, or investigate failures. Use for "what did my
+  strategy do", "change history for my strategy", "what happened yesterday",
+  "show my activity", "why did that fail", "what trades were made". A
+  hidden engine (scripts/audit.py) pulls and normalizes the audit trail; you
+  summarize it. Requires a USER-scoped Senpi token. Not for live positions
+  (senpi-portfolio) or market data (senpi-market-pulse).
+license: Apache-2.0
+compatibility: OpenClaw, Hyperclaw, Claude Code
+metadata:
+  author: Senpi
+  version: "1.0.0"
+  platform: senpi
+  exchange: hyperliquid
+---
+
+# Senpi Audit — what happened
+
+A hidden engine pulls the audit trail; **your job is to make it readable** — what the account (or a
+strategy) did, and why, with failures surfaced.
+
+## Golden rules
+
+- **Run the engine; never hand-pull.** `python3 scripts/audit.py` (recent), `--strategy <id>` (one
+  strategy's history), `--failures` (debugging), `--tool <name>` (one tool). Read its JSON.
+- **Lead with the answer to "what happened," then the detail.** Summarize the activity (counts by
+  action, the notable mutations), then list entries if asked.
+- **Surface failures loudly.** Any `success: false` entry is a flag — lead with it when debugging.
+- **Cite the reasoning.** Audit entries carry the agent's `reason` for each action — quote it; that's
+  the "why" the user actually wants.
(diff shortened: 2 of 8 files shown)


betashop:
**Scope clarification (per Jason): this pilot denies no tools.**

Earlier framing assumed we'd `deny` the read tools once skills covered them. We're **not** doing that by default — the risk (a denied tool is unreachable if routing ever misses) outweighs the token savings until @Sarvesh and Jason discuss further.

What this PR actually ships is **routing by order-of-preference**, not access control:
1. Senpi skill first (faster/cheaper when one fits)
2. then `senpi-help` directory
3. then **Tool Search** — pulls up any tool by name and calls it directly; nothing is hidden or removed
4. cap at ~5 attempts, then stop and suggest another path

Tool Search is what makes Deny unnecessary: every tool stays reachable, and a skill is just the preferred route. Whether we later collapse any tool schemas to save context is a **separate, deferred decision** — not baked into this PR. (The diff contains zero `tools.deny/allow` config.)

betashop:
**Verified against OpenClaw source (v2026.5.7) — settles the defer-vs-deny question.**

Ferried an agent to grep its own runtime. Findings, with source evidence:
- **Schema loading is fully eager.** `createOpenClawCodingTools()` rebuilds the entire tool list every turn (`pi-tools.ts`); no deferred/lazy mechanism.
- **No tool-search tool exists** (the `deferred-followup` file just rewrites `exec`/`process` descriptions — unrelated).
- **Deny = fully removed** from the model's view (`filterToolsByPolicy` in `pi-tools.policy.ts` — `.filter()`, denied tools never appear).
- **No config flag** for lazy loading or per-turn schema caps.

**Consequence:** on OpenClaw, **deny is the only lever that reduces per-turn tool context** — there is no defer-not-deny. With deny off (current decision), these skills add capability/quality and set up a *future* safe deny, but save no per-turn context yet. A tool can only be safely denied once a skill fully covers it.

Router corrected accordingly: step 3 no longer says 'Tool Search' (unfollowable here) — it points to calling the tool directly from the already-loaded list, found via `senpi-help`.

betashop:
**Correction to my earlier 'no tool search' comment — that was wrong.** OpenClaw **does** have Tool Search (`tools.toolSearch`); it's **experimental and off by default**, so the ferried agent saw only the eager path and concluded the feature didn't exist. The authoritative OpenClaw docs confirm it.

**What it actually is:** compact descriptors instead of full schemas up front; the model uses `tool_search` / `tool_describe` / `tool_call` (or the `tool_search_code` JS bridge) to discover and call any catalog tool on demand. Policy-filtered catalog spans OpenClaw + plugin + MCP + client tools. Every call still crosses back to the Gateway, so allow/deny/approval/hooks/telemetry all still apply (fail-closed). Three modes — `code` (default), `tools`, `directory`. **`directory` looks ideal for Senpi:** a bounded name+description directory + a small hot set of likely schemas + on-demand hydration. Enable: `{ tools: { toolSearch: { mode: "directory" } } }`.

**This overturns the prior conclusion:** defer-not-deny IS available on OpenClaw. We do **not** need to deny tools to cut context — enable Tool Search, keep every tool reachable, and full schemas stop being sent up front. Built-in telemetry (serialized prompt bytes, catalog size, search/describe/call counts) makes the savings measurable. Router updated: step 3 now points to Tool Search when enabled.

betashop:
**@Sarvesh — decision needed before merge.** This PR is written *assuming* OpenClaw Tool Search is enabled (`tools.toolSearch`, **`directory` mode**) so the catalog is deferred-not-denied. Two things are yours to confirm:

1. **Does our deployed OpenClaw build include `tools.toolSearch`?** Our last-checked build (2026.5.7) showed no tool-search symbols — so this may need a Railway-template version bump, not just a flag.
2. **OK to enable `directory` mode** (`{ tools: { toolSearch: { mode: "directory" } } }`) — per-host for a pilot first, fleet later?

No tools are denied here, so the PR is safe to hold. The context savings land only once Tool Search is on. **Holding for your go.** (Ignore my two comments above — they were a wrong 'no tool search exists' finding and its correction.)

betashop:
Superseded by #388 — all read skills consolidated onto strategy-v2 in one branch (`sv2-read-skills`). This skill's latest version (with all cross-cutting fixes) is included there. Closing to clear the piecemeal stack.

PR #394, opened 2026-06-26
docs(scan-contract): guard optional MCP reads so a read error can't kill the tick

betashop:
## What
Adds a **"Reads can fail — guard the optional ones"** subsection to `scan-contract.md` (the canonical `scan(inputs, ctx)` author contract), right after the read-only MCP boundary.

## Why
`ctx.senpi_mcp.call_tool` **propagates** read errors, and per the contract any exception in a tick rolls the whole tick back to `[]`. So an unguarded **optional** read (smart-money / discovery / cross-asset flow that only *adds* to the score) erroring makes a scanner **silently never emit** — not crash visibly, just quietly stop trading.

Generalized authoring guidance surfaced by the first live Runtime-3.0 scan deploy (Kodiak): an unguarded `leaderboard_get_markets` read would have zeroed the scanner on any auth/rate-limit gap. Documents the **required-vs-optional** split and the **try/except-and-degrade** pattern so every future scan author (and the curator skill) inherits it.

## Scope
Docs only. The Kodiak-specific debug instrumentation (per-cycle stderr log, per-tick history) is deliberately **not** generalized here — that's debug-only, per prior scoping.

### DIFF
--- a/senpi-trading-runtime/references/scan-contract.md
+++ b/senpi-trading-runtime/references/scan-contract.md
@@ -130,6 +130,29 @@ raises `PermissionError` (loud, fail-fast — never a silent `None`).
 The allowlist is scaffold-owned in source and **empty by default** — there is no env/operator/author
 knob. As an author: **assume you cannot mutate anything.** Produce signals; the runtime executes.
 
+### Reads can fail — guard the optional ones
+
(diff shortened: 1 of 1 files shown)


betashop:

PR #411, opened 2026-06-29
fix(discover): named-asset queries no longer exclude universe scanners

betashop:
## Problem
`--assets SOL` returned 27 survivors but **0 universe scanners** — octopus/python/otter/condor/owl were all hard-rejected because their `assets` list is empty (dynamic universe). The broad scanners that would actually trade SOL never reached the ranker; the broaden-to-class fallback only fires when survivors is *empty*, so it didn't help.

## Fix
- `_record_trades_named(r, named)`: a record matches a named ticker by explicit fixed-asset match **or** as an in-domain universe scanner (`asset_scope: universe` covers any name in its domain). Cross-domain stays excluded (a crypto-only scanner is *not* returned for NVDA).
- Honest caveat on those cards: *"Scans the whole universe — trades {ticker} among many names, not exclusively."*
- `test_degrade` updated: DOGE is now covered directly by universe scanners (no broaden needed).

After: `--assets SOL` → 42 survivors incl. octopus/python/otter/condor/owl, fixed SOL specialists still present; `--assets NVDA` → caracal (xyz-universe) present, crypto-only scanners correctly absent. **Not merged.**

### DIFF
--- a/senpi-strategy-discover/scripts/discover.py
+++ b/senpi-strategy-discover/scripts/discover.py
@@ -177,6 +177,19 @@ def infer_class_for_named(named):
     return "major_alts"
 
 
+def _record_trades_named(r, named):
+    """True if record r could trade ticker `named` — by an explicit fixed-asset match OR as an
+    in-domain universe scanner. A broad scanner (asset_scope 'universe') ships no fixed `assets`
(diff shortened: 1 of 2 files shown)


betashop:

PR #412, opened 2026-06-29
fix(discover): named-asset broaden no longer mis-buckets non-crypto tickers

betashop:
## Problem
`infer_class_for_named()` (the named-asset broaden fallback) returned `major_alts` for every non-BTC/ETH ticker — so an equity passed without the `xyz:` prefix (e.g. `NVDA`) would broaden a query into the **crypto** bucket, surfacing the wrong domain.

## Fix
- Only known crypto majors (`_CRYPTO_NAMED`) infer `major_alts`; `BTC`/`ETH` → `btc_eth`; `xyz:*` → no constraint (unchanged).
- An unknown ticker → **no class constraint** (broaden wide, never a wrong-domain cut) — consistent with the engine's 'a bad cut hides the answer, a full list never does' principle.
- Adds `test_infer_class`.

Lowest-severity of the three (the broaden path is rarely hit, especially once #411 lands), but it removes a latent wrong-domain bug. **Not merged.**

### DIFF
--- a/senpi-strategy-discover/scripts/discover.py
+++ b/senpi-strategy-discover/scripts/discover.py
@@ -168,13 +168,22 @@ def asset_matches(named, assets):
     return False
 
 
+# Crypto majors the fleet trades by name — used only to infer a class for the named-asset broaden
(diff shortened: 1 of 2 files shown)


betashop:

PR #424, opened 2026-07-01
feat: make agents aware deployed strategies are already DSL-protected

betashop:
**Problem:** consumer agents recommend *"set a 10–15% strategy-level SL on all active strategies — I can do this with `strategy_update`."* They treat deployed strategies as naked positions because nothing surfaces the built-in DSL. Every strategy deployed from a Senpi template already runs a two-phase DSL trailing stop-loss on each position **plus** daily-loss/drawdown circuit breakers, enforced every tick — so that recommendation is redundant and reads as not knowing the book.

Prompt-layer guardrail placed where risk advice + deploys actually happen:

- **senpi-portfolio** (1.0.0→1.0.1) — golden rule: deployed template strategies are already risk-managed; when flagging risk, say what's *already protecting* the position; never prescribe a redundant manual SL; distinguish **managed** (template-deployed) vs **naked** (manual HL position / raw custom strategy with no runtime); "tighter than built-in" = a DSL/guard-rail change, not a missing stop. CTA1 (`strategy_update`) qualified the same way.
- **senpi-strategy-ops** (2.0.1→2.0.2) — the deploy-completion summary now states the built-in protection on **every** deploy, and an explicit "never recommend a manual SL on a template strategy" rule.
- **senpi-strategy-author** (1.0.0→1.0.1) — invariant: protection is built in (DSL + guard-rails per instance); the only unprotected path is a raw `strategy_create_custom_strategy`.

The behavioral rule is intentionally **duplicated, not shared** — skills load independently, so each must be self-contained; a stable rule is safe to repeat. Per-strategy protection *facts* stay as data (`runtime.yaml`), never hand-copied prose (see the answer to Q2 in the thread).

### DIFF
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -11,7 +11,7 @@ license: Apache-2.0
 compatibility: OpenClaw, Hyperclaw, Claude Code
 metadata:
   author: Senpi
-  version: "1.0.0"
+  version: "1.0.1"
   platform: senpi
   exchange: hyperliquid
 ---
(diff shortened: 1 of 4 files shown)


betashop:
Superseded. Author-skill DSL bullet → the full 3.0 rewrite in the C1 PR; portfolio golden rule → #428; validator DSL-exit check → the C1 PR; ops guardrail → already merged (ops is at 2.1.1). Nothing lost.

PR #429, opened 2026-07-01
fix: rewrite senpi-strategy-author to Runtime 3.0 + document marginPct (C1 + H2)

betashop:
Fixes the review's **C1** (the biggest finding) + **H2**.

## C1 — `senpi-strategy-author` taught the retired v2 model
The skill an agent loads to build a strategy taught the v2 producer-daemon model (`senpi_runtime_helpers` / `push_signal` / `producer_daemon` / `load_params` / `strategy.yaml params` / `config.fields` / 404 `-producer.py` URLs) while the fleet, validator, ops, and discover are all **Runtime 3.0**. So it contradicted its own validator, and an agent following it produced a **non-deployable** package.

Rewrote the whole skill to the shipped 3.0 model — `scan(inputs, ctx) -> list[dict]`, sibling `scoring.py`, `inputs:`, `ctx.senpi_mcp`, `marginPct` — pointing at the authoritative `scan-contract.md`:
- **`SKILL.md`** → 3.0, **v1.0.0 → 1.1.0**.
- **6 reference docs rewritten** (strategy-creation, python-producer-sdk → "there is no SDK", signal-schema, yaml-schema, strategy-yaml-schema, strategy-examples, momentum-guarded) — each grounded in `scan-contract.md` + `runtime-yaml.md` + the kodiak/bison gold templates.
- **`producer-patterns.md` deleted** (v2 artifact) — its archetype→gold-template map absorbed into `strategy-creation.md`; all references repointed; grep is 0.
- **`validate_strategy.py`**: re-added the DSL-exit-block check (a strategy with no exit is a naked position) — **all 78 pass**.
- **`tests/test_no_v2_tokens.py`**: a drift guard that fails if any retired v2 token is *used* in the skill — so it can't silently rot back (the exact failure mode that caused C1). **11 docs clean.**

## H2 — the contract omitted `marginPct`
`scan-contract.md` documented `marginUsd` and never `marginPct`, but the fleet uses `marginPct` (~97/102). Added `marginPct` as the fleet-standard sizing field. **senpi-trading-runtime skill 3.0.1 → 3.1.0.**

## Dogfood
The subagent materialized this doc's own Step-3/Step-5 templates into a throwaway package and ran the real validator → **✓ pass**. The fast-path an author follows is confirmed deployable.

Both skills minor-bumped (auto-upgrade installed users). **Supersedes the author-skill portion of the now-closed #424.** Does not touch main-era files.


betashop:
Closing — superseded by #386 (@betashop's interactive Runtime-3.0 author-skill rewrite, which is the intended C1 fix and correctly *deletes* the retired daemon-model ref docs rather than rewriting them, as I mistakenly did here). The two #386-independent wins from this PR — marginPct in scan-contract (H2) + the validator DSL-exit check — are salvaged into a separate small PR. The retired-ref pruning + drift guard should land with #386 (its final ref set decides them).

PR #431, opened 2026-07-02
feat: strategies become known state at deploy — memory step in ops (2.2.0) + author (2.1.0)

betashop:
Fixes the "agent doesn't know the strategies it launched" problem on both sides of the deploy: the strategy's **name · mandate · DSL preset** become durable known-state the moment it deploys, so later "how's my `<label>` doing?" is judged against what it was *designed* to do — not a generic momentum benchmark, and not a re-discovery.

## `senpi-strategy-ops` — 2.1.1 → **2.2.0**
New **"Remember it — the last step of every deploy (and close)"** section. After a deploy, ops records to memory:
- `strategy_wallet → label` (e.g. `0x4c68…9f91 → ox`)
- one-line **mandate** + expected behavior (from the catalog `belief_plain`/`thesis`/`archetype`)
- **DSL preset** (intent) from `runtime.yaml exit:`
- **Not** live state — PnL / positions / DSL floor / is-running are read live, never memorized
- Reconcile against `status.py`/`strategy_list` before asserting; full config = look it up in the package, not memory

This fires for author-built strategies too (they deploy through ops).

## `senpi-strategy-author` — 2.0.0 → **2.1.0**
New note in **Handoff**: the `strategy.yaml` `belief_plain`/`thesis` the author writes **is** what ops records to memory and what `senpi-portfolio` later judges the strategy against — so a vague or momentum-flavored mandate is exactly how an all-weather core, a crisis hedge, or a selective waiting strategy gets misread as "dead weight." Write the mandate true. No behavior change.

Both are MINOR bumps → auto-upgrade installed users. Pairs with the portfolio enrichment in #428 (memory holds the names + why; the live engine holds the truth; both judged against the mandate).

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -12,7 +12,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.0.0"
+  version: "2.1.0"
   platform: senpi
   exchange: hyperliquid
   requires:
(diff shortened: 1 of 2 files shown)


betashop:
Pulling this back. Per Sarvesh: with Runtime 3.0 every deployed strategy already carries a strategy.yaml in the runtime that describes everything (belief_plain/thesis/archetype/direction/asset_classes/DSL) — that's the source of truth, versioned with the deploy and always current. The portfolio skill should read THAT rather than the agent re-deriving the mandate into memory. Folding the effort into #428 (portfolio) instead.

PR #453, opened 2026-07-06
improve-trades: detect partial closes (profit taken on open positions) — v1.2.0

betashop:
`review.py` reconstructed trades **only** from `discovery_get_trader_history` → `closedPositions[]` = **fully-closed only**. A partial close (TP/SL fill, size reduction on a position that stays open) creates **no closed-trade entry**. So a user who'd taken **~80% profit off two open positions** got "no closed trades, nothing to review" — and guardrail 9 framed a genuinely-trading strategy as "brand-new." The skill's whole premise ("did I exit well?") was blind to partial exits — *the* most common "sold too early" pattern.

## Fix
- **`fetch_partial_closes()`** — reads each current strategy's open positions (`strategy_get_clearinghouse_state`), batches the `(wallet, coin)` pairs into one **`discovery_get_open_position_realized_pnl`** call (the purpose-built "profit already taken on a still-open position" tool the engine never used), and surfaces `partial_closes[]` (realized taken, fees, remaining exposure). Fail-open.
- Wired into `run()` + `step_strategies()` (consistent) with `meta.has_partial_closes`; the "no closed trades → degraded" branch now also requires no partial closes.
- **Guardrail 9 rewritten:** "no closed trades" ≠ "no activity" — check `partial_closes` first, review the profit-taking + remaining exposure; the fresh-strategy framing only at genuinely-zero activity.
- v1.1.1 → **1.2.0**. Tests **46/46** (6 new: detection, flat, truly-empty, fail-open, current-only, step/all consistency).

## Two flags for the reviewer
1. **Field-name verified against source.** The MCP handler returns the list under **`realized_pnl`** (snake_case — `createResponse` renames the GraphQL `realizedPnl`); the extraction reads both defensively so it can't silently null. Confirmed at `senpi-hyperliquid-mcp/src/tools/discovery.tools.ts`.
2. **Fixture is type-derived, not a live recording** — the Senpi MCP wasn't callable this session, so the fixture is built from the MCP's type definitions (`senpi.types.ts`) + the verified handler, not a recorded response. **Wants a quick live smoke-test before merge** (the exact "fixtures from real shapes" discipline the strategy-v2 review flagged).

### DIFF
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -6,10 +6,11 @@ description: >-
   to the best whales", "how could I make more gains", "suggest improvements", "review my trades", "am I
   getting shaken out too early / how are my exits firing", "what did my own limits block / what couldn't
   I take", "where am I leaking", "walk me through / explain my [asset] trade", "what am I paying in fees /
-  maker vs taker", "why is [strategy] losing". A hidden engine (scripts/review.py) reconstructs every
-  CLOSED trade from discovery, enriches each exit reason + blocked signals from the runtime telemetry
-  event log, computes the honest "if I'd held to now" counterfactual, and crosses the book against what
-  the market did — you narrate it under strict guardrails: process over outcome (lead with the aggregate,
+  maker vs taker", "why is [strategy] losing", "did I take profit on my open positions". A hidden engine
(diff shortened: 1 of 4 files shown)


betashop:

PR #454, opened 2026-07-06
Cuttlefish v1.0.0 — regime-adaptive L/S divergence baskets (fleet-first CLOSE_POSITION)

betashop:
## What

New catalog strategy: **Cuttlefish — Regime-Adaptive Divergence Baskets** 🦑. Two books on separate wallets; the differentiator is that it **auto-adjusts the existing book on re-reads of the market** — no other basket package does; they score once at entry and let DSL attrition do the rest.

Built on the **real engines** of the two skills it composes (math ported from `senpi-market-pulse/scripts/pulse.py` and `senpi-smart-money/scripts/smartmoney.py`, constants verbatim — commit `4bfad934` replaced the v1 leaderboard proxy).

## Design

**The market read (every 15m tick):**
- **Universe: derived live** — every main-dex perp over $25M 24h notional, top-16 by volume, off the instruments read. No preset list (`asset_scope: universe`).
- **Pulse gate** (market-pulse `compute_signals` port): cross-asset day classification — crypto (= the derived universe), semis, megacap software, crypto proxies, indices, commodities, macro FX groups off `prevDayPx`, ±0.5% up/down with 2:1 dominance — plus the GOLD/DXY/VIX checklist. The **long book stands down on a risk_off day, the short book on risk_on**.
- **Divergence core** (smart-money engine port): the PROVEN cohort (lifetime realized ≥ $1M) vs the CROWD ($10k–100k) from discovery pages + trader-state batches; per-coin bias = net/gross signed notional; divergence = opposite sides or gap ≥ 0.50 with ≥ 5 members each. Cohorts are heavy, so they ride a **4h-clock cache in ctx.state** — never per tick. Hard block when the proven cohort leans ≥ 0.40 against the book.
- The 4h leaderboard board is demoted to its actual role in those engines: the **near-term confirm factor** (and the fallback block when discovery is unavailable — flagged in reasons, never silent).
- Conviction bands → leverage + margin: long 6x/5x/4x @ 22/17/12%, short 5x/4x/3x @ 20/15/10% (stricter short gates), leverage clamped to per-asset venue max.

**Rebalance — the auto-adjust half** (every 15m, `rebalance.py` → rule-mode **CLOSE_POSITION** action, first use in the fleet):
1. `pulse_flip` — the pulse day against the book for 2 consecutive ticks (anti-whipsaw) → close the book.
2. `divergence_reversed` — the proven cohort flips ≥ 0.40 against a held name → close it now (board ≥ 58% fallback when cohorts unavailable, flagged).
3. `basket_refresh` — every `regimeRefreshHours` (default 4h) held names re-scoring < 3.5 get recycled.

DSL still owns every price-action exit (long: 12%/6 phase-1, ladder to +80%; short tighter: 10%/5, ladder to +60%; weak-peak 360m/240m). Rebalance owns **regime** exits only.

## Novel wiring — please review these specifically

- **CLOSE_POSITION action** (`cuttlefish_*_adjust`): semantics source-verified in `senpi-trading-runtime/src/actions/close-position/close-position.action.ts` — rule mode passes scanner signals through; per-signal `hasOpenPosition` guard makes a stale close a clean `no_open_position` skip (idempotent); `reason` lands in telemetry as `closeReason`.
- **Two external scanners in one runtime**: supported by construction (`create-scanner-module.ts` builds per scanner entry) but unprecedented in fleet YAMLs.
- **Cohort reads in a scanner**: `discovery_*` is read-allowlisted for `ctx.senpi_mcp`, but needs a user-scoped token — when empty, the composite degrades to the near-term board and says so (smartmoney.py's honest-degrade pattern). Worth one live canary tick before merge.
- Entries and close signals are strictly separated by scanner (both actions are rule-mode passthrough).
- Post-close re-entry bounded by `per_asset_cooldown_seconds: 14400`.

## Gates

- `py_compile` clean; **35** engine-semantics tests + **76** package-shape tests pass (incl. cohort-constants-verbatim, 4h cache clock, and no-hardcoded-universe guards).
- `gen_catalog` → 80 strategies, zero cuttlefish warnings; both catalogs regenerated.
- `validate_universe` → **4/4 live** (the only hardcoded names are pulse checklist anchors).
- ⚠️ Mechanical `catalog.json` conflict expected with #455 (gorilla) — whichever merges second re-runs `gen_catalog.py`.

## Deploy

Standard package: `senpi-strategy-ops` with `CUTTLEFISH_LONG_WALLET` / `CUTTLEFISH_SHORT_WALLET`, funding 50/50, min budget $200.


betashop:
✅ validate_universe caveat cleared — gateway is back up; ran on the branch: `strategies/cuttlefish: 12 hardcoded ticker(s), all live`. (Earlier failure was the local MCP gateway being down, plus a validator footgun worth a separate fix: running it against a package dir that doesn't exist on the current branch silently passes as 'derived universe' instead of erroring.)

betashop:
Pushed `4bfad934` — same correction as gorilla: the thesis now runs the **real** market-pulse + smart-money engines (ported verbatim), not a leaderboard proxy. Pulse day classification gates the books every 15m; proven-vs-crowd cohort divergence (4h-clock cache) picks and un-picks the names; the 4h board is demoted to near-term confirm. Universe also now derived live (same lesson as gorilla). 35+76 tests, catalog clean, 4/4 hardcoded names live.

betashop:
Pushed `5459b60d`: same dispersion port as gorilla — rotation days (the adaptive book's home turf) are now first-class: DISPERSION MODE logging with the breaking group + standout movers, `dispersionRead` on every emitted signal, dispersion + movers in every tick's result record. Telemetry now reads like the pulse skill's own output. 41+76 tests.

betashop:
**Cohort-axis hysteresis added** (`cuttlefish: cohort-axis hysteresis`). Cuttlefish was already free of the bucket set-difference defect (its closes were already per-name), but entries/rebalance keep separate 4h cohort caches, and both keyed off the same 0.40 threshold — a name straddling it could be opened by one and closed by the other. Now: entry hard-block at `leanThreshold` 0.40, close only at `reversalThreshold` 0.55 (`divergence_reversed` fires on `reversed`, not `against`). The 0.40–0.55 band is un-openable but not force-closed. `enforce_hysteresis()` guards both the score and cohort axes; shape test asserts both. 47 engine/coherence + 84 shape tests.

betashop:
Closing — restarting the design from scratch per Jason. Same two design flaws to carry forward: (1) derived universe was main-dex-only (should span xyz); (2) a 15-minute adaptive loop is wrong for a longer-horizon play — too much churn machinery. Branch retained for reference.

PR #449, opened 2026-07-04
strategy-discover: --theme worldview search + template-first routing default — v2.4.0 (for Sarvesh's review)

betashop:
**For Sarvesh** — owner of `senpi-strategy-discover`. This re-opens, as a proper PR for your review, changes to your skill that I had prematurely direct-merged (#443) — my mistake; reverted via #447 and brought here for you to approve/merge.

Two changes to `senpi-strategy-discover` (v2.2.1 → **2.4.0**):

**1. `--theme` — a SOFT worldview search** (was #443). Scores every candidate on `thesis`/`tags`/`tagline` match, expands regime synonyms (so `k-shape` also matches `long/short`, `two-speed`, `divergence`), floats matches to the top via `meta.theme_matches` — **without hard-filtering** (all survivors still returned). Fixes agents eyeballing 78 theses and missing obvious fits (e.g. Cougar for a K-shape ask). New `test_theme` suite (5/5); existing tests green.

**2. Routing default** (the template-first on-ramp). Broadened the description to claim open-ended "make me a strategy / build me a strategy / I want a strategy / get me set up" asks and made discover the default first stop — route to `senpi-strategy-author` only for an explicit from-scratch / custom-DSL design. Pairs with strategy-author v2.4.0 (merged, #448), which now offers a matching template first and hands the match to *this* skill.

Verified: `--theme "equity long short"` → cougar/cub/octopus. Tests 5/5. Your call on whether to squash the revert-of-revert commit on merge.

(Two commits: re-apply of #443, then the routing-description change.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #456, opened 2026-07-06
discover: budgets are minimums, not recommendations (size from user funds)

betashop:
## Problem
The agent recommends **fixed dollar amounts** per strategy ($400 Rhino, $200 Spider, $200 each), independent of how much the user has. Root cause in `senpi-strategy-discover`:

- `catalog.json` `min_budget` is a **floor** (`max(declared, $100 × wallets)`) — the catalog even says *"NOT a hard gate; positions scale with budget."*
- But `discover.py::_suggested_budget()` returned that floor whenever the user hadn't typed an explicit amount, and the card rendered it as **"Suggested: $X"**.
- The user's real balance **was** fetched (`meta.user_context.budget`) but never used to size — and only attached when market enrichment happened to run.

So the "suggestion" was the floor, blind to the user's funds. Exactly the bug.

## Fix
- **Engine:** rename the emitted field `suggested_budget → min_budget` (an honest floor via new `_min_budget()`); drop the intent-echo fallback.
- **Engine:** **always** attach `meta.user_context` (available funds), independent of market enrichment / candidate count, so the LLM can size from real balance.
- **SKILL.md Layer 3:** size from `user_context.budget` — use the user's stated amount if given (≥ floor), else **propose** a share of available funds per pick (each ≥ its `min_budget`, keep a buffer) and **confirm before install**; never default everyone to the floor. Card now reads **"Min to start ~$X"**; field table + card template updated.
- **strategy-ops unchanged** — it already *requires* `--budget`, splits by `funding_share`, caps at live balance; it was never the source.

## Behavior change
Before: *"Rhino — $400 suggested"* to everyone.
After: *"Rhino needs ~$400 minimum; you've got ~$X free — I'd put ~$Y here. Good?"*

## Tests
- Field renamed in tests; assert `spider min_budget == 200` (floor, not a recommendation).
- `test_discover.py` **86/86**, `test_scenarios.py` **26/26** pass.
- Live engine verified: card carries `min_budget`, no `suggested_budget`, `user_context` attached even with `--no-market`.
- MINOR bump **2.2.2 → 2.3.0** (auto-upgrades users; within manifest `maxMajor: 2`).

### DIFF
--- a/senpi-strategy-discover/SKILL.md
+++ b/senpi-strategy-discover/SKILL.md
@@ -12,7 +12,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.2.2"
+  version: "2.3.0"
   platform: senpi
   exchange: hyperliquid
 ---
@@ -96,7 +96,7 @@ Each candidate is a flat record. You rank on the soft fields; you narrate from t
 | `direction`, `funding_split` | direction match · whether it's already a multi-wallet fund (skip stacking) |
 | `market_facts` | the live "why now" for your lead |
 | `caveats` | honesty — surface **verbatim** |
-| `suggested_budget` | sizing in the card |
+| `min_budget` | the **floor** to start (`max(declared, $100 × wallets)`) — NOT a recommendation. Size the actual budget from the user's funds (`meta.user_context.budget`); see Layer 3 |
 | `id`, `version` | the handoff to ops |
 
 ## Conversation flow
@@ -148,11 +148,20 @@ otherwise keep it in your head and rank on `archetype_label`/`belief_plain`/`the
    | a specific market | "Which — a stock, pre-IPO name, commodity, index, or coin?" → `--assets`. |
    | copy | "Multi-week proven winners, live hot streaks, or specific whales?" → rank on `tags`. |
    | breakout / structural | the one drill-down that matters for that branch. |
-3. **Size & lock in (Layer 3)** — ask budget (and risk if not implied) only to size + pick the DSL
-   preset. Optionally `discover.py --context-only` to reference holdings (confirm first; never silently
-   infer). Then run the engine with the FULL concrete flag set (this run does the live market read),
-   **rank the eligible set**, and narrate 2–3 cards leading with the top pick's `market_facts` "why now";
-   surface `caveats` verbatim.
+3. **Size & lock in (Layer 3)** — pick the DSL preset and size the budget. **`min_budget` on each card is
+   the FLOOR to run it, not a recommended amount** — never just parrot it as "Suggested: $X". Size from
+   the user's **available funds** (`meta.user_context.budget`, always attached):
+   - If the user named an amount, use it (must be ≥ `min_budget`; if below, surface the floor honestly).
+   - Otherwise **propose** an amount that scales with their free balance and how many strategies they're
+     deploying — a sensible share of available funds per pick, each **≥ its `min_budget`**, leaving a cash
+     buffer — then **confirm before install** ("you've got ~$X free; I'd put ~$Y in Rhino, ~$Z in Spider —
+     good?"). Never invent a number the user hasn't confirmed, and never default everyone to the floor.
+   - If funds are unavailable (`user_context` missing/errored), ask for the budget rather than assuming.
+
+   Optionally `discover.py --context-only` to reference holdings (confirm first; never silently infer).
+   Then run the engine with the FULL concrete flag set (this run does the live market read), **rank the
+   eligible set**, and narrate 2–3 cards leading with the top pick's `market_facts` "why now"; surface
+   `caveats` verbatim.
 
 ### Always-available moves (any point, on demand)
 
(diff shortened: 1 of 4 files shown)


0xsarvesh (approved):
**Approve.** Correct diagnosis, honest fix, right base branch (`main`), correct version bump (`2.2.2 → 2.3.0`). Verified by running the PR's code — not just reading.

**Ran the tests + the engine live against prod:**
- Offline suite green: `test_discover` 86/86, `test_scenarios` 26/26, `test_cli` 54/54, `test_fullfleet` 24/24, `test_default_catalog` ✓.
- Live, all four paths work: normal run → `min_budget:100`, `suggested_budget` gone, `user_context.budget:$663.95`, no warnings; `--no-market` → user_context **still attached** (the headline fix); zero-ish candidates → still attached; `--context-only` → still returns `{user_context:{budget}}`.
- Before/after on the actual bug:
  ```
  main  --no-market:  user_context: False | suggested_budget = 100   ← floor labeled "suggested", no funds
  PR456 --no-market:  user_context: True  | min_budget      = 100   ← funds attached, honest floor
  ```
- Rename blast radius is self-contained — nothing outside discover reads `suggested_budget`. strategy-ops correctly untouched (it already requires `--budget` and caps at live balance — never the source).

**Two non-blocking nits for a follow-up (neither breaks anything):**
1. **Doubled client init.** In the market-on path there are now two `MCPClient`s — the new block does `fetch_user_context(_get_client())` and the market block still does its own `client = _get_client()`. Before, one client was reused for both. That's an extra `initialize` handshake per normal run. Hoist a single client above the user_context block and thread it into both fetches. (`--no-market` is correctly one client — no issue.)
2. **No test guards the headline behavior.** "user_context always attached, incl. `--no-market`" is only manually verified. The suite already has fixture-client patterns — a tiny test asserting `meta.user_context` present on a `--no-market` run would stop a silent regression back to this exact bug.

**Merge-order note:** two open discover PRs bump the version — this one → 2.3.0, #449 → 2.4.0. Land this first (clean, on main), then #449 rebases on top.

Nice fix — the "size from real funds, confirm before install, never parrot the floor" Layer-3 guidance is the right behavior.

0xsarvesh (approved):
**Approve.** Correct diagnosis, honest fix, right base branch (`main`), correct version bump (`2.2.2 → 2.3.0`). Verified by running the PR's code — not just reading.

**Ran the tests + the engine live against prod:**
- Offline suite green: `test_discover` 86/86, `test_scenarios` 26/26, `test_cli` 54/54, `test_fullfleet` 24/24, `test_default_catalog` ✓.
- Live, all four paths work: normal run → `min_budget:100`, `suggested_budget` gone, `user_context.budget:$663.95`, no warnings; `--no-market` → user_context **still attached** (the headline fix); zero-ish candidates → still attached; `--context-only` → still returns `{user_context:{budget}}`.
- Before/after on the actual bug:
  ```
  main  --no-market:  user_context: False | suggested_budget = 100   ← floor labeled "suggested", no funds
  PR456 --no-market:  user_context: True  | min_budget      = 100   ← funds attached, honest floor
  ```
- Rename blast radius is self-contained — nothing outside discover reads `suggested_budget`. strategy-ops correctly untouched (it already requires `--budget` and caps at live balance — never the source).

**Two non-blocking nits for a follow-up (neither breaks anything):**
1. **Doubled client init.** In the market-on path there are now two `MCPClient`s — the new block does `fetch_user_context(_get_client())` and the market block still does its own `client = _get_client()`. Before, one client was reused for both. That's an extra `initialize` handshake per normal run. Hoist a single client above the user_context block and thread it into both fetches. (`--no-market` is correctly one client — no issue.)
2. **No test guards the headline behavior.** "user_context always attached, incl. `--no-market`" is only manually verified. The suite already has fixture-client patterns — a tiny test asserting `meta.user_context` present on a `--no-market` run would stop a silent regression back to this exact bug.

**Merge-order note:** two open discover PRs bump the version — this one → 2.3.0, #449 → 2.4.0. Land this first (clean, on main), then #449 rebases on top.

Nice fix — the "size from real funds, confirm before install, never parrot the floor" Layer-3 guidance is the right behavior.

0xsarvesh (approved):
**Approve.** Correct diagnosis, honest fix, right base branch (`main`), correct version bump (`2.2.2 → 2.3.0`). Verified by running the PR's code — not just reading.

**Ran the tests + the engine live against prod:**
- Offline suite green: `test_discover` 86/86, `test_scenarios` 26/26, `test_cli` 54/54, `test_fullfleet` 24/24, `test_default_catalog` ✓.
- Live, all four paths work: normal run → `min_budget:100`, `suggested_budget` gone, `user_context.budget:$663.95`, no warnings; `--no-market` → user_context **still attached** (the headline fix); zero-ish candidates → still attached; `--context-only` → still returns `{user_context:{budget}}`.
- Before/after on the actual bug:
  ```
  main  --no-market:  user_context: False | suggested_budget = 100   ← floor labeled "suggested", no funds
  PR456 --no-market:  user_context: True  | min_budget      = 100   ← funds attached, honest floor
  ```
- Rename blast radius is self-contained — nothing outside discover reads `suggested_budget`. strategy-ops correctly untouched (it already requires `--budget` and caps at live balance — never the source).

**Two non-blocking nits for a follow-up (neither breaks anything):**
1. **Doubled client init.** In the market-on path there are now two `MCPClient`s — the new block does `fetch_user_context(_get_client())` and the market block still does its own `client = _get_client()`. Before, one client was reused for both. That's an extra `initialize` handshake per normal run. Hoist a single client above the user_context block and thread it into both fetches. (`--no-market` is correctly one client — no issue.)
2. **No test guards the headline behavior.** "user_context always attached, incl. `--no-market`" is only manually verified. The suite already has fixture-client patterns — a tiny test asserting `meta.user_context` present on a `--no-market` run would stop a silent regression back to this exact bug.

**Merge-order note:** two open discover PRs bump the version — this one → 2.3.0, #449 → 2.4.0. Land this first (clean, on main), then #449 rebases on top.

Nice fix — the "size from real funds, confirm before install, never parrot the floor" Layer-3 guidance is the right behavior.

PR #462, opened 2026-07-08
strategy-debug hardening: turn a live-deploy failure into deterministic gates

betashop:
## Why

A hand-authored strategy (**divergence-play**, 2 legs, \$4k) deployed **ACTIVE + funded** but never traded, and an operator agent couldn't get it live across many turns and **four redeploys** — because the skills gave *capability* but **no diagnostic path, no gate, and no clear definition of done**. It looped on cheap-to-check things (does the file exist? indentation?), declared "done" three times prematurely (ACTIVE → signal fired → positions opened), and nearly left a naked short by misreading a DSL state file. Every one of those failures is now a **tool that names the cause** or a **gate that makes the mistake impossible** — validated against the real incident end to end.

## The incident → the fix

| What actually went wrong | What now catches it |
|---|---|
| `interval_seconds: 0` → scanner never scheduled (agent saw the clue, misread it as "runtime not wired up") | `validate_strategy` rejects interval ≤ 0; `diagnose.py` verdict `interval-zero` |
| `scan()` threw (missing `dex` arg on `xyz:` reads) → deployed + funded anyway | **pre-fund smoke gate** — runs `scan()` before funding, refuses on throw/bad-shape |
| missing `import sys` → the `except` handler itself crashed (NameError), masking the real error | static **`sys`-import lint** in `validate_strategy` (no token needed) |
| `marginPct: 0.18` (decimal) vs `strategy.margin_pct: 18` → sized **\$3.60 on \$2k**, schema-valid so nothing flagged it | **sizing-sanity** check names "emit 18, not 0.18"; scan-contract + author invariant |
| agent declared "done" at each surface-green step | **5-point Definition-of-done** (registered · ticks · schema-valid signal · sized ≈ intended % · DSL-protected · no orphans) |
| naked-position scare: agent read a **DSL state file**, saw `active:false`, flip-flopped "NAKED"↔"closed" | **`protect.py`** reconciles the **chain** vs tracked vs stops → PROTECTED/NAKED; doc trap: archived ≠ closed |
| **6 copies** of the package on disk; edits kept landing in a copy the runtime didn't load (stale fix persisted) | **duplicate-copy guard** — `deploy`/`diagnose` warn which copy is in use |
| the validator that would've caught the bad field **couldn't run** (imported PyYAML; PEP-668 host blocks pip) | `validate_strategy` falls back to vendored `_yaml` — runnable everywhere |

## What's in the PR

**New tools** (`senpi-strategy-ops/scripts/`)
- **`diagnose.py`** — "why isn't it trading?" doctor; leads with the runtime's own `senpi scanner` BARREN RPC; `--run-scan` runs `scan()` live and shows the literal return + schema + sizing warnings.
- **`protect.py`** — "are my positions protected?" doctor; reconciles clearinghouse × DSL-tracked × resting stops → PROTECTED/NAKED/STOP-NOT-ON-VENUE; refuses to call anything protected unless the chain agrees.
- **`_smoke.py`** — isolated (subprocess + timeout) `scan()` runner enforcing the read-only mutation boundary (a smoke test can never place a trade); classifies clean/empty/threw/bad-return/bad-shape + sizing warnings. Shared by the deploy gate and `diagnose --run-scan`.

**Changed**
- **`deploy.py`** — pre-fund smoke gate (`--no-smoke` escape); `verify` liveness uses `alive`/invocations not `runCount` (which counts *emits*, so a quiet scanner read "never ticked"); duplicate-copy warning.
- **`validate_strategy.py`** — PyYAML→vendored fallback; assert external_scanner fields (interval>0, path/entrypoint/schema/validity); `sys`-import lint.
- **Docs/DoD** — `troubleshooting.md` (new symptom→cause→fix table); `dsl-protection-check.md` (the archived-DSL trap); `scan-contract.md` (marginPct is a percent); ops/author/runtime SKILL.md (5-point DoD + anti-patterns). Versions: ops `2.4.0`, author `2.5.0`, runtime `3.0.3`.

## Design deviations from the original spec (deliberate)
- The smoke gate **allows an empty result** — blocking "0 signals" would break Rhino/Ox/regime allocators that correctly emit nothing on a cold tick. It blocks only unambiguous defects (throw / non-list / bad-shape).
- Sizing + duplicate-copy issues are **loud warnings, not hard blocks** — both have legitimate (if rare) forms, and a false block on the money-path is worse than a warning the DoD backstops.

## Tested
- Every `_smoke` branch incl. the safety ones (scan calling `create_position` → blocked `PermissionError`; mutating `ctx` → frozen guard).
- `diagnose` static causes + `--run-scan` on a real scanner; `protect.reconcile` on the exact HYPE(naked)/META(protected) end-state; the sizing catch on the exact 0.18-vs-18 case.
- The gate refuses throwing/bad-shape and passes clean-but-quiet; dupe-guard detects a two-copy layout with no false positive on the normal one.
- `validate_strategy` runs with PyYAML blocked; **all 83 catalog packages still pass** (zero false positives) after every static addition.

## Out of scope (runtime-team item)
The runtime archiving a position on breach **before confirming the close order filled** is what *creates* the naked window (breach → cancel SL → maker close doesn't fill → archived + open + unprotected). That's a runtime-plugin fix, filed separately — this PR makes it *detectable* (`protect.py`), not impossible.

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -13,7 +13,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.5.0"
+  version: "2.6.0"
   platform: senpi
   exchange: hyperliquid
   requires:
@@ -192,6 +192,21 @@ the catalog entry, then unit-test → validate → hand to smoke-test."* Then ti
    status ACTIVE" is **not** done — `ACTIVE` only means `position_tracker` runs; done means a scanner tick
    that **emitted a schema-valid signal**. Never declare a strategy fixed/built without that tick, and
    never substitute a raw `strategy_create_custom_strategy` (it drops the DSL and automated exits).
+10. **Deploy, then confirm it's LIVE *and* PROTECTED — the job is NOT done until this passes.** Authoring a
+    clean package is not the finish line. A strategy the user can't see running — or one running
+    **unprotected** — is the exact failure this flow exists to prevent (a user lost real money on a
+    strategy that funded but never got a runtime, so it had **no DSL and no guardrails**, and nothing said
+    so). Hand to `senpi-strategy-ops` (`deploy.py create <id> --budget <usd>` → `deploy.py runtime <id>`),
+    then verify the **full** definition of done — never stop at "ACTIVE":
+    - **Live** — `python3 senpi-strategy-ops/scripts/diagnose.py <id>`: the external_scanner is registered,
+      `interval_seconds > 0`, has **ticked**, and emitted a **schema-valid signal sized ≈ the intended %**.
+    - **Protected** — `python3 senpi-strategy-ops/scripts/protect.py <id>`: every open position is
+      **DSL-tracked with a resting stop** (not NAKED).
+    - **Guard-railed** — `openclaw senpi risk -r <id>-<inst>`: the `risk.guard_rails` (drawdown halt,
+      daily-loss, per-day / per-asset limits) are loaded and eligible.
+    If the runtime never registered (gateway scope, a bad path, a raw-create shortcut), the strategy is
+    **funded but running with NO runtime — no DSL, no guardrails.** Say so plainly; **never report a
+    strategy as live or protected until these three confirm it.**
 
 Report each numbered stage as it lands — a short line is enough. The point is the user sees forward motion
 the whole way and can catch a wrong turn early, instead of after the entire package is already built.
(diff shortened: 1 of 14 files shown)


betashop:
Closing — superseded by #467, a focused ~350-line rework (this PR at ~1,750 was too broad). #467 keeps the load-bearing pieces — the deploy **liveness gate** (runtime + scanner + **DSL-wired** + budget) and the DSL-exit-block deploy refusal — and drops the heavier diagnose/smoke/monkeypatch machinery. Branch kept for reference/salvage.

PR #464, opened 2026-07-09
Guardrail: validate coin symbols vs live instruments before funding (xyz: prefix)

betashop:
**Prevents the M404726 incident class:** an agent created custom strategies with coins as bare `NVDA`/`AVGO`/`SKHX` instead of `xyz:NVDA`/… → every position rejected → 3 strategies marked `FAILED` **after funding** → ~$6,700 parked in sub-wallets. (Funds were never lost — they sit `withdrawable` on-chain; `FAILED` is a record status, not an on-chain freeze.)

The hole: `validate_universe` is deliberately lenient (accepts bare `NVDA` if `xyz:NVDA` is live, assuming the scan code prefixes it), and the raw `strategy_create_custom_strategy` path runs through no deploy gate at all. Nothing validated the **actual coin string** sent to Hyperliquid.

## What this adds
- **Deploy path (enforced):** the #462 pre-fund smoke gate now cross-checks every **emitted** `signal.asset` against the live instrument list and **BLOCKS before funding** a scan that emits a bare `NVDA` (`did you mean 'xyz:NVDA'?`). Best-effort — no token/network → skip, never false-block.
- **`validate_universe`:** strict `unknown_symbols()`/`suggest_symbol()` (exact-match) alongside the lenient input checker, + a `--coins xyz:NVDA,AVGO` **raw-create preflight**.
- **Docs:** new `references/coin-symbols.md` (canonical rule + *estimate-before-create* for the raw path + "FAILED funds are on-chain, not locked"), an ops anti-pattern bullet, and a `scan-contract.md` emitted-asset callout.
- **Tests:** +5 (strict vs lenient split; emitted-asset helper; smoke blocks bare / passes prefixed). **16/16.**

## Not covered here (raw MCP has no skill-side enforcement)
The one *hard* guard for the raw `strategy_create_custom_strategy` path is the **MCP tool description** itself — recommend requiring an `estimate_custom_strategy_positions_opening` dry-run + naming the FAILED-funds-parked consequence. Cross-team; flagged separately.

> ⚠️ **Stacked on #462** (`strategy-debug-hardening`) — it extends that PR's smoke gate. Base is set to `strategy-debug-hardening`; merge #462 first, or retarget to `main` after. Skill bump: strategy-ops 2.4.0 → 2.5.0.

### DIFF
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -18,7 +18,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.4.0"
+  version: "2.5.0"
   platform: senpi
   exchange: hyperliquid
   requires:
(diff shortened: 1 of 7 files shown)


betashop:
Closing — part of the #462/#464 rework (#467). Note: #467 focuses on the deploy **loop + budget** gates and does **not** re-land this coin-symbol / `xyz:`-prefix validation. The existing main `validate_universe` preflight still blocks unknown tickers before funding, but the strict literal-string `xyz:` check from this PR is a separate follow-up (tracked). Branch kept for salvage.

PR #465, opened 2026-07-09
improve-trades v1.2.0 — total PnL (open book) + honesty guardrails

betashop:
Fixes the distorted (and in one spot **dangerous**) trade review we saw on M404726 — where the skill judged on a **partial ledger**, headlined a **hindsight counterfactual**, **diagnosed exit calibration on 100%-UNKNOWN attribution**, reported **"no leaks / all clear" while telemetry was down**, and **fabricated aggregates**.

## Root cause
The engine only read **closed** trades — blind to the live book's **unrealized** PnL. So it (a) called a book riding open winners a "loser," (b) leaned on the hindsight "if I'd held to now" counterfactual *because* it couldn't see the open positions, and (c) that counterfactual drove "you left $2,400 on the table / 70% premature" → implicitly *hold longer / loosen stops* on a leveraged, losing book.

## Engine (`review.py`)
- **WS1 — total ledger:** `fetch_open_book` pulls each **current** wallet's clearinghouse → `strategies[]` gain `unrealized_pnl` / `total_pnl` / `open_positions[]`; new top-level `pnl_summary {realized, unrealized, total, realized_by_book}`. Fail-open → `unrealized: None` (UNKNOWN), never a fake 0. The **open book is now the real "are winners running" evidence**, so the skill no longer relies on the counterfactual.
- **WS2 — neutral labels:** `exit_vs_hold` `beat/worse → exit_ahead/held_higher` (a model reads "worse" as "you did badly").
- **WS3 — undetermined ≠ all-clear:** new `telemetry_availability {status, streams_computed, exit_attribution}` — `streams_computed:false` when telemetry's down + `undetermined` when 0 exits attributed.
- **WS4 — anti-fabrication:** emits the `realized_by_book {current, closed}` split so the narrator quotes it (no more invented −$405).
- Wired through `run()` + the `timing`/`strategies`/`telemetry` steps (parity preserved).

## SKILL.md (binds a small model)
- **HARD RULES** block at the top (5 one-liners): lead with TOTAL PnL · counterfactual isn't a verdict · undetermined ≠ all-clear · quote engine numbers · strategy-not-user.
- **Guardrail 1 made symmetric** — a *positive* counterfactual is **not** "money left on the table," `held_higher` is **not** "premature," and the open book is the letting-winners-run evidence. **Guardrail 6** gains the undetermined rule. Output-shape / contract / quick-actions updated.

## Tests
**42/42.** +4: open-book unrealized/total; open-book-unreadable → UNKNOWN not 0; telemetry-down → `undetermined` (not all-clear); partial. Skill bump **1.1.1 → 1.2.0** (auto-upgrade; bump the external skills-manifest on merge).

**Acceptance:** re-run against M404726 and confirm the review (a) leads with **total** PnL, (b) never says "premature / left on the table," (c) reports exit-quality **UNDETERMINED** (telemetry was down), (d) all numbers reconcile.

### DIFF
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -39,7 +39,11 @@ more" questions; use `senpi-portfolio` for live state.
 ## HARD RULES (never violate — obey these even if you skim the rest)
 
 1. **Lead with TOTAL PnL** (`pnl_summary.total` = realized + unrealized), never realized alone. Realized-only
-   is half the ledger — it calls a book riding open winners a "loser" and penalizes hold-strategies.
+   is half the ledger — it calls a book riding open winners a "loser" and penalizes hold-strategies. **If
+   `pnl_summary.unrealized_partial` is true (or `unrealized_coverage.read < .current_strategies`), TOTAL is a
+   FLOOR, not a complete number** — some current wallets couldn't be read. Say **"at least $X (N of M wallets
+   readable)"**, never present it as the finished total. (`total: null` = the open book was *entirely*
+   unreadable → UNKNOWN, not 0 and not a floor.)
 2. **The hold-to-now counterfactual is CONTEXT, never a verdict.** `held_higher` / a positive
    `if_all_reclosed_now_total` just means the asset kept running THIS window (hindsight; it ignores the risk
    the exit avoided). NEVER say "you exited too early / N% premature / left $X on the table," and NEVER let it
@@ -196,6 +200,9 @@ These are non-negotiable. Each fixes a real failure from live agent responses to
 
 **Lead with `pnl_summary.total`** (realized closed + unrealized open) — NOT `realized_pnl_total` alone.
 Realized-only is half the ledger: it calls a book riding open winners a "loser" and penalizes hold-strategies.
+**Degraded-read honesty (same principle as `undetermined ≠ all-clear`):** if `pnl_summary.unrealized_partial`
+is true (some current wallets read, some didn't), `total` is a **FLOOR** — headline it as *"at least $X (N of M
+wallets readable)"*, never as a complete total. `total: null` = the open book was entirely unreadable → UNKNOWN.
 Then give the aggregate exit counts (`timing_summary`: `exits_ahead` / `exits_held_higher` / flat). Only after
 the aggregate may you mention a single trade, and only as *one data point*, never the headline.
 
@@ -398,7 +405,8 @@ trades[]      per CLOSED trade (from strategies of ALL statuses — a churned bo
 pnl_summary      TOTAL LEDGER — LEAD WITH THIS (realized closed + unrealized open):
   realized, unrealized (None = UNKNOWN read, not 0), total (None when unrealized UNKNOWN),
   realized_by_book{ current, closed },      # quote this split — never re-derive a closed-book figure
-  unrealized_coverage{ read, current_strategies }
+  unrealized_coverage{ read, current_strategies },
+  unrealized_partial   # true → some wallets UNKNOWN; unrealized/total are a FLOOR ("at least $X, N of M") — HARD RULE 1
 
 telemetry_availability   the 'undetermined ≠ all-clear' signal — READ IT FIRST (guardrail 6):
   status ∈ available | partial | undetermined | no_trades,
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -17,7 +17,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.1.1"
+  version: "1.2.0"
   platform: senpi
   exchange: hyperliquid
 ---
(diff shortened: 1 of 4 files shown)


0xsarvesh (changes requested):
Strong PR — correct root cause, and I verified the important parts by running it rather than reading it. **Requesting one change** before merge: the partial-coverage path reintroduces the exact failure class this PR exists to fix.

## Verified ✅

- **MCP call + parsing are right, not guessed.** `strategy_get_clearinghouse_state(strategy_wallet=…)` and the `main`/`xyz` → `assetPositions[].position.{szi,unrealizedPnl,returnOnEquity,entryPx,positionValue,leverage.value}` shape match the **live prod payload** I probed, and mirror `portfolio.py` (which reconciles to the cent live). Summing uPnL across dexes is correct — positions are disjoint per dex.
- **Fixture is real-shape** — this avoids the fabricated-fixture disease that bit account-status and trader-research.
- **All-fail honesty holds**: `unrealized: None`, `total: None`, never a fake 0, and `test_open_book_unreadable_is_unknown_not_zero` guards it.
- **Live run clean**: `pnl_summary` + `telemetry_availability` present and sane; `streams_computed:false` with the explicit "UNDETERMINED, not zero — do NOT diagnose exit calibration."
- 42/42 tests, compiles, correct base + bump (1.1.1 → 1.2.0).

The design calls are genuinely good: `beat/worse → exit_ahead/held_higher` (a model reads "worse" as a grade), the `realized_by_book` split so the narrator *quotes* instead of re-deriving, and HARD RULE 2 killing the dangerous "hold longer / loosen stops" inference. The open book as the real letting-winners-run evidence is the right insight.

## 🔴 Requested change — partial coverage silently understates TOTAL

`_pnl_summary` sums only the readable wallets. All-fail → `None` (correct). **Partial → a partial sum presented as a complete total.** Simulated directly:

```
wallet A readable (+$500 uPnL), wallet B read FAILED
→ unrealized: 500.0   (B's uPnL silently missing)
→ total:      650.0   (presented as COMPLETE)
→ coverage:   {read: 1, current_strategies: 2}
```

**HARD RULE 1 says "lead with TOTAL" unconditionally**, and `unrealized_coverage` appears in SKILL.md only in the output-shape reference block — bound to no rule or guardrail. So one wallet blip on a multi-wallet user (exactly the M404726 profile) → the model confidently headlines an understated total as if it were complete. That is "judged on a partial ledger," relocated from closed-only to partially-read-open-book. **No test covers the partial path** — the existing one drops the sole clearinghouse entry, so it exercises all-fail.

### And there are no retries anywhere in this path

- `mcp_client.mcp_call` is a **single-shot POST**; it raises on HTTP ≥ 400 and (since the strict-error client) also on JSON-RPC errors, `result.isError`, and app-level `{"success": false}`.
- `review.py` wraps it in a bare `try/except` → **one failure = that wallet is `None` for the whole run.**
- Nothing in the fleet retries MCP calls. (The only `retry` grep hit, `ops/_cli.py:214`, is the openclaw CLI gateway, not MCP.)
- This read runs **parallel across up to 8 wallets on a 12s timeout** — precisely where transient 429/500/timeout live. The strict-error client means *more* conditions now raise than before, so the `None` path is **more reachable** than the code's surface suggests.

Net: partial coverage isn't a pathological edge case, it's a single blip on any one of N wallets.

## What I'd like to see

**1. Add bounded retries** on the clearinghouse read — 2 attempts, short backoff, **transport/5xx/timeout only**. Never retry a legitimate `{"success": false}` (that's a real answer, not a blip). Reasonable to scope it to this read; a fleet-wide read-retry in `mcp_client` is a bigger conversation.

**2. But design around failure — retries reduce frequency, they never remove it.** The invariant I care about: **when the engine cannot produce a correct answer, the LLM must know that.** A number the model can't distinguish from a complete one is worse than no number. Concretely:

- Engine emits `"unrealized_partial": len(known) < len(strat_reads)` (and keep the coverage counts).
- **HARD RULE 1 gains a clause:** *"If `unrealized_coverage.read < .current_strategies` (or `unrealized_partial`), TOTAL is a **floor**, not a total — say 'at least $X (N of M wallets readable)'. Never present it as complete."*
- Add a partial-coverage test (one wallet readable, one `None`) asserting the flag and that `total` is labeled a floor.

That's the same principle you already applied beautifully to telemetry (`undetermined ≠ all-clear`) and to the all-fail case (`None`, not 0) — this PR just needs it extended to the *partially*-known case. Every degraded read should surface as UNKNOWN / PARTIAL / FLOOR in the payload, so the narrator can never mistake a degraded number for a good one.

Happy to re-review as soon as that's in — everything else here is ready to land.

betashop:
Both changes are in (`81cad2e`) — thanks for the rigorous review; you caught the exact relocation of the failure class.

**1. Partial coverage → FLOOR, flagged (no more silent partial-as-complete).**
- `_pnl_summary` now emits `unrealized_partial = read < current_strategies` alongside `unrealized_coverage`. Your simulation now returns `unrealized: 500, total: 650, unrealized_partial: true, coverage: {read:1, current_strategies:2}`.
- **HARD RULE 1 gained the clause:** *"if `unrealized_partial` (or `read < current_strategies`), TOTAL is a FLOOR — say 'at least \$X (N of M wallets readable)', never present as complete. `total: null` = entirely unreadable → UNKNOWN, not 0 and not a floor."* Mirrored into guardrail 1 and the output-shape block, so it's bound to a rule, not floating.

**2. Bounded retry, transient-only.**
- `fetch_open_positions` retries the clearinghouse read — 2 attempts, short backoff — via `_is_transient`: **transport / 5xx / 429 / timeout only.** A definitive answer (`{success:false}`, other 4xx, JSON-RPC error, tool `isError`) is never retried. Scoped to this read, not `mcp_client`.
- Per your framing: retries only reduce the `None` frequency, they never remove it — so a still-failed read stays UNKNOWN and partial stays flagged. The invariant holds: every degraded read surfaces as UNKNOWN / PARTIAL / FLOOR in the payload.

**Tests:** +6 (partial flag + floor value; full-coverage & all-fail controls; `_is_transient` policy; retry-then-succeed; no-retry-on-definitive). **48/48 pass.**

Ready to merge — re-review whenever.

0xsarvesh (approved):

PR #467, opened 2026-07-09
strategy deploy: gated create→runtime→verify loop + honor requested budget

betashop:
## Problem
Two deploy gaps behind recent incidents:
1. **A strategy could be \`registered\` but not actually live.** \`verify\` only checked the scanner had ticked — nothing confirmed the **DSL was wired** or the **budget landed**. Funded-but-no-DSL and FAILED/idle deploys slipped through (M404726).
2. **\`create\` silently under-funded.** \`plan_funding\` scaled every wallet down to the live balance, floored at $100 — so a "$1,000 / $2,000" request could quietly become two $100 wallets and still report success.

## Solution — clean rework of #462/#464 (~350 lines vs #462's ~1,750)
- **\`verify\` is now THE liveness gate.** \`live\` only when *every* instance is **runtime-running + scanner-active (ticked or scheduled, never erroring) + DSL-wired (\`exit.dsl_preset\` present AND monitor enabled) + funded to request**. It does **not** wait for a slow first tick (a \`scheduled\` scanner passes) — a fast required step, not an optional poll. Returns \`not-live\` naming the failing component (\`scanner=broken\`, \`dsl=config-missing\`, \`budget=underfunded\`).
- **\`create\` honors \`--budget\` as a hard target.** If the balance can't cover it, it **HALTS \`underfunded\`** with the exact shortfall (fund more / lower the ask) — never silently funds less. Records the requested amount; reconciles against actual funded at verify.
- **\`_pkg.validate\` refuses a DSL-less package** (defense-in-depth vs funded-but-no-DSL). **83/83** existing templates still pass.
- **SKILLs** (strategy-ops, strategy-author) document the mandatory **create → runtime → verify** loop — *not live until verify passes*.

## Testing
- \`tests/test_deploy_gates.py\` — 24 hermetic tests (funding halt, has_dsl, scanner/dsl/budget verdicts). All pass.
- Regression: 83/83 packages pass the new DSL gate; 0 false-fails.

## Merge checklist
- Bump external **skills-manifest**: \`senpi-strategy-ops\` → 2.4.0, \`senpi-strategy-author\` → 2.5.0 (MINOR = auto-upgrade).
- Supersedes #462 + #464 (being closed).

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -13,7 +13,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.4.2"
+  version: "2.5.0"
   platform: senpi
   exchange: hyperliquid
   requires:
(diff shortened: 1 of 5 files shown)


dmcr:
**Budget gate: don't merge as-is — the precheck false-blocks valid deploys.**

`available_usd` reads only `total_in_hyperliquid` (free HL **perp** USDC). But strategy funding draws on three buckets in a cascade — perps → HL spot → EVM USDC across 7 chains (`strategy_manager/utils.ts:468`). Verified against the backend resolver (`strategyService.ts:206`) and `createFundingPlan` (`utils.ts:488`: `totalAvailable = perps + spot + Σ EVM≥min`). So a user funded on spot or any non-Base chain gets HALTed `underfunded` before `create` is even called. A portfolio-based precheck *can't* be made correct here: portfolio visibility is Base-only by default, but funding reach is 7 chains (the contract's own example: `$100 Base + $50 BSC + $10 Arbitrum` funds `$155`).

The key part: this check is **redundant with a correct one that already exists**. `strategy_create_custom_strategy` already computes true multi-chain availability and throws `SERR037: Insufficient USDC ... required X, available Y` (`strategy_manager/resolvers/strategy.ts:101`).

**Fix (keeps the good intent — no more silent scale-down-to-floor):**
1. Remove the `available_usd` / `plan_funding` shortfall precheck + `underfunded` HALT. Split by `funding_share`, no `available` cap, pass full amounts to `strategy_create_custom_strategy`.
2. Map `SERR037` → `underfunded` in the existing `MCPError` catch (`deploy.py:243`, currently generic `failed`), surfacing the backend's real `required`/`available`.
3. Update `test_deploy_gates.py` — the `PlanFunding` shortfall tests assert the behavior being removed; replace with a `SERR037 → underfunded` test.

The liveness gate, `has_dsl` defense, and create→runtime→verify docs are good — leave them.

---
_Low-priority follow-up (separate PR, different repo — not blocking this one):_ the MCP guide `hyperliquid_mcp/.../account-get-portfolio.resource.ts:28` defines `total_in_hyperliquid` as "value of open perpetual positions" — the resolver proves it's free withdrawable perp USDC. Doc drift worth fixing on its own.

betashop:
Folded into #476 (the consolidated deploy PR) — your verify-as-gate, budget-hard-target, and no-DSL refusal are merged there with commits/authorship preserved. Safe to close in favor of #476.

betashop:
Folded into #476 (consolidated deploy PR) — verify-as-gate + budget-hard-target + no-DSL refusal merged there with commits preserved. Closing in favor of #476.

PR #471, opened 2026-07-13
fleet-new: 4 strategies — raven, starling, ant (deployable) + crane (blocked) + catalog gen skip

betashop:
Four new strategy packages on the tier-3 (imperative/stateful) frontier, plus a generator upgrade so non-deployable packages stay out of the user catalog. Each is a self-contained package (`strategy.yaml` + `main/runtime.yaml` + `scanners/{scan,scoring}.py` + `tests/`), fleet-standard throughout (marginPct sizing, leverage caps, mandatory DSL, drawdown gate, read-only single-pass scanners).

## Deployable (in catalog · tested · `_pkg.validate` ✓)

**`raven` — Self-Calibrating Momentum** (7 tests)
Bison's momentum engine (ported verbatim) under a self-tuner: on a slow clock it reads its **own** closed-trade record (`discovery_get_trader_history`), rolls up win-rate / profit-factor / losing-streak, and ratchets the entry-score floor + a conviction size-scale within bounded rails. **Holds below `minTrades`** — never tunes on noise. Uses the granular DSL ladder (breakeven-first, no back-loaded gap).

**`starling` — Smart-Money Rotation Follower** (9 tests)
Derives a proven-profitable cohort (paginated top-traders ≥ $1M realized), snapshots their books, and opens **only on fresh/rising consensus** (a true snapshot-diff — never stale standing consensus), sized by agreement count. Open-only + DSL exit; held + recent-signal dedup prevent churn as consensus climbs. Preserves the `xyz:` venue prefix on emitted names.

**`ant` — Funding Harvester** (6 tests)
The best Senpi-achievable cash-and-carry. Senpi is perps-only (no HyperEVM spot), so ant runs the achievable half: **short** top-OI names paying high positive funding — but **only when the long crowd is exhausted** (overbought *and* stalling / rolling over), never a name still ripping. Low leverage (2–4×), tight 8% squeeze-stop, 24h rotation. **Directional, not delta-neutral** — the missing spot hedge is the residual risk, documented in [`ant/NOTES.md`](strategies/ant/NOTES.md).

## Not deployable — tracked + tested, excluded from the catalog

**`crane` — Managed Pairs / Stat-Arb** (7 tests)
Market-neutral alpha engine is complete and tested (log-spread → z-score → pair state machine, with `CLOSE_NAKED` as the unconditional first branch), but it **ships disarmed**: the runtime has no coordinated multi-leg close, so a DSL-only pair would leave a **naked directional leg** when one side stops first. [`crane/DESIGN.md`](strategies/crane/DESIGN.md) documents the exact gap and the one primitive that unblocks it. Concrete evidence that market-neutral needs a capability neither the runtime nor the strategy-spec proposal covers yet.

## Manifest / generator
- `gen_catalog.py` now **skips packages flagged `catalog.status: blocked`** — the catalog is the *deployable* registry, so crane is tracked/tested in-repo but never surfaces as installable.
- `catalog.json` regenerated (both locations). Also **syncs a pre-existing stale entry**: `spider` 6.0.0 → 6.0.1 (its `strategy.yaml` was already 6.0.1; the committed catalog hadn't been regenerated). Purely additive otherwise — `raven`/`starling`/`ant` added, nothing removed, only `sort_order` renumbering on existing entries.
- All declared catalog terms reuse **existing** glossary vocabulary — zero new warnings.

## Verification
`raven` 7 · `starling` 9 · `ant` 6 · `crane` 7 engine tests — all pass; `py_compile` clean; `_pkg.validate` ✓ for all four; catalog regenerates warning-clean.

**One caveat before funding any of these:** the self-tune / cohort / funding reads (`discovery_get_trader_history`, `discovery_get_trader_state`/`top_traders`, `market_get_funding_history`) use tolerant multi-key extraction + hold-on-zero, but their payload shapes **weren't live-verified** (the prod token went invalid mid-build). Worth one real-payload check before funding.

### DIFF
--- a/senpi-strategy-discover/SKILL.md
+++ b/senpi-strategy-discover/SKILL.md
@@ -15,7 +15,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.4.0"
+  version: "2.5.0"
   platform: senpi
   exchange: hyperliquid
 ---
--- a/senpi-strategy-discover/catalog.json
+++ b/senpi-strategy-discover/catalog.json
@@ -1,6 +1,6 @@
 {
   "_version": "2.1",
-  "_updated": "2026-06-17",
+  "_updated": "2026-07-13",
   "_generated": true,
   "_instructions": "GENERATED from each strategy package's strategy.yaml — do not hand-edit; run senpi-trading-runtime/scripts/gen_catalog.py. Agents read this via senpi-strategy-discover (scripts/discover.py). DECLARED fields (archetype, sub_style, asset_classes, asset_scope, risk_level, tier, belief_plain, direction) are author-set in strategy.yaml `catalog:`; DERIVED fields (assets, leverage_max, funding_split, instance_count, cadence_seconds, time_horizon, max_slots) are computed from instances/params. Labels/glosses come from senpi-strategy-discover/references/glossary.yaml. 'min_budget' is the effective floor (max(declared, 100 x instance_count)); positions scale with budget — NOT a hard gate. A strategy is a deployable package; install via senpi-strategy-ops.",
   "skills": [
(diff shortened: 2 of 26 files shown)


betashop:
Follow-up: bumped **senpi-strategy-discover 2.4.0 → 2.5.0** so users auto-upgrade and the picker surfaces `raven`/`starling`/`ant`. The bundled `senpi-strategy-discover/catalog.json` was already regenerated (identical to `strategies/catalog.json`, 86 skills) — the version bump is the auto-upgrade trigger (no separate manifest in this repo; confirmed against prior bump history). 20 discover tests pass.

0xsarvesh (approved):
Reviewed for functional + structural correctness. All 4 packages: py_compile clean, unit tests pass, validate_strategy/validate_universe ✓, scan() drives end-to-end (rich+empty payloads) without crashing and emits invariant-valid signals. Catalog regenerates in sync (both locations), crane correctly excluded as blocked. Approving.

PR #473, opened 2026-07-15
fix(improve-trades): closed-strategy discovery gap ≠ no trades; account value ≠ P&L (v1.3.0)

betashop:
## The failure (live)

User: *"Improve my last 10 trades. Did I sell too early or late?"* — on an account whose **Lion strategies were closed**. The agent answered *"there are zero closed trades to review… the question can't be answered."* Then, pushed, it read a `$0` account-value curve as *"the account drained to $0 — grinding losses"* — but `$0` was the **user closing the strategy (funds returned)**, and realized PnL was only **−$20.87**.

## Root cause — it's baked into the engine's data source

- `review.py` sources closed trades **only** from `discovery_get_trader_history` (review.py:706), which returns **EMPTY for closed/torn-down strategies**. No fallback to `strategy_get_pnl_and_account_value_history`.
- So `trades[]` is empty → the engine reports no trades → **guardrail 9** (*"no closed trades = complete, correct result: nothing to review, stop"*) fires → the agent confidently says "zero trades."
- The SKILL.md model was actually **wrong about this**: the Sources note claimed that for a closed strategy *"the engine fails open to discovery: the trades are still listed."* Discovery does **not** still list them.
- When the engine returns empty, the agent bypasses to raw `pnl_history` and narrates the **value curve as trades** — and **no guardrail** said a `$0`-after-close is a withdrawal, not a loss. Mistake #2 was completely unguarded.

## This PR — the guardrail half (doc-only, safe)

- **Sources note corrected:** discovery itself is empty for a closed strategy (not just its telemetry ring); an empty result is a **coverage gap, not zero trades**.
- **Guardrail 9:** distinguish *"never traded"* (genuinely new book → "nothing to review" is correct) from a *"closed-strategy discovery gap"* (`meta.closed_strategy_count > 0` + empty `trades[]` → trades existed). Never tell a user with a history "zero closed trades / can't be answered," and don't hand-read `pnl_history` and narrate its curve.
- **Guardrail 3:** **account value ≠ P&L** — value snapshots aren't trades; a `$0` after a close is a **withdrawal**, not a loss; reconcile realized PnL vs the value drop (the gap is the withdrawal).
- Version **1.2.0 → 1.3.0** (MINOR — manifest auto-upgrades users).

## ⚠️ Required follow-up — the ENGINE half (not in this PR)

The guardrails reduce the harm, but the **complete** fix is in `review.py`: when a **closed** strategy's `discovery_get_trader_history` is empty, fall back to **`strategy_get_pnl_and_account_value_history(wallet)`** and use its all-time realized PnL so `closed_strategies[].realized_pnl` / `pnl_summary.realized_by_book.closed` carry the **real total** (per-trade detail stays unavailable — flag it, don't fabricate trades). Then the agent never needs to touch raw `pnl_history` at all. I left this out because it needs the engine's fixtures/tests — happy to implement it as a stacked PR.

## Validate
On a closed strategy: `discovery_get_trader_history` (empty) vs `strategy_get_pnl_and_account_value_history` (has realized total). Confirms both the gap and the fallback source.

Pairs with the guide/SOUL fixes (senpi-hyperliquid-mcp #173, senpi-agent #52) — same failure class, this is the skill-engine layer.

### DIFF
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -17,7 +17,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.2.0"
+  version: "1.3.0"
   platform: senpi
   exchange: hyperliquid
 ---
@@ -89,6 +89,8 @@ on-disk ring is already gone) the engine **fails open to discovery**: the trades
 recorded on this build"** (or "…this strategy's event ring is gone — it's closed") — **never** report it as a
 bug. `meta.telemetry_source` (`available` / `partial` / `unavailable`) and `meta.exit_reason_source_counts`
 tell you exactly how much enrichment landed; surface that honestly.
+
+**Discovery itself is empty for a *closed* strategy — this is the trap.** `discovery_get_trader_history` returns **no trades** once a strategy is closed/torn down (not just its telemetry ring — the onchain trade **list** too). An empty discovery result for a closed strategy is a **coverage gap, NOT zero trades**: the realized total still exists (via `strategy_get_pnl_and_account_value_history` — all-time realized PnL). Never treat a closed strategy's empty discovery list as "no history." See guardrail 9.
 
 ## Quick actions this skill handles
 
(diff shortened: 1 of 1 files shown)


betashop:
Closing in favor of a proper engine fix (Jason's call). The diagnosis here is correct and the guardrails are good — but the premise that per-trade detail is 'genuinely gone' for a closed strategy is wrong: the fills persist on-chain. I pulled 62 and 138 fills for the two closed Lion sub-wallets directly from Hyperliquid, with full entry/exit/closedPnl/fee. So the fix shouldn't concede the skill's core question ('did I sell too early/late') — it should source closed-strategy trades from the durable on-chain fills, using pnl_history only for the withdrawal-vs-realized reconciliation. Reworking as a fresh PR that carries the corrected guardrails + the engine change.

PR #475, opened 2026-07-15
improve-trades v1.6.0: current-book telemetry scope + stdout slim + Duncan's review + SKILL bloat (folds #479)

betashop:
## The bug (repro: M404726)

Running **"improve my last 10 trades, did I sell too early or late?"** returned *"Telemetry was not available — all 168 trades across every runtime had their event log time out; every `exit_reason` falls back to UNKNOWN."* — while the exit history was sitting in the collector the whole time.

**Root cause — wrong source + no scope, not missing data:**
- The engine enriches exit reasons by shelling `openclaw senpi events --runtime <id>` — the **local, on-disk** event ring.
- A **CLOSED** strategy's ring is torn down with its runtime, so the call **hangs to the timeout**.
- Retrospective review over the default 7-day window fanned that across **dozens of closed runtimes** (M404726 has ~93). N hangs blew the enrichment budget → the whole review reported "telemetry unavailable," even for the live strategies whose ring *was* readable.
- The old `_telemetry_dead` breaker is defeated by the parallel fan-out (every worker is seeded before any completes), and a `TimeoutExpired` never tripped it anyway.

The same exit events (`dsl_breach` / `weak_peak_cut` / `exchange_sl_hit` / …) are **durable in the central event log**, keyed by strategy address. Recovering those for closed strategies is the follow-up (**#2**, a user-scoped collector read).

## This PR (#1 — scope + fast-fail, no platform dependency)

- **Current-book-only telemetry.** Only ACTIVE/PAUSED strategies are ever event-fetched — a closed strategy's ring is gone, so we never shell at it. `exit_reason` falls to the ratchet/UNKNOWN fallback (same outcome as today, minus the hang). Applied to both the `all` and stepped `telemetry` paths.
- **`--last N` is now GLOBAL** — the N most-recent trades across the whole book, not N-per-strategy. "last 10" no longer expands to 10 × strategy_count (that's why the run pulled 142/168 trades).
- **Timeout breaker + shorter timeout** — repeated event-log timeouts trip `_telemetry_dead` after `EVENTS_TIMEOUT_BUDGET`, and the per-call timeout drops 8s → 5s, so a slow host can't cascade even within the current book.
- **SKILL.md** — route "last N" → pass `--last N`; document current-book-only telemetry + the durable-log recovery path for closed strategies; version 1.3.0 → **1.4.0** (MINOR, auto-upgrades users).

## Tests

+3 (`test_closed_strategy_is_never_event_fetched`, `test_last_n_is_global_not_per_strategy`, `test_event_timeout_trips_the_breaker`). **54 pass.**

## Follow-up (not in this PR)

**#2 — durable exit reasons for CLOSED strategies** via a user-scoped read of the central event log (`otel_logs`, keyed by `user.id` + strategy address). That's what recovers the WHY-layer for closed strategies instead of falling back to UNKNOWN; it needs a server-side per-user scoping path (platform).

### DIFF
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -17,7 +17,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.3.0"
+  version: "1.4.0"
   platform: senpi
   exchange: hyperliquid
 ---
@@ -83,12 +83,21 @@ Two sources, two jobs. Never mix them, and never reconstruct one from the other.
   the discovery trades — it fills `exit_reason` and produces the standalone streams; it **never** becomes the
   trade list and **never** re-derives a price or PnL.
 
-**When telemetry is unavailable** (an older runtime build without the event RPC, or a *closed* strategy whose
-on-disk ring is already gone) the engine **fails open to discovery**: the trades are still listed, and
-`exit_reason.terminal` falls back to the ratchet record or honest `UNKNOWN`. Say **"exit mechanism not
-recorded on this build"** (or "…this strategy's event ring is gone — it's closed") — **never** report it as a
-bug. `meta.telemetry_source` (`available` / `partial` / `unavailable`) and `meta.exit_reason_source_counts`
-tell you exactly how much enrichment landed; surface that honestly.
+**Telemetry is read ONLY from the current book.** The event-log ring lives on-disk with a running runtime, so
+the engine shells `openclaw senpi events` **only for ACTIVE/PAUSED strategies** — a *closed* strategy's ring is
+torn down with its runtime, and querying it just hangs to a timeout. (Fanning that hang across a churned book of
+dozens of closed redeployments is what once made a whole review report "telemetry unavailable / every exit
+UNKNOWN," even for the live strategies whose ring was readable. The engine no longer does that.) A closed
+strategy's exit reasons therefore come from the ratchet record or honest `UNKNOWN` — and the **durable central
+event log** (keyed by strategy address, survives the close) is the recovery path for them, not the ephemeral
+local ring.
+
+**When telemetry is unavailable** (an older runtime build without the event RPC, or a closed strategy) the
+engine **fails open to discovery**: the trades are still listed, and `exit_reason.terminal` falls back to the
+ratchet record or honest `UNKNOWN`. Say **"exit mechanism not recorded on this build"** (or "…this strategy is
+closed — its live event ring is gone; the durable log is the recovery path") — **never** report it as a bug.
+`meta.telemetry_source` (`available` / `partial` / `unavailable`) and `meta.exit_reason_source_counts` tell you
+exactly how much enrichment landed; surface that honestly.
 
 **Closed strategies are recovered ON-CHAIN — the engine handles the trap for you.** `discovery_get_trader_history` returns **empty** once a strategy is closed/torn down — Senpi clears its own index, but the trades are **NOT gone** (Hyperliquid keys fills by wallet **address**, so they survive the close). The engine detects an empty discovery result and **falls back to on-chain HL fills**, rebuilding the real round-trips (`meta.closed_trade_source == "onchain_fills"`, wallets in `meta.onchain_recovered_wallets`); `realized_pnl` then comes from HL's own `closedPnl`. So a closed strategy's trades and realized total come back **real** — an empty discovery result is never "no trades." You do **not** hand-read `strategy_get_pnl_and_account_value_history` for this. If `trades[]` is still empty after the on-chain fallback, the book genuinely never traded (guardrail 9).
 
@@ -100,7 +109,7 @@ step(s) the ask needs; route every fix through the depth choice at the end — n
 
 | Intent (what the user asks) | Step(s) to run | Data (engine output) | Actionable lever |
 |---|---|---|---|
-| *"Did I sell too early / late? / improve my last 10"* | `timing` | `timing_summary` (exit_ahead/held_higher/flat) + per-trade `if_held_delta_usd`, `exit_vs_hold` | NEUTRAL context — a reversal is one data point, never "premature"; the counterfactual is not a grade (guardrail 1) → the DSL tier that fired (`exit_reason`) |
+| *"Did I sell too early / late? / improve my last 10"* | `timing`, then `telemetry` for the exit *mechanism* — **pass `--last N`** when the ask names a count ("last 10" → `--last 10`) | `timing_summary` (exit_ahead/held_higher/flat) + per-trade `if_held_delta_usd`, `exit_vs_hold`; `dsl_close_reason_mix` for how each exit fired | NEUTRAL context — a reversal is one data point, never "premature"; the counterfactual is not a grade (guardrail 1) → the DSL tier that fired (`exit_reason`) |
 | *"Master my week" / "analyze my strategies and trades" / "suggest improvements"* | **all steps in order** (`timing`→`strategies`→`telemetry`→`market`) | `timing_summary` + `strategies[]` (per-mandate) + `book_vs_market` + the telemetry streams | Process recap, each strategy vs its own mandate |
 | *"What did I miss this week? / compare to market"* | `market` (+ `telemetry` for the blocked cohort) | `book_vs_market.gaps` (unheld movers) + `missed_signals` (telemetry-blocked) | Is the missed mover in the mandate? loosen a gate only if so |
 | *"How could I make more gains?"* | `strategies` + `telemetry` | `strategies[]` mandate reads + `dsl_close_reason_mix` + `blocked_summary` | Strategy tune (DSL / entry gate), never a $/week promise |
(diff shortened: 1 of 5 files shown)


dmcr (commented):
Direction is right — current-book scoping + global `--last N` match the runtime CLI's actual behavior, and capping before pricing is a nice extra win. Verified the CLI contract (`--runtime/--json/-l/--since`, `{ok, entries}` shape, ACTIVE/PAUSED status vocabulary) against trading_runtime. Two things worth fixing before merge (breaker + narrative, inline), and two responsiveness upgrades left on the table:

1. **Halve per-fetch cost now**: `openclaw senpi events` internally `spawnSync`s a second full `openclaw gateway call` process — every fetch is two Node CLI boots. Shelling `openclaw gateway call senpi.getEvents --json --params '{"runtimeId":…,"limit":500,"since":…}'` directly is one boot for the same data (entries under `payload`).
2. **Structural fix (runtime follow-up)**: the gateway holds all runtime handles in one map, so a batched `senpi.getEvents` (`runtimeIds: []`) would make the whole current book one process spawn instead of N. Worth filing on trading_runtime.

senpi-improve-trades/scripts/review.py
```diff
@@ -484,12 +486,22 @@ def _fetch_events(runtime_id, since_ms, meta):
...
         _note_telemetry_unavailable(meta, "openclaw CLI not found; exit reasons from ratchet fallback only")
         return []
-    except Exception as e:  # noqa — timeout / OS error → fail-open
+    except subprocess.TimeoutExpired:        # a hung event RPC (an unreachable ring). Budget the damage:
+        # a live ring answers in well under a second, so repeated timeouts mean the host's event log is
+        # unreachable — after a few, mark it dead so we stop paying the per-call timeout across the whole
+        # book. This is the safety net behind the CURRENT-strategy-only guard (the pre-fix failure mode was
+        # N runtimes × the timeout → the entire review reported "telemetry unavailable").
+        meta["_telemetry_timeouts"] = meta.get("_telemetry_timeouts", 0) + 1
+        if meta["_telemetry_timeouts"] >= EVENTS_TIMEOUT_BUDGET:
```
dmcr: This breaker can never trip in the engine: both fan-out paths give each worker a **private** meta dict (`priv = {…}` at the two `_worker`s), each worker makes at most one `_fetch_events` call, and `_merge_meta` propagates `_telemetry_dead` but never sums `_telemetry_timeouts`. So no dict ever reaches the budget of 2 — same "every worker seeded before any completes" flaw the PR description diagnoses in the old breaker. Also note `list(ex.map(…))` collects everything before the merge loop, so merge-time accumulation would only help across steps, not mid-fan-out.

Fix: a module-level lock-guarded counter/flag — `_fetch_events` already re-reads dead-state at call time, so a shared flag short-circuits queued strategies beyond the first 8. Plus sum `_telemetry_timeouts` in `_merge_meta` (re-checking the budget there) so the trip carries into the next step.

senpi-improve-trades/scripts/review.py
```diff
@@ -1046,10 +1058,16 @@ def _collect_one_strategy(client, strat, meta, since_ms, until_ms, cap, enrich_e
              "risk_halts": {"count": 0, "samples": []}}
     fills = {"maker": 0, "taker": 0, "unknown": 0}
     closed = fetch_closed_trades(client, strat["wallet"], since_ms, until_ms, cap, meta)
-    # telemetry ring for this runtime (enrichment only; guarded + fail-open to []). Fetched even when
-    # the wallet has no closed trades in-window, since its missed_signals + leaks still count. SKIPPED
-    # entirely on the fast timing path (enrich_exit=False) so telemetry latency never blocks it.
-    events = _fetch_events(strat.get("runtime_id"), since_ms, meta) if enrich_exit else []
+    # telemetry ring for this runtime (enrichment only; guarded + fail-open to []). SKIPPED on the fast
+    # timing path (enrich_exit=False). ALSO skipped for non-current strategies: only ACTIVE/PAUSED runtimes
```
dmcr, line 1060: The mechanism described here isn't what the CLI does: a torn-down runtime id gets an **immediate `NOT_FOUND`** from the gateway (`getRuntimeHandleOrRespond`) — exit 1, no hang. The real per-probe cost is that `openclaw senpi events` `spawnSync`s a second full `openclaw gateway call` process, i.e. two complete Node CLI boots per fetch. Fan 93 of those across the 8-thread pool and CPU starvation pushes *everything* — including live-ring reads — past the timeout, which is why live strategies also reported unavailable.

The fix is still correct (the win is not spawning the process pairs at all), but worth rewording here and in SKILL.md — a future maintainer designing around "closed → hangs" is designing around a failure mode that doesn't exist, and it changes how the timeout should be calibrated.

senpi-improve-trades/scripts/review.py
```diff
@@ -61,6 +61,8 @@
 # ("what did I miss"). It NEVER reconstructs a trade or re-derives a price/PnL — discovery OWNS those.
 EVENTS_FIXTURE_ENV = "SENPI_EVENTS_FIXTURE"   # offline test hook: JSON {"<runtime_id>": [entries…]}
 EVENTS_PULL = 500                             # events to pull per runtime before matching (recent ring)
+EVENTS_CALL_TIMEOUT_S = 5                      # per-runtime event shell-out timeout (a live ring answers <1s)
```
dmcr: "a live ring answers <1s" measures the gateway read, but this timer wraps **double CLI boot + read** (see the nested `gateway call` spawn). On a loaded host a healthy fetch can plausibly exceed 5s — it demonstrably exceeded 8s during the incident. Cutting 8→5s makes spurious UNKNOWNs more likely, and once the breaker actually trips, budget=2 means two slow-but-healthy calls kill telemetry for the whole current book — recreating the reported symptom. Suggest keeping 8s (current-book scope already bounds worst case to ~ceil(N/8) × timeout wall) and/or budget ≥3.

senpi-improve-trades/SKILL.md
```diff
@@ -166,6 +175,12 @@ the Quick-actions "Step(s)" column (e.g. "did I sell too early" → just `timing
 each step returns valid JSON with `meta.warnings` on partial data and never crashes on a missing/corrupt
 state file (it recomputes).
 
+**Scope the ask.** When the user names a trade count — "improve my last 10", "review my last 20 trades" —
+**pass `--last N`**: it's a GLOBAL bound (the N most-recent closed trades across the whole book), so the review
+stays about what they asked and the telemetry step touches only the handful of strategies those trades belong
```
dmcr: Overclaim: both engine paths event-fetch **every** current strategy regardless of the trade cap (intentionally — the same events feed `missed_signals` + leak/fill streams). The honest claim is "closed strategies are never probed and the trade list stays at N." Implementing the claim as written would silently lose missed-signal/leak coverage, so reword rather than implement.

senpi-improve-trades/tests/test_review.py
```diff
@@ -863,6 +863,83 @@ def test_no_trades_when_both_sources_empty():
...
+    4). The two returned are exactly the two newest by close_time."""
+    full = _mixed_result()
+    assert full["meta"]["trade_count"] == 4
+    top2 = [t["asset"] for t in full["trades"][:2]]         # the engine returns trades newest-first
+    capped = _mixed_result(last_n=2)
+    assert capped["meta"]["trade_count"] == 2                # GLOBAL cap — not 2 per wallet
+    assert [t["asset"] for t in capped["trades"]] == top2
+
+
+def test_event_timeout_trips_the_breaker():
```
dmcr: This passes only because it reuses one meta dict across 5 sequential calls — a pattern the engine never produces (per-worker private metas, one call each). It masks the breaker never tripping in production; after fixing the breaker, assert through `_collect_trades`/`_enrich_exit_and_streams` with a timing-out subprocess stub instead of calling `_fetch_events` directly.

shnoodles (approved):

betashop:
Thanks @dmcr — all five addressed in `821cedb1`, and you were right on every one (the breaker one especially: I'd diagnosed that exact "seeded before any completes" flaw in the *old* breaker and then reproduced it in the new one).

- **Breaker can't trip** → **removed** `EVENTS_TIMEOUT_BUDGET`/`_telemetry_timeouts` entirely. The current-book-only guard is the real, sufficient bound; a `TimeoutExpired` now just fails open for that one strategy. No lock-guarded counter added — simpler to not have a breaker than a correct one the scoping makes unnecessary.
- **Masking test** → replaced with `test_event_timeout_fails_open_without_a_breaker` (asserts fail-open + not-terminal + `not hasattr(review, "EVENTS_TIMEOUT_BUDGET")` so it can't come back).
- **"closed → hangs" is wrong** → reworded both code comments **and** both SKLL passages to the real mechanism: immediate `NOT_FOUND`, and the cost is `openclaw senpi events` spawnSync-ing a second CLI process → process-pair fan-out starves CPU → even live reads time out.
- **5s too tight** → reverted to **8s**, with the comment noting the timer wraps double-boot+read.
- **`--last` overclaim** → reworded: both paths still fetch every *current* strategy (those events feed `missed_signals`/leaks); the win is closed strategies are never probed.

**Perf upgrades you left on the table** — agree both are worth it, filing as follow-ups rather than in this PR since the fixture tests bypass the subprocess and I can't offline-verify the new wire shape:
1. Shell `openclaw gateway call senpi.getEvents` directly (one boot, entries under `payload`) — halves per-fetch cost.
2. Batched `senpi.getEvents(runtimeIds:[])` → one spawn for the whole current book — runtime-team change.

This PR also now folds in the SKILL bloat cleanup (was #479, which stacked on this branch): output-shape catalog → `references/output-shape.md`, and a small strategy-author tone-trim. 92 tests green. Ready for re-review.

shnoodles (approved):

PR #480, opened 2026-07-17
Temporary template-only deploy gate (while we harden the custom builder)

betashop:
## What
Two layers, one flag, fully reversible.

1. **`strategy-author` SKILL — the steer (primary gate).** Any "build from scratch / design my own" ask is routed to the closest proven template via `senpi-strategy-discover`, with warm, power-user framing (*every strategy is real quant engineering — scanner + scoring + runtime + DSL exit — that we're upgrading right now; start from proven code*). The original 3-path interview is **kept intact below** a single comment block.

2. **`deploy.py` — the backstop (enforcement).** `create` refuses to **fund** a package whose id isn't a catalog template, *before any wallet is created*, with the same warm pointer. Membership is checked **by catalog id** (survives fetch-caching, so a re-deployed template is never falsely blocked). **Fail-open** if the catalog is unreachable — safe, because funding needs the network anyway, so a fail-open can't complete an offline scratch deploy.

## Reversibility (by design)
- Flip **`TEMPLATE_ONLY_MODE = False`** in `deploy.py` — one line, custom deploys return instantly.
- Delete the one comment block in the author SKILL — the 3-path "Start here" governs again.
- **`SENPI_ALLOW_CUSTOM_DEPLOY=1`** overrides per-invocation (internal fleet / CI deploying a new, not-yet-catalogued strategy).

## Scope decision
**Strict template-only** (blocks scratch *and* fork-to-new-id): matches "only allow launching from templates," simplest to enforce, and a stopgap — editing a template's config (universe/thresholds/sizing/DSL) stays fully open and is the reliable way to customize today.

## Note
#476 (merged) already targets the deploy-contract failures behind most of that >90%. Recommend **re-running the bake-off on current main** to re-measure before deciding how long this stays on — this is the safety net while confidence rebuilds, likely short-lived.

## Tests
+7 (block scratch / allow template / override / non-create / fail-open / flag-off / catalog parse+cache). Full ops suite: **41 passed**. Live-verified against the real catalog (91 templates, raven allowed, scratch blocked with pointer, validate not gated).

MINOR bumps: strategy-ops 2.5.0→2.6.0, strategy-author 2.6.1→2.7.0.

### DIFF
--- a/senpi-strategy-ops/scripts/deploy.py
+++ b/senpi-strategy-ops/scripts/deploy.py
@@ -24,6 +24,7 @@ on first use if it isn't on disk.
 # Copyright 2026 Senpi (https://senpi.ai) — Apache-2.0
 import argparse
 import json
+import os
 import re
 import sys
 import time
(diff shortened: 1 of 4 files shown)


senpi-strategy-ops/scripts/deploy.py
```diff
@@ -44,6 +45,71 @@
...
+    global _CATALOG_IDS
+    if _CATALOG_IDS is not None:
+        return _CATALOG_IDS
+    try:
+        status, body = _fetch._get("raw.githubusercontent.com",
+                                   f"/{_fetch.REPO}/{_fetch.REF}/strategies/catalog.json", "*/*", 15)
+        if status != 200:
+            return None
+        skills = json.loads(body).get("skills", [])
+        _CATALOG_IDS = {s["id"] for s in skills if isinstance(s, dict) and s.get("id")}
```
dmcr, line 94: **Recommended (unless intentional):** a 200 response with a missing/renamed/empty `skills` key yields an empty set, which gets cached and then blocks *every* `create` — including real templates. That's fail-closed on catalog schema drift, while fetch failures deliberately fail open. Since `catalog.json` is generated and version-stamped, drift is plausible during exactly the window this gate is on. One-line fix — treat "parsed but found nothing" as can't-verify:

```python
_CATALOG_IDS = {s["id"] for s in skills if isinstance(s, dict) and s.get("id")} or None
```

Worth a test next to the existing fail-open ones.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -44,6 +45,71 @@
...
+    """The allow-list — ids in strategies/catalog.json on the deploy ref, fetched from the public repo
+    (same channel as package fetch) and cached. Returns None if the catalog can't be read; the guard
+    treats None as 'can't verify → don't block' (fail-open). That's safe: funding a wallet needs the
+    network anyway, so a fail-open here can't let an offline scratch deploy actually complete."""
+    global _CATALOG_IDS
+    if _CATALOG_IDS is not None:
+        return _CATALOG_IDS
+    try:
+        status, body = _fetch._get("raw.githubusercontent.com",
+                                   f"/{_fetch.REPO}/{_fetch.REF}/strategies/catalog.json", "*/*", 15)
```
dmcr, line 90: Note (minor, internal knowledge): the catalog is always fetched from `_fetch.REF` (env default), while the package fetch honors `--ref` — so a template that only exists on a feature ref, deployed with `--ref feat/x`, would be blocked. Internal-only scenario and `SENPI_ALLOW_CUSTOM_DEPLOY=1` covers it; the docstring's "same channel as package fetch" is just slightly stronger than what the code does. No action needed.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -44,6 +45,71 @@
...
+
+def _template_only_guard(pkg, cmd):
+    """Backstop for TEMPLATE_ONLY_MODE: refuse to FUND a package whose id isn't a catalog template.
+    Gates `create` only (the funding step); with no wallet, runtime/verify can't proceed anyway."""
+    if not TEMPLATE_ONLY_MODE or cmd != "create":
+        return
+    if os.environ.get("SENPI_ALLOW_CUSTOM_DEPLOY"):
+        return
+    ids = _catalog_template_ids()
+    if ids is None or pkg.id in ids:  # can't verify → don't block;  in catalog → allow
```
dmcr, line 108: Note: since membership is by id and a local package dir is authoritative, a scratch package named after a catalog template (e.g. `id: raven`) passes the backstop. Unavoidable given "template + edits" stays deployable under the same id — just flagging that the hard gate is soft against an agent that names its scaffold after a template; the SKILL steer is the real defense there. No action needed.

dmcr (commented):
Looks good overall — verified guard placement (pre-funding), id-based membership, and reversibility. One recommended change below (if the current behavior isn't intentional); the other two comments are just notes, no action needed.

dmcr (approved):

betashop:
Closing — decided not to take this approach. We're not gating custom builds behind a template-only path; hardening the custom builder directly instead.

PR #481, opened 2026-07-21
Fix evm_usdc always empty: parse balanceInUSD in token_balances

YosephKS:
## Summary

Fixes the reported bug where `evm_usdc` in the `senpi-portfolio` embedded-wallet report was always `[]` / `$0` even when the embedded wallet held USDC on EVM chains, under-reporting `grand_total` and making `reconciles` false.

**Root cause:** `fetch_embedded()` parsed `account_get_portfolio.data.portfolio.token_balances` looking for the USD amount under `usdValue` / `usd_value` / `amountUsd` / `balanceUsd` / `amount` — but the live GetPortfolioV3 response carries it as `balanceInUSD` (with `formattedBalance` / `tokenPriceInUSD`). Every token fell through to the `0.0` default, so the `evm_usdc` list was never populated.

**Fix** (in `senpi-portfolio/scripts/portfolio.py`):
- Added `balanceInUSD` to the `_f()` field-name fallback chain.
- When the amount still resolves to `0.0`, fall back to `formattedBalance` × `tokenPriceInUSD` (price defaults to `1.0`).

## Tests

Two regression tests added in `senpi-portfolio/tests/test_portfolio.py`, mirroring the existing nested-`total_in_hyperliquid` regression test:
- `test_embedded_evm_usdc_reads_balanceInUSD` — uses the exact live response shape from the bug report (USDC on Base, `balanceInUSD: 7.9839…`) and asserts `evm_usdc == [{"chain": "EVM", "usd": 7.98}]` with `idle_in_embedded == 7.98`.
- `test_embedded_evm_usdc_falls_back_to_formattedBalance_times_price` — covers the `formattedBalance × tokenPriceInUSD` fallback when `balanceInUSD` is absent.

Full suite: **37/37 passing.**

## Note

The live shape only carries `chainId` (e.g. `8453`), and the chain-label lookup only knows `chain`/`network`/`chainName`, so live-API entries are labeled `"EVM"` rather than `"Base"`. Dollar amounts are correct; a chainId → name mapping could be a follow-up if chain-specific labels are wanted.

https://claude.ai/code/session_01JZmRaUdkg67mraUcCr78d3

---
_Generated by [Claude Code](https://claude.ai/code/session_01JZmRaUdkg67mraUcCr78d3)_


shnoodles (approved):

0xsarvesh:
### Verified against a live payload — LGTM with one nit inside the added lines

Ran this branch's `portfolio.py` against a verbatim `account_get_portfolio` response captured from production telemetry — the exact bytes that produced a wrong "how much can I deploy?" answer in a real session. The fix does what it says:

| | before | after |
|---|---|---|
| `evm_usdc` | `[]` | `[{"chain": "EVM", "usd": 93.95}]` |
| `idle_in_embedded` | `135.21` | `229.16` |
| `totals.reconciles` | `False` (gap $95.74, tol $2.00) | `True` (gap $1.79, tol $2.86) |
| suite | — | 37/37 pass |

`formattedBalance` also coerces correctly when it arrives as a string (`_num` → `float()`).

### Nit — blocking, because it's in the new code: `tokenPriceInUSD: 0` re-opens the bug

`portfolio.py:530`

```python
price = _f(tb, "tokenPriceInUSD", default=1.0)
```

`_f` returns the first key **present** with a numeric value, so a present-and-zero price yields `0.0`, and `raw * 0 == 0` makes the balance invisible again — the exact failure this PR fixes. Probed on this branch:

| `token_balances` entry | result |
|---|---|
| `balanceInUSD` present | ✅ `[{'chain': 'EVM', 'usd': 93.95}]` |
| `balanceInUSD` absent, price present | ✅ `[{'chain': 'EVM', 'usd': 93.95}]` |
| `balanceInUSD` absent, `tokenPriceInUSD: 0` | ❌ `[]` |
| `balanceInUSD: 0`, no price key | ✅ `[{'chain': 'EVM', 'usd': 94.0}]` |

Zero-as-missing is a known shape on this API — #537's description documents having to coerce `0` sentinels to `None` in the discovery payloads. Suggested:

```python
price = _f(tb, "tokenPriceInUSD", default=1.0) or 1.0
```

### Not blocking — pre-existing, will land in a separate PR

**1. `idle_total` still omits the HL spot USDC bucket.** `spot_usd` is read at `portfolio.py:520` and never added at `:536`. `deploy.py`'s funding waterfall is perps + spot + EVM; on the payload above that is `135.21 + 1.81 + 93.95 = 230.96`, while this branch reports `229.16`. Note the residual hides *under* the reconciliation tolerance (gap $1.79 < the $2.00 floor), so the flag now reads `True` — this PR turns a loud $95.74 mismatch into a quiet $1.81 one.

**2. The canonical fixture encodes a payload shape the API does not return.** `tests/fixtures/portfolio_fixture.json` uses `{"symbol", "chain", "usdValue"}`; live is `{"tokenSymbol", "chainId", "balanceInUSD", "formattedBalance", "tokenPriceInUSD"}`. The two tests added here use the real shape 👍, but the shared fixture still doesn't — so `test_embedded_idle_is_not_total_withdrawable` and the `step_money` tests validate against fiction. This is why the bug shipped green: the suite proved the parser reads the fixture, not the API.

**3. `totals.reconciles == false` emits no `meta.warnings` entry** — every other failure mode in this file does (13 append sites). A reader scanning `meta.warnings` sees `[]` next to a silent mismatch, so `SKILL.md`'s "STOP and re-run on `reconciles: false`" rule never fires.

### Context worth having

`senpi-strategy-ops/scripts/deploy.py:200-247` already had the correct reader — `_usdc_usd(dig(port, "token_balances"), "balanceInUSD", "formattedBalance")` — and its docstring names the earlier incident: a $440 deploy refused as `[E_FUNDS_SHORT] only $198.42 is accessible` with ~$248 USDC idle on Base. Same field, same class, already fixed on the ops side; `portfolio.py` was never brought along.

The two parsers can drift again, and a shared module isn't the answer (skills install standalone — `mcp_client.py` / `_yaml.py` are duplicated per skill by design). The durable guard is a fixture **recorded from a real response** in each suite, rather than one written by hand.

Ship it with the `or 1.0`.

PR #483, opened 2026-07-21
verify: read state (stderr + scannerId + runCount-0) — stop false 'scanner=broken'

betashop:
Fixes the **false `scanner=broken` on a healthy deployment** that makes agents thrash. **Confirmed twice in prod** — M404726 (agent close/create loop) and **M408027's LION deploy** (GLM spent ~13 min reverse-engineering `verify` before confirming a deploy that had actually succeeded). Three compounding bugs, all in the `verify` read path:

## 1. `cli_json` was blind to state on stderr
`verify` reads state via `openclaw senpi state -r <id> --json`, but that command prints its JSON to **stderr** (M408027's agent measured it: `out_len=823` log-noise vs `err_len=28735` JSON). `cli_json` parsed stdout only → `None` → false broken. **Fix:** parse stdout, fall back to stderr.

## 2. Scanner is keyed `scannerId`, not `name`
The real state nests scanners at `states[].components.scanners.state.scanners[]`, each keyed **`scannerId`** (+ `health`) — `_deep_find_scanner` searched `name`/`scanner`/`scannerName`, so **even with the JSON in hand it never found the scanner** → still false broken. **This is why fix #1 alone was insufficient.** **Fix:** also match `scannerId`/`id`; honor the `health` field (real objects carry `health`, not `enabled`/`lastError`).

## 3. `runCount 0` on a slow scanner is healthy, not broken
lion-short showed `runCount 0, health healthy` (slow first tick). `verify` already scores that `scheduled` = live; hardened the message so an agent can't misread it and thrash, plus a SKILL anti-thrash rule (never close/recreate to force a tick — cites the LION incident).

## Tests — 45 pass (incl. +3 on the **real prod state schema**: ticked / scheduled@runCount0 / unhealthy)
MINOR: strategy-ops → **2.7.0** (2.6.0 taken by #482).

> Root cause for the runtime/openclaw side (@sarvesh): `senpi state --json` should print the JSON payload to **stdout**, not stderr — then no consumer has to guess the stream.

### DIFF
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -18,7 +18,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.6.0"
+  version: "2.7.0"
   platform: senpi
   exchange: hyperliquid
   requires:
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -127,6 +127,17 @@ flagged component and re-run. It's a **single fast check** — a scheduled scann
 wait for the first scan tick and you must **never `sleep` then verify**. `deploy.py status <id>` shows deploy
 state any time.
 
+> **⛔ `runCount 0` / no positions ≠ broken — do NOT thrash the strategy.** A newly-deployed scanner shows
+> **`runCount = 0`, no signals, no positions until its first interval elapses** — and on a **slow clock that is
+> HOURS** (the 6h regime scanners; a from-scratch 4–6h build). That is **healthy** — `verify` reports it as
+> `scanner=scheduled` = **live**. It does **not** mean the deploy failed. **Never `close`/recreate, re-run
+> `create`, or re-`runtime` to "force a tick"** — that tears down a working strategy on a fresh wallet and, if it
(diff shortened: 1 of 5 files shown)


betashop:
Superseded by #486, which fixes this at the right level.

#483 patched the SURFACE of the false-`scanner=broken` verify (JSON-on-stderr + `scannerId` key matching). #486 went a level deeper: the real problem isn't that the read is parsed wrong, it's that `senpi state`/`status` are intermittently unavailable for a minute+ after a runtime starts, so gating liveness on them is fragile at all. #486 moves the gate onto the reliable backbone (`runtime list` + runtime.yaml + `strategy_list`), demotes `status`/`state` to downgrade-only enrichment, and fails OPEN — which is the correct direction for a liveness check, since a false `broken` is what made agents thrash and destroy working strategies. It also caught a money-critical mirror bug on the close path that #483 missed entirely.

Branch is also conflicting now. Closing rather than rebasing.

One piece of #483 is NOT subsumed and is being tracked separately: `cli_json` reads stdout only, but `senpi state --json` puts its payload on STDERR (measured on M408027's lion deploy: 823B of log noise on stdout vs 28,735B of JSON on stderr). Without the stderr fallback, #486's downgrade path can never actually read state, so it always fails open and can't catch a genuinely broken scanner. Fail-open is the right default, but it should be a fallback, not the only reachable branch.

PR #486, opened 2026-07-22
strategy-ops v2.8.0: verify gates on the reliable `runtime list` backbone — flaky status/state only downgrade

0xsarvesh:
## Problem

`deploy.py verify` reported **fully-healthy** strategies as not-live and sent the agent into long debug loops — then it would override its own gate and tell the user "live" anyway.

**Root cause:** scanner liveness was judged from `openclaw senpi state --json` (`getSystemState`) — and that JSON gateway read is **intermittently flaky-empty / throws for a minute+ right after a runtime starts**. verify failed *closed* on an unreadable read (`scanner=broken → not-live`), with no fallback.

Confirmed live across three sessions (telemetry):
- **lion** (M408027): scanner logged 519 lines, POSTed signals, opened positions while `verify` called it `broken` 4×; `senpi status` said `2/2 enabled and healthy` the whole time.
- **thesis-fund** (M8): `verify` got `scanner=unknown` (empty reads) while a manual `status -r`/`state -r` **seconds apart** returned fully healthy.

## The insight: `runtime list` is reliable, `status`/`state` are not

| Command | Answers | Reliability post-deploy |
|---|---|---|
| `openclaw senpi runtime list` | runtime **running?** (id/wallet/source/status) | **reliable** (authoritative text inventory) |
| `senpi status` / `senpi state` `--json` | per-**component** health (scanner/DSL) | **flaky-empty/throws** for a minute+ after start |

A running runtime **spawns and supervises** its declared external scanner (restarting it on crash), so **runtime-running + scanner-declared ⇒ the scanner is being driven**.

## Fix — gate on the reliable backbone; flaky reads only downgrade

The gate now rests only on signals that are reliably readable right after deploy:
- **`runtime list`** — is the runtime running?
- the deployed **runtime.yaml** — external scanner declared + DSL preset present
- MCP **`strategy_list`** — funded to what was requested

`status`/`state` are demoted to **enrichment that can only DOWNGRADE** a scanner to `broken` on *positive* evidence (disabled / erroring / runtime-reported unhealthy). When they're unreadable, the scanner reads **`supervised` = live** (the DSL protects positions regardless). Liveness is never read from on-disk state files. External scanners are judged by `health` / `lastAliveAt`, **not** `runCount` (which stays 0 for a barren-but-healthy scanner; the scaffold also runs tick 1 on launch, so a long-interval — even 4h — strategy passes immediately as `scheduled`/`supervised`, never blocking on a scan tick).

Result: `verify` is a clean **`live | not-live`** again (no `unverified` limbo), returns in **one fast pass** (dropped the redundant per-call retries that briefly blew the tool timeout in live testing; skips the 60s-timeout `runtime list` call when the health map already confirms running).

## Close side — same lesson (trust `runtime list`, not a gateway command's exit code)

Teardown had the identical bug on the write path. `close_one` failed the whole instance — **aborting before the money-critical `strategy_close`** — whenever `openclaw senpi runtime delete` returned non-zero. But that exit code is not a reliable failure signal: `senpi.deleteRuntime` responds `NOT_FOUND` (→ non-zero) when the runtime is **already gone** (so idempotent poll re-runs false-failed), and the delete rides the same flaky gateway with `stdio: inherit` (banner leaked into the captured "error"; transient hiccup → spurious non-zero). Seen live (M8): `close.py --all` reported `thesis-fund: failed` with the OTel banner as the error and left the strategy + runtime for manual cleanup.

Fix: after `runtime delete`, judge success by whether the runtime is **gone from `runtime list`** — not `rc`. Gone ⇒ success (whether we deleted it or it was already gone) ⇒ proceed to `strategy_close`. Still present after one retry ⇒ genuine failure with a clean, actionable message (no banner spam), and we do **not** `strategy_close` (keeps the safe stop-before-close ordering so a live runtime can't re-enter positions).

## Changes
- `_cli.py` — `runtime_state`/`runtime_status` single fast reads; `scanner_health_in_status` reads per-scanner health from `senpi status`.
- `deploy.py` — `_scanner_verdict(inst, state, status)` with the `supervised` terminal verdict; `_check_live` skips the redundant `runtime list` call in the common path; `cmd_verify` back to `live | not-live`.
- `close.py` — `close_one` confirms runtime deletion via `runtime list`, not the delete's exit code (idempotent + flaky-tolerant; never false-aborts `strategy_close`).
- Docs — `references/liveness-verification.md` reframed (runtime-list backbone; status/state downgrade-only; `runCount>0` not required; on-disk files aren't a contract) + SKILL.md verify section.
- **Version** — SKILL.md `2.6.0 → 2.8.0`.

## Tests
`tests/test_deploy_gates.py` — covers state-throws-but-status-healthy, heartbeat-without-runCount, unhealthy downgrade, and the fail-open `supervised` path. 29 gate tests + 13 pkg tests pass.

## Validated live
M8 installed the branch and deployed `ant` / `thesis-fund` end-to-end; the fixes were iterated against those runs.

### DIFF
--- a/senpi-strategy-ops/scripts/close.py
+++ b/senpi-strategy-ops/scripts/close.py
@@ -36,6 +36,17 @@ LIVE_STATUSES = ["ACTIVE", "PAUSED", "CREATE_WALLET", "FUND_WALLET", "INITIALIZE
                  "SUBSCRIBE_TRADER", "CLOSING_POSITIONS"]
 
 
+def _runtime_gone(name):
+    """True ONLY when `runtime list` was read successfully AND `name` is absent from it. An UNREADABLE
+    inventory (rc!=0 / garbled → None) returns False: on teardown's money path we must never mistake
+    'couldn't read the inventory' for 'the runtime is gone' and then strategy_close a strategy whose
+    runtime is still live and could re-enter positions. Fail CLOSED here — the caller retries / reports."""
+    rts = _cli.list_runtimes_or_none()
+    if rts is None:
+        return False
+    return not any(_cli.runtime_name(r) == name for r in rts)
+
+
 def close_one(label, strat, runtimes, dry_run, log):
     """Stop the runtime (FIRE — no confirm-wait) + TRIGGER strategy_close, then return immediately. The
     agent polls by re-running close.py (idempotent: runtime already gone → skip; status closing/closed →
@@ -74,13 +85,15 @@ def close_one(label, strat, runtimes, dry_run, log):
         log(f"  [{label}] stopping runtime {rname!r}…")
         _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
                       "--address", wallet or ""], timeout=60)
-        if _cli.find_runtime(rname) is not None:   # still listed → one retry, then believe it's stuck
+        if not _runtime_gone(rname):   # still listed OR inventory unreadable → one retry, then treat as stuck
             _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
                           "--address", wallet or ""], timeout=60)
-        if _cli.find_runtime(rname) is not None:
+        if not _runtime_gone(rname):
             rec["status"] = "failed"
-            rec["error"] = (f"runtime {rname!r} still in `runtime list` after delete — it may re-enter "
-                            f"positions; delete it (`openclaw senpi runtime delete {rname}`) then re-run close")
+            rec["error"] = (f"runtime {rname!r} still in (or unreadable from) `runtime list` after delete — "
+                            f"it may re-enter positions; delete it "
+                            f"(`openclaw senpi runtime delete --id {rname} --address {wallet or '<wallet>'}`) "
+                            f"then re-run close")
             return rec
         rec["runtime"] = "stopped"
     else:
--- a/senpi-strategy-ops/scripts/close.py
+++ b/senpi-strategy-ops/scripts/close.py
@@ -64,15 +64,23 @@ def close_one(label, strat, runtimes, dry_run, log):
         rec["status"] = "closed"
         return rec
 
-    # 1. stop the runtime if one is live — FIRE only, no confirm poll (the re-run poll confirms via
-    #    strategy_list status; blocking here cost ~30s/instance). Idempotent across re-runs.
+    # 1. stop the runtime if one is live. Trust `runtime list` (the authoritative inventory), NOT the
+    #    delete's exit code: `runtime delete` rides the flaky gateway (its OTel/HyperDX banner leaks into
+    #    our captured output) AND returns NOT_FOUND / non-zero when the runtime is ALREADY gone — so a
+    #    non-zero is NOT proof of failure. Trusting it both broke idempotent re-runs (a poll re-run hit
+    #    NOT_FOUND → false 'failed') and false-aborted the money-critical strategy_close below, stranding
+    #    the strategy open (seen live). The delete succeeded iff the runtime is gone from `runtime list`.
     if rt:
         log(f"  [{label}] stopping runtime {rname!r}…")
-        rc, _o, err = _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
-                                    "--address", wallet or ""], timeout=60)
-        if rc != 0:
+        _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
+                      "--address", wallet or ""], timeout=60)
+        if _cli.find_runtime(rname) is not None:   # still listed → one retry, then believe it's stuck
+            _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
+                          "--address", wallet or ""], timeout=60)
+        if _cli.find_runtime(rname) is not None:
             rec["status"] = "failed"
-            rec["error"] = (err or "runtime delete failed").strip()[:300]
+            rec["error"] = (f"runtime {rname!r} still in `runtime list` after delete — it may re-enter "
+                            f"positions; delete it (`openclaw senpi runtime delete {rname}`) then re-run close")
             return rec
         rec["runtime"] = "stopped"
     else:
--- a/senpi-strategy-ops/scripts/deploy.py
+++ b/senpi-strategy-ops/scripts/deploy.py
@@ -517,7 +517,7 @@ def _scanner_verdict(inst, state, status):
             return "ticked", "external scanner heartbeat live"
         return "scheduled", f"awaiting first tick (~{inst.interval_seconds or '?'}s cadence)"
     # `state` unreadable → trust the runtime's own scanner-health from `senpi status`
-    sh, _list_seen = _cli.scanner_health_in_status(status, name)
+    sh = _cli.scanner_health_in_status(status, name)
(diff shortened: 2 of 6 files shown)


senpi-strategy-ops/scripts/close.py
```diff
@@ -64,15 +64,23 @@ def close_one(label, strat, runtimes, dry_run, log):
...
         log(f"  [{label}] stopping runtime {rname!r}…")
-        rc, _o, err = _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
-                                    "--address", wallet or ""], timeout=60)
-        if rc != 0:
+        _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
+                      "--address", wallet or ""], timeout=60)
+        if _cli.find_runtime(rname) is not None:   # still listed → one retry, then believe it's stuck
+            _cli.run_cli(["openclaw", "senpi", "runtime", "delete", "--id", rname,
+                          "--address", wallet or ""], timeout=60)
+        if _cli.find_runtime(rname) is not None:
```
danielmbirochi: **Fail-open on the money path if `runtime list` itself fails.** `_cli.find_runtime(rname)` returns `None` both when the runtime is genuinely gone **and** when the `runtime list` read errors/times out (`list_runtimes()` returns `[]` on `rc != 0`, and also if the header line never parses). In that failure mode a still-running runtime reads as "gone", and we proceed to `strategy_close` — exactly the "live runtime re-enters positions on a closed strategy" hazard the comment above warns about. The old code failed closed here.

If `runtime list` is as reliable as the live validation suggests, this is probably acceptable — but it silently conflates *absent* with *unreadable* on the highest-stakes path in the repo. Cheap hardening: a `list_runtimes` variant that returns `None` on `rc != 0`, so `close_one` can treat *unreadable* as "still present → retry/fail" and keep fail-open only for a successful-but-empty read.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -475,26 +475,61 @@ def _deep_find_scanner(obj, name):
...
+        if str(_cli.dig(sc, "health") or "").lower() == "unhealthy":
+            return "broken", "scanner reported unhealthy by the runtime"
+        runs = _cli.dig(sc, "runCount", "ticks", "runs", default=0) or 0
+        if isinstance(runs, (int, float)) and runs > 0:
+            return "ticked", f"{int(runs)} scan(s)"
+        if _cli.dig(sc, "lastAliveAt"):
+            return "ticked", "external scanner heartbeat live"
+        return "scheduled", f"awaiting first tick (~{inst.interval_seconds or '?'}s cadence)"
+    # `state` unreadable → trust the runtime's own scanner-health from `senpi status`
+    sh, _list_seen = _cli.scanner_health_in_status(status, name)
```
danielmbirochi: **`list_seen` is returned but never used.** `scanner_health_in_status`'s docstring says the flag exists so "absent from a populated list" can be told apart from "status unreadable" — but this caller discards it and treats both as `supervised`. Absence from a *populated* scanner list is arguably positive evidence of a mount/name mismatch (the classic silent no-trade). Either use the flag to downgrade that case, or drop the second return value — as written it's dead code implying a check that doesn't happen (and `test_absent_from_state_and_status_is_supervised_not_broken` codifies the ignore as intended).

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -475,26 +475,61 @@ def _deep_find_scanner(obj, name):
...
+        err = _cli.dig(sc, "lastError")
+        cec = _cli.dig(sc, "consecutiveErrorCount", default=0) or 0
+        if err or (isinstance(cec, (int, float)) and cec >= 1):
+            return "broken", f"scanner erroring: {str(err)[:120] if err else f'{int(cec)} consecutive errors'}"
+        if str(_cli.dig(sc, "health") or "").lower() == "unhealthy":
+            return "broken", "scanner reported unhealthy by the runtime"
+        runs = _cli.dig(sc, "runCount", "ticks", "runs", default=0) or 0
+        if isinstance(runs, (int, float)) and runs > 0:
+            return "ticked", f"{int(runs)} scan(s)"
+        if _cli.dig(sc, "lastAliveAt"):
```
danielmbirochi, line 516: **Doc/code mismatch on `lastAliveAt` freshness.** `references/liveness-verification.md` says a heartbeat counts only when `now − lastAliveAt ≤ 2 × interval_seconds`, but this accepts *any* truthy `lastAliveAt` as `ticked`. Fine for a right-after-deploy gate (staleness is unlikely then), but the doc presents the freshness rule as the verdict criterion for hand triage too — worth either checking freshness here or noting in the doc that `deploy.py verify` intentionally skips it.

senpi-strategy-ops/scripts/close.py
```diff
@@ -64,15 +64,23 @@ def close_one(label, strat, runtimes, dry_run, log):
...
             rec["status"] = "failed"
-            rec["error"] = (err or "runtime delete failed").strip()[:300]
+            rec["error"] = (f"runtime {rname!r} still in `runtime list` after delete — it may re-enter "
+                            f"positions; delete it (`openclaw senpi runtime delete {rname}`) then re-run close")
```
danielmbirochi: **The suggested command won't run as written.** The message says `openclaw senpi runtime delete {rname}`, but the code itself always invokes it as `runtime delete --id <name> --address <wallet>`. Someone pasting the suggested command mid-incident hits a usage error. Suggest: `openclaw senpi runtime delete --id {rname} --address {wallet}`.

danielmbirochi (commented):
Reviewed — design is sound and all 45 tests pass on the branch. Four inline notes: one fail-open risk on the close money path to consciously accept (or harden), plus three minor nits.

danielmbirochi (approved):

0xsarvesh:
Thanks @danielmbirochi — all four addressed in 30d6636.

1. **close.py:80 — fail-open on unreadable `runtime list` (the money-path one).** Good catch, this was the real risk. Added `_cli.list_runtimes_or_none()` which returns `None` when the read fails (`rc != 0`) *or* the output doesn't parse as the table (`_parse_runtime_list` now reports validity, so a garbled/banner-only stdout is `None`, not `[]`). `close_one` uses a new `_runtime_gone()` that is True **only** on a successful-and-absent read — an unreadable inventory fails **closed** (retry → `failed`, never `strategy_close`). New test `test_unreadable_inventory_fails_closed`.

2. **deploy.py:520 — dead `list_seen`.** Dropped it. The caller genuinely treats "absent from a populated list" and "unreadable" the same (a running runtime supervises the scanner either way — and right after deploy an absent row is usually still-mounting, not a permanent mismatch), so returning the flag implied a check that wasn't happening. `scanner_health_in_status` now returns just the health string.

3. **deploy.py:516 — `lastAliveAt` freshness doc/code mismatch.** Documented the intent rather than adding a clock to the gate: `deploy.py verify` intentionally accepts any `lastAliveAt` (deploy-time gate; staleness can't have accrued, and the runtime's own `health` catches a stalled scanner). `references/liveness-verification.md` now says hand triage checks freshness, verify doesn't.

4. **close.py:83 — bad suggested command.** Fixed to the exact invocation `openclaw senpi runtime delete --id <name> --address <wallet>`.

33 gate + 13 pkg tests pass.

dmcr (approved):

PR #488, opened 2026-07-23
strategies: make the fleet dual-DEX (main/xyz) aware — 3 silent bugs

betashop:
Found while debugging **bobcat** and **gibbon** live. Hyperliquid is two sub-DEXes behind **one** cross-margined wallet — ~230 bare-named main perps plus **103 `xyz:`-prefixed** equity/commodity/index perps (verified against prod MCP today). Three idioms across the fleet were blind to that.

**All three fail silently.** No exception, no error log — the strategy just never trades, or re-opens what it already holds. That's why they survived this long.

Each reproduces against `origin/main`:

| # | Bug | Pre-fix behaviour | Blast radius |
|---|---|---|---|
| 1 | `leaderboard_get_markets` returns a **bare** ticker + a separate `dex` field, but the universe carries `xyz:NVDA` — so `if token != coin` never matched | `bobcat._get_sm_direction("xyz:NVDA")` → `(None, 0.0)` = "no smart-money data". Where SM is a **hard gate**, the entire xyz universe was blocked permanently | **26 scanners** |
| 2 | `strategy_get_clearinghouse_state` returns `{"main":…, "xyz":…}` with `assetPositions` **inside each section**; scanners read them off the top level | `gibbon._held()` → `set()` while holding BTC **and** `xyz:GOLD` → scanner re-opened held names (two failed duplicate `xyz:GOLD` opens) | **11 scanners** |
| 3 | `market_list_instruments(dex="")` returns **both** sub-DEXes, and `_pick`'s dedupe set was per-call | `[('BTC',''), ('xyz:NVDA',''), ('xyz:GOLD',''), ('xyz:NVDA','xyz'), ('xyz:GOLD','xyz')]` — every xyz name duplicated, first copy misclassified as crypto | regime family (chimp/gorilla/gibbon/orangutan, byte-identical `scan.py`) |

### Fixes
1. `_sm_row_matches()` — compares **bare** tickers **and** requires the `dex` to agree, so main `GOLD` can't cross-match `xyz:GOLD`. Also fixes bison's compare, which was additionally case-sensitive.
2. Enumerate **both** sections, with a legacy-flat fallback. `bison` already did this correctly (and documents why `accountValue` must use `max()`, never `sum` — one wallet, two views) — it's now named as the reference implementation.
3. Shared `seen` set + prefix-based routing. Confirmed **zero** bare-ticker collisions between the DEXes today, so nothing legitimate is dropped; the shared set also guards a future twin listing.

### Regression guards
Same pattern as the marginPct (#482) and candle-key (#484) fixes — fix the code, pin the shape, block the reintroduction, correct the doc that taught it:

- **`senpi-strategy-ops/tests/test_dex_awareness.py`** — 12 tests pinned to the real upstream shapes; every one fails against pre-fix code. Includes two fleet-wide lints so neither idiom can come back.
- **`dex_blind_offenders()`** in *both* validators (author + ops). Fires on both pre-fix files; **0 false positives** across 253 fleet `.py` files.
- **`creating-a-strategy.md`** — all three shapes in one table. This scaffold doc is where the marginPct and candle-key bugs originated too, so it's the actual root-cause surface for authored strategies.

### Verification
- 44 test suites green
- `deploy.py validate` clean on all 13 touched packages
- Skills MINOR-bumped to **2.9.0** (auto-upgrades users); touched strategies patch-bumped; catalog regenerated in both locations (diff is exactly 35 version bumps + date — `crane` is `status:blocked` and correctly skipped)

### Note on scope
The fixes the agent applied during debugging were **local to its own box** — the catalog templates were still shipping all three bugs. This PR is the fleet-wide port.

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -13,7 +13,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.8.0"
+  version: "2.9.0"
   platform: senpi
   exchange: hyperliquid
   requires:
--- a/senpi-strategy-author/references/creating-a-strategy.md
+++ b/senpi-strategy-author/references/creating-a-strategy.md
@@ -84,6 +84,16 @@ No I/O, no MCP, no clock, no state — just functions over candles/numbers, so i
 
 > **Candle schema (`market_get_asset_data`):** each candle is keyed `t,o,h,l,c,v` (+ `T,s,i,n`) — short OHLCV, **string** values. Close is `c`; read fields as `float(candle["c"])`, not `candle["close"]` (no such key).
 
+> **Two sub-DEXes, one wallet (`main` + `xyz`).** Hyperliquid lists ~230 crypto perps on the main DEX and ~100 equity/commodity/index perps on the **xyz** sub-DEX (a large share of volume). They are two views of ONE cross-margined wallet, and each API names them differently. Getting this wrong is **silent** — no exception, no log, the strategy just never trades:
+>
+> | Read | Shape | Consequence of getting it wrong |
+> |---|---|---|
+> | `market_list_instruments(dex="")` | returns **BOTH** DEXes; xyz names carry the `xyz:` prefix in `name`, main names are bare. There is no `dex` field. | xyz names leak into your "main" pool → duplicated universe + misclassified asset class. Route by prefix, or pass an explicit `dex`. |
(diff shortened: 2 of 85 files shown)


betashop:
Split into two PRs so the template repair and the authoring-time prevention can be reviewed by the right people independently:

- **#490 — templates only** (78 strategy files + catalog + the 12-test suite). Pure strategy content, no skill behavior change, self-verifying.
- **#491 — author-side prevention** (scaffold doc + `dex_blind_offenders()` in both validators). Stops a newly authored strategy from reintroducing the bugs.

Neither blocks the other. Same code as reviewed here, just partitioned.

PR #489, opened 2026-07-23
strategy-ops: cli_json falls back to stderr (salvaged from #483)

betashop:
The one piece of #483 that **#486 did not subsume**. #483 is now closed; this is its salvage, rebased onto current main.

### The bug
`openclaw senpi state --json` **exits 0** but writes its JSON payload to **stderr**, while stdout carries only banner/log noise. Measured on M408027's lion deploy: **823B of noise on stdout vs 28,735B of JSON on stderr.**

`cli_json` read stdout only, so `_extract_json(out)` found nothing and returned `None` — for a read that had actually *succeeded*.

### Why it matters after #486
#486 correctly made `status`/`state` **downgrade-only enrichment** behind a fail-open gate. But `state` is the only source of the rich per-scanner row — `runCount`, `lastAliveAt`, `lastError`, `consecutiveErrorCount`. With it permanently unreadable, `_scanner_verdict` never reaches its downgrade evidence and always falls through to `senpi status`.

Failing open is the right default for a liveness gate. It should be the **fallback**, not the only reachable branch — otherwise verify can never catch a genuinely broken scanner.

### The fix
Try stdout, then stderr. A non-zero exit still returns `None`, so a genuine `getSystemState` throw stays unreadable and the verdict fails open exactly as #486 intends.

```python
rc, out, err = run_cli(args, timeout)
if rc != 0:
    return None
return _extract_json(out) or _extract_json(err)
```

### Verification
6 tests pinned to the real stream shape, including stdout-wins-when-both-carry-JSON and the non-zero-exit fail-open case. The noisy-stdout case returns `None` against `origin/main`.

### Notes
- **Durable fix is runtime-side:** send the `--json` payload to stdout. This is a client-side workaround for a stream-routing bug.
- **Version:** 2.8.0 → **2.9.1**. 2.8.1 is claimed by open #487 and 2.9.0 by open #488, so this takes the next free number to avoid the duplicate-version collision that stranded #482/#483.

### DIFF
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -18,7 +18,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.8.0"
+  version: "2.9.1"
   platform: senpi
   exchange: hyperliquid
   requires:
--- a/senpi-strategy-ops/scripts/_cli.py
+++ b/senpi-strategy-ops/scripts/_cli.py
@@ -56,11 +56,20 @@ def _extract_json(text):
 
 
 def cli_json(args, timeout=60):
-    """Run a CLI command expected to emit JSON on stdout; return the parsed object or None."""
-    rc, out, _err = run_cli(args, timeout)
-    if rc != 0 or not out.strip():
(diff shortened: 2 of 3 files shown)


betashop:

PR #490, opened 2026-07-23
strategies: make the fleet dual-DEX (main/xyz) aware — 3 silent bugs

betashop:
**Templates only.** The author-side prevention (scaffold doc + validator guards) is split into a separate PR so this one can move on strategy content alone.

Found while debugging **bobcat** and **gibbon** live. Hyperliquid is two sub-DEXes behind **one** cross-margined wallet — ~230 bare-named main perps plus **103 `xyz:`-prefixed** equity/commodity/index perps (verified against prod MCP today). Three idioms across the fleet were blind to that.

**All three fail silently.** No exception, no error log — the strategy just never trades, or re-opens what it already holds. That's why they survived this long.

Each reproduces against `origin/main`:

| # | Bug | Pre-fix behaviour | Blast radius |
|---|---|---|---|
| 1 | `leaderboard_get_markets` returns a **bare** ticker + a separate `dex` field, but the universe carries `xyz:NVDA` — so `if token != coin` never matched | `bobcat._get_sm_direction("xyz:NVDA")` → `(None, 0.0)` = "no smart-money data". Where SM is a **hard gate**, the entire xyz universe was blocked permanently | **26 scanners** |
| 2 | `strategy_get_clearinghouse_state` returns `{"main":…, "xyz":…}` with `assetPositions` **inside each section**; scanners read them off the top level | `gibbon._held()` → `set()` while holding BTC **and** `xyz:GOLD` → scanner re-opened held names (two failed duplicate `xyz:GOLD` opens) | **11 scanners** |
| 3 | `market_list_instruments(dex="")` returns **both** sub-DEXes, and `_pick`'s dedupe set was per-call | `[('BTC',''), ('xyz:NVDA',''), ('xyz:GOLD',''), ('xyz:NVDA','xyz'), ('xyz:GOLD','xyz')]` — every xyz name duplicated, first copy misclassified as crypto | regime family (chimp/gorilla/gibbon/orangutan, byte-identical `scan.py`) |

### Fixes
1. `_sm_row_matches()` — compares **bare** tickers **and** requires the `dex` to agree, so main `GOLD` can't cross-match `xyz:GOLD`. Also fixes bison's compare, which was additionally case-sensitive.
2. Enumerate **both** sections, with a legacy-flat fallback. `bison` already did this correctly (and documents why `accountValue` must use `max()`, never `sum` — one wallet, two views).
3. Shared `seen` set + prefix-based routing. Confirmed **zero** bare-ticker collisions between the DEXes today, so nothing legitimate is dropped; the shared set also guards a future twin listing.

### Evidence
`senpi-strategy-ops/tests/test_dex_awareness.py` — 12 tests pinned to the real upstream shapes, **every one fails against pre-fix code**, plus two fleet-wide lints so neither idiom can come back. The suite is self-contained: it imports no skill code, only the scanners themselves, so it validates this PR standalone.

- 44 test suites green
- `deploy.py validate` clean on all 13 touched packages
- Touched strategies patch-bumped; catalog regenerated in both locations (diff is exactly 35 version bumps + date — `crane` is `status:blocked` and correctly skipped)

### Scope note
The fixes the agent applied while debugging were **local to its own box** — the catalog templates were still shipping all three bugs. This is the fleet-wide port.

Replaces #488 (split into this + the author-guard PR).

### DIFF
--- a/senpi-strategy-discover/catalog.json
+++ b/senpi-strategy-discover/catalog.json
@@ -1,6 +1,6 @@
 {
   "_version": "2.1",
-  "_updated": "2026-07-13",
+  "_updated": "2026-07-23",
   "_generated": true,
   "_instructions": "GENERATED from each strategy package's strategy.yaml — do not hand-edit; run senpi-trading-runtime/scripts/gen_catalog.py. Agents read this via senpi-strategy-discover (scripts/discover.py). DECLARED fields (archetype, sub_style, asset_classes, asset_scope, risk_level, tier, belief_plain, direction) are author-set in strategy.yaml `catalog:`; DERIVED fields (assets, leverage_max, funding_split, instance_count, cadence_seconds, time_horizon, max_slots) are computed from instances/params. Labels/glosses come from senpi-strategy-discover/references/glossary.yaml. 'min_budget' is the effective floor (max(declared, 100 x instance_count)); positions scale with budget — NOT a hard gate. A strategy is a deployable package; install via senpi-strategy-ops.",
   "skills": [
@@ -10,7 +10,7 @@
       "emoji": "🦅",
       "tagline": "An onboarding-grade breakout trader on the liquid majors (BTC/ETH/SOL): goes LONG when price breaks above its 7-day high AND smart-money is net long, SHORT when it breaks below the 7-day low AND smart-money is net short, then cuts failed breakouts fast (tight 8% Phase 1) while letting a real one run on a wide Phase 2 ratchet.",
       "belief_plain": "Watches BTC, ETH and SOL. When one of them breaks above its highest price of the last week and the best traders are also leaning long, it buys; when one breaks below its weekly low and the best traders are leaning short, it sells short. A fake breakout gets cut quickly, but a real one is given lots of room to run.",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "thesis": "Most breakout strategies lose because they fight the trend, chase fake breakouts into stop-hunts, or hold failed breakouts too long. Hawk requires three things at once — a genuine break of the 7-day range, the 4h trend pointing the same way, and the smart-money leaderboard net in the breakout direction — so it only fires on confirmed breaks. Then the asymmetry does the work: an 8% Phase 1 floor cuts a breakout that fails immediately, while a wide Phase 2 ladder (first lock only at +10% ROE) keeps the rare breakout that runs, because that tail pays for the quiet weeks.",
       "tags": [
         "btc",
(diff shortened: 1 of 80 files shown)


betashop:
**Verification status — read before merging.**

Splitting what was confirmed first-hand from what rests on the debugging transcript, since the three bugs are not equally evidenced:

| # | Bug | Evidence | Confidence |
|---|---|---|---|
| 3 | universe duplication | Called `market_list_instruments(dex="")` against prod directly: **335 rows = 232 bare + 103 `xyz:`-prefixed, no `dex` field**. Ran the pre-fix `derive_universe` from `origin/main` against that live data → literal duplicates, first copy mistagged `dex=''` | **Confirmed first-hand** |
| 2 | dual-DEX positions | The MCP tool's own schema (*"Returns { main, xyz }, each with positions, margin summary, withdrawable"*), bison's implementation + docstring, and a prior measured incident all agree. Confirmed `_read` unwraps the `data` envelope, so the fix operates on the right dict | **Confirmed by API contract**, not re-run live today |
| 1 | leaderboard token match | Transcript only — `leaderboard_get_markets` 401s from my session, so I could not reproduce it | **Second-hand** |

On #1, two things support it without settling it: the transcript reported a *behaviour* change (9/12 names `sm_no_data` → `xyz:TSM SHORT score 8/9`), which a misread schema wouldn't produce; and the agent's fix checked the `dex` field, which would have blocked everything had that field not existed.

**One call settles it:** `leaderboard_get_markets` and paste one raw row for an xyz name. `{"token":"NVDA","dex":"xyz"}` → bug confirmed. `{"token":"xyz:NVDA"}` → those 26 scanners were fine and that part of this PR is a harmless no-op.

**This is safe to merge either way.** `_sm_row_matches` is a strict superset of the old `token != coin` compare — it matches everything the old code matched, plus the bare-token+dex case. Latest push pins **both** plausible row shapes and adds an explicit superset test, so the behaviour is covered rather than assumed. 15 tests.

Bugs #2 and #3 stand on their own regardless.

betashop:
**All three bugs now confirmed against live production data.** This resolves the open item in my previous comment.

**Bug #1 — leaderboard token match: CONFIRMED.** Real row:
```json
{ "token": "NVDA", "dex": "xyz", "direction": "long", "pct_of_top_traders_gain": 0.20094… }
```
Bare ticker + separate `dex` field, exactly as reported. Running that row through both versions:

| | `_get_sm_direction("xyz:NVDA")` |
|---|---|
| `origin/main` | `(None, 0.0)` — sm_no_data, hard gate blocks |
| this PR | `('LONG', 80.08)` |

Main-DEX regression check: BTC returns `('LONG', 63.0)` identically before and after.

**Bug #2 — dual-DEX positions: CONFIRMED**, on a live wallet holding xyz positions (orangutan, `0xe39d…16d1`). The raw response is `{"success":true,"data":{"main":{…},"xyz":{…}}}` with `assetPositions` nested inside each section and **no top-level `assetPositions` key at all**. Feeding the real payload to the scanner:

| | `_held()` |
|---|---|
| `origin/main` | `set()` — while actually holding **5 positions** |
| this PR | `{'BTC', 'ETH', 'XYZ100', 'AAPL', 'BRENTOIL'}` |

Orangutan is one of the four regime templates in this PR, so this is the predicted failure caught live: it believed it held nothing while carrying ~.2M notional across both sub-DEXes, and would re-open every one of those names.

The same payload also confirms the `max()`-not-`sum()` rule: `main.accountValue` 2562.41 vs `xyz.accountValue` 2702.85 — summing gives 5265.26 and would size ~1.95×.

**Bug #3** was already confirmed first-hand against `market_list_instruments`.

No changes needed to the PR — this is confirmation of what's already here.

PR #491, opened 2026-07-23
strategy-author: block the dual-DEX (main/xyz) bug classes at authoring time

betashop:
Rewritten against every point in @sarvesh's review. The core change is the one he recommended: the guard no longer enumerates **wrong** spellings, it checks for **positive evidence of the right thing**.

**Two rules, replacing five regexes**

| | rule |
|---|---|
| leaderboard | a file that *calls* `leaderboard_get_markets` must either consult the row's `dex` or normalise the `xyz:` prefix |
| positions | AST, scoped to the read: `assetPositions` off the raw clearinghouse result, reached outside a both-sections loop |

Any spelling of the bug fails rule 1; any spelling of the fix passes it.

**The six issues**

1. **Branch 16 commits behind (blocking).** It predated #490, so its own tree flagged 37/253 files including `bison` — the package this PR cites as the reference implementation. Rebuilt on current main. The sweep is now reproducible on the branch it ships with (**0 findings across 248 package `.py` files**) and `FleetSweep` asserts it in CI, so the claim can't drift again.
2. **File-wide suppressor → false negative.** Any `"main","xyz"` anywhere in a file disabled the check for that whole file, so a correct file that *grew* a buggy read went unflagged. Now function-scoped and AST-based; `test_a_correct_file_that_grows_a_buggy_read_is_still_caught` covers exactly that case.
3. **False positives on correct code.** `("xyz", "main")` reversed and a generic `.values()` section walk both read as correct now.
4. **Single-spelling token rule.** All six of his variants are caught, including `if token != coin: continue` on one line — which the old regex could not express at all. Every spelling of the *fix* passes.
5. **No market awareness.** Suppressed for packages with no **positive** xyz exposure. Worth noting: the word "xyz" is not exposure — `jaguar` and `vulture` mention it only to *ban* it, and they were the last two false positives before the rule was tightened to positive exposure + `xyzBanned`.
6. **author/ops scope parity.** Both sides now run the identical guard over the identical file set via a shared `dex_scan_files()` — package sources, `tests/` excluded so a deliberate bug fixture isn't a defect. Previously author scanned package-wide including tests, ops scanned only the scanner dir; author-green did not mean deploy-green. `SharedContract` asserts parity.

**Two false-positive classes found while sweeping, both mine:** the legacy-flat fallback #490 deliberately added *beside* a correct read (now allowed — same-function scope), and `leaderboard_get_markets` appearing only in a scoring-module docstring (now requires an actual call, detected via AST).

17 tests, one per case above plus the fleet sweep. 43 suites green. Scaffold doc carries the rule so authors get it before the validator does. author 2.9.0 → 2.10.0, ops 2.8.0 → 2.9.0.


0xsarvesh (commented):
Reviewed intent + approach, and exercised both validators manually across market combinations (both-DEX, main-only, xyz-only) and the detection edge cases.

**Verdict: intent is right, approach has real gaps. Rebase + two fixes before merge.**

## Intent — sound ✅

Fixing the **scaffold doc** is the correct root-cause move, and the evidence supports it: the same doc produced the `marginPct` fraction slip (#482) and the candle-key bug (#484). Patching packages without patching the doc regenerates the bug — which demonstrably happened twice. The doc table is accurate, and splitting prevention (this) from repair (#490) is good hygiene.

## Claims I verified ✅

| Claim | Result |
|---|---|
| "0 false positives across the fleet" | **TRUE** — 0 of **273** `.py` fire against post-#490 `main` |
| Fires on both pre-fix offenders | **TRUE** — bobcat + gibbon both fire |
| author/ops copies identical | **TRUE** — byte-identical (md5 match) |
| Catches the canonical bugs end-to-end | **TRUE** — real scripts flag an injected `if token != coin:` |

---

## Issues

### 1. 🔴 Branch is 13 commits behind — pre-#490. Must rebase.
The "0 FP" evidence **can't be reproduced on this branch**. Running the validator against its own tree flags **37 of 253 files — including `strategies/bison`**, the package this PR names as *"the reference implementation for all three."* It's only clean against post-#490 `main`. This is bigger than the version-collision note already in the description.

```bash
# on this branch (pre-#490): 37 files fire, incl. bison
# against origin/main (post-#490): 0 of 273 fire
```

### 2. 🔴 The `assetPositions` suppressor is file-wide and trivially defeated (false negative)
`_DUAL_SECTIONS` matching `"main", "xyz"` **anywhere in the file** disables the check for that whole file.

Demonstrated live: I appended a genuinely buggy top-level read to the **already-fixed** bison —

```python
def _held(ctx):
    st = ctx.call("strategy_get_clearinghouse_state")
    return [x for x in st.get("assetPositions", [])]   # real bug
```

— and it was **not flagged**, because line 92 already has `for section in ("main", "xyz")`. A stray comment or an unrelated `{"main","xyz"}` literal suppresses it identically.

This is the *most likely real-world regression* — a correct file that later grows a new buggy read — and the guard is blind to exactly that case.

### 3. 🟠 `_DUAL_SECTIONS` false-positives on correct code
Both of these are correct implementations and get flagged:

```python
for sec in ("xyz", "main"):     # reversed order -> FIRES
for sec in state.values():      # no literals    -> FIRES
```

### 4. 🟠 `_RAW_TOKEN_CMP` catches exactly one spelling
All of the following are the same bug and **slip through**:

```python
if token.upper() != coin:
if row["token"] != coin:
if tok != coin:
if token != coin and ok:
if token != coin: continue      # trailing content on the same line
```

Since the stated purpose is stopping a *newly authored* strategy from reintroducing the bug, trivial rewording defeats it.

### 5. 🟡 Token rule has no market awareness
It fires on a pure **main-only** strategy that will never see an `xyz:` name, telling it to handle xyz. Defensible as defense-in-depth (universes change), but it is noise.

### 6. 🟡 author/ops scope parity gap — contradicts "author-green == deploy-green"
Author scans `pkg.rglob("*.py")`; ops scans only the scanner dir. A package-root helper carrying the bug:

```
rootbug    author_dex=1    ops_dex=0     <-- ops is weaker
```

Conversely, author scans `tests/` (25 fleet files live there), so a deliberate bug-fixture in a test would false-positive.

---

## Strongest recommendation: invert the check

Enumerating *wrong* spellings is a losing game (issue 4). Flip to a **positive-evidence** requirement:

> *"uses `leaderboard_get_markets` but never references a `dex` field"*

That is **one** rule that catches every misspelling, instead of five regexes that each catch one. Same for positions: flag a `.get("assetPositions")` whose receiver is the raw clearinghouse result (AST-level), rather than text-suppressing the entire file.

**Blocking:** #1 (rebase onto post-#490 `main`, then re-run the FP sweep to reproduce the claim) and #2 (tighten the suppressor). #3–#6 are worth filing but needn't block.

betashop:

PR #496, opened 2026-07-23
portfolio: answer "why hasn't it traded?" — the `idle` step (P0-2)

betashop:
**P0-2** from the [User Negative Prompts action plan](https://app.notion.com/p/3a61feaac8068016911ac6c74211987d). The #1 support complaint by volume — M192299 asked **five times over three days**; M192397, M298631, M3601, M347101 and M193436 (Turkish) all asked a version of it.

> **Reworked:** this originally landed in `senpi-improve-trades`. That was the wrong skill — see below.

## Why portfolio, not improve-trades

This is a **live-state** question — *is this strategy doing its job right now* — and the skills' own boundary says so: improve-trades states *"Portfolio answers 'how are my strategies doing right now'… use senpi-portfolio for live state."*

More importantly, **portfolio owns the mandate**, and the mandate is what makes the answer good. Its existing rule — *"a strategy is doing its job when its behavior matches its design, even if that design means idle right now"* — **is** the selective-vs-broken distinction. So `waiting` gets narrated against the strategy's own design rather than a generic activity benchmark. improve-trades could only ever give half that answer.

improve-trades now **hands the question over** instead of dead-ending on guardrail 9's *"say 'nothing to review yet', and stop"* — which was the exact non-answer users kept getting.

## Liveness alone cannot answer it

`senpi status` reports the **runtime**, and a runtime can be perfectly healthy while every candidate its scanner produces is rejected downstream. That is exactly how **ibis** shipped 100% dead this morning while every surface said healthy.

The verdict needs the **event ring**, where `signal.outcome` carries the runtime's own ruling on each candidate. Ported `_fetch_events` into `portfolio.py` using its existing fail-open + fixture conventions (mirrors `_fetch_runtime_status`, including the `_telemetry_dead` short-circuit). Engine duplication is the house pattern — both skills already ship their own `_yaml.py` and `mcp_client.py`.

## The step

```sh
python3 scripts/portfolio.py idle
```

| Verdict | Meaning |
|---|---|
| `traded` | opened positions recently — the premise is wrong |
| `blocked` | candidates produced, runtime rejected them **all** → `reason_code` is the answer |
| `accepted_no_open` | accepted but nothing opened (execution/order failure) |
| `waiting` | scanning normally, nothing cleared the bar — **selective, not broken** |
| `silent` | no events at all — *cannot confirm* it's scanning |
| `no_runtime` | funded with no runtime bound — it cannot scan |
| `unknown` | event log unreadable (fail-open; **not** evidence of a fault) |

Each verdict carries the strategy's `mandate`.

## Safety-critical: `waiting` must never read as broken

A selective strategy in a quiet market is working perfectly. Calling that "broken" is how users tear down healthy books — the LION incidents. It has its own test, and the narration rules lead with the mandate plus an **option**, never a fix.

## Verification
- **14 tests** — pure, no MCP and no subprocess, plus an events fixture. All 7 verdicts, the dominant-reason tally, fail-open, fan-out ordering, and that a closed strategy is never probed.
- **43 suites green**
- portfolio **1.9.0 → 1.10.0**; improve-trades **1.6.0 → 1.7.0** (handoff only)

## Not included
The always-on visual timeline is **P0-3** (Trust Roadmap #1, Vysakh) and needs the user-scoped collector read. This is the answer-on-demand half, which ships **without** that enabler.

Independent of #497 — the two can merge in any order.

### DIFF
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -17,7 +17,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.6.0"
+  version: "1.7.0"
   platform: senpi
   exchange: hyperliquid
 ---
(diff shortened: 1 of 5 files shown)


betashop:

PR #501, opened 2026-07-29
fix(strategies): whale-mirror notional shape + case-sensitive coin symbols

0xsarvesh:
Two silent-failure bugs in the strategies that mirror other traders. Neither errors, logs, or fails a validator — they just trade the wrong thing, or nothing.

## 1. `position_notional()` didn't recognise the leaderboard position shape

`scan.py` fetches positions via `leaderboard_get_trader_positions`, which returns:

```json
{ "market": "ETH", "direction": "short", "size": -59267.5639,
  "entry_price": 1869.4, "notional_size": 112270546.30 }
```

`position_notional()` looked for `szi` / `entryPx` / `entryPrice` / `marginUsed` — **none of which exist in that shape**. So `entry` resolved to `0`, `size × entry` was `0`, the `marginUsed` fallback missed too, and it fell through to `return size`: the **raw token count**, returned as if it were USD.

The failure mode differs by how each strategy uses the value:

| | role | effect |
|---|---|---|
| **remora** | ranking (`top_position`) | A 184,160,374-token PUMP position (**$350K**) outranked a 59,268-token ETH short (**$112M**). Remora mirrored meme dust. The real signal — 4 of the top 10 whales short ETH — never cleared `minScore`. Live strategy `feca601f` made 3 trades, all PUMP SHORT. |
| **cuckoo** | filter (`< min_notional`, $2k) | Every memecoin cleared the floor on token count alone, while a $10M BTC position (~100 tokens) fell **below** it. The consensus vote structurally could not see BTC. |

**Fix:** prefer `notional_size` — the documented USD field (`|size| × current_price`) per `senpi://guides/hyperfeed-trader-positions` — and add `entry_price` to the fallback chain. All clearinghouse keys preserved, so `discovery_get_trader_state`-fed callers (oxpecker) are unaffected.

## 2. Coin symbols were upper-cased before being emitted

Hyperliquid coin names are **case-sensitive**. The 1000x-denominated names carry a lowercase `k` (`kPEPE`, `kSHIB`, `kBONK`) and `KPEPE` is rejected as `INVALID_ARGUMENT`; HIP-3 assets carry a lowercase `xyz:` prefix. These symbols are emitted as the signal asset and passed to `market_get_asset_data`, so upper-casing them was a silent no-trade on every affected name.

Not hypothetical for remora: the verified whale book holds `xyz:GOLD` and `xyz:BRENTOIL` in its top 5 positions.

Case is now preserved at the source — `position_asset()` in remora/cuckoo/oxpecker, and the `market_list_instruments` parse in owl/otter. Comparisons still upper-case **at the comparison site**, which is correct.

Three lookups were keyed upper-case against a value that had just stopped being upper-case, and needed correcting or the fix would have introduced a new bug:

- `owl` `sm_map` lookup
- `otter` `sm_map` lookup
- `otter` `_is_cooled_down` — `cooldowns[]` is written upper-case, so without this **`kPEPE` could re-emit inside its own cooldown window**

⚠️ **Side effect:** owl/otter `ctx.state` history + cooldown keys change case, so each does a one-time persistence reset on the first tick after deploy. Self-healing, but their crowding-persistence counters restart.

## Verification

Confirmed against **live** `senpi-dev` `discovery_get_trader_state` for whale `0xb83d…6e36`:

| | `szi` | `entryPx` | `positionValue` |
|---|---|---|---|
| ETH | -59,267.5639 | 1869.42 | **$112,069,036** |
| PUMP | -184,160,374 | 0.001634 | **$345,116** |

Plus the `hyperfeed-trader-positions` guide, which documents the position fields as `market` / `size` / `direction` / `notional_size` / `account_value` / `leverage` — no `szi`, no `entryPx`, no `marginUsed`. The fall-through is provable from the contract, not inferred.

- 11/11 assertions pass across the three scoring modules (leaderboard shape, casing, clearinghouse no-regression, oxpecker `positionValue` precedence intact)
- `validate_strategy.py` ✓ and `validate_universe.py` ✓ on all five packages
- `py_compile` clean

## Versions

`remora 1.1.0→1.1.1`, `cuckoo`/`oxpecker`/`owl`/`otter` `1.0.0→1.0.1`, `catalog.json` regenerated to both locations. Bumped so pre-fix and post-fix trades stay separable in attribution — otherwise the PUMP trades and the corrected trades are indistinguishable in fleet analytics.

## Audit scope — please read before assuming coverage

Checked all 98 strategies / 248 scanner files for related shape-mismatch classes. Clean: nested `leverage` `{type,value}` (all 5 real reads handle it), `coinDisplayName` misuse (zero), clearinghouse parse consistency (all 84 agree). `raptor` and `cheetah` were checked and are safe.

**The casing list is a floor, not a ceiling.** My static tracer follows only direct variable assignments and initially missed remora/cuckoo/oxpecker, whose asset reaches `.upper()` indirectly through `position_asset()`. owl and otter are what it caught plus manual review. Other strategies among the 96 that touch symbol case could have the same issue via paths the tracer can't see — a complete sweep needs per-strategy reading, which is not done here.

**No test coverage exists** for these strategies (no `tests/` dir, no pytest config, nothing referencing `position_notional`). Verification was an ad-hoc harness, not committed. Given this is the *second* shape bug in this one function, real unit tests for `position_notional` are worth adding.

### DIFF
--- a/senpi-strategy-discover/catalog.json
+++ b/senpi-strategy-discover/catalog.json
@@ -1,6 +1,6 @@
 {
   "_version": "2.1",
-  "_updated": "2026-07-23",
+  "_updated": "2026-06-17",
   "_generated": true,
   "_instructions": "GENERATED from each strategy package's strategy.yaml — do not hand-edit; run senpi-trading-runtime/scripts/gen_catalog.py. Agents read this via senpi-strategy-discover (scripts/discover.py). DECLARED fields (archetype, sub_style, asset_classes, asset_scope, risk_level, tier, belief_plain, direction) are author-set in strategy.yaml `catalog:`; DERIVED fields (assets, leverage_max, funding_split, instance_count, cadence_seconds, time_horizon, max_slots) are computed from instances/params. Labels/glosses come from senpi-strategy-discover/references/glossary.yaml. 'min_budget' is the effective floor (max(declared, 100 x instance_count)); positions scale with budget — NOT a hard gate. A strategy is a deployable package; install via senpi-strategy-ops.",
   "skills": [
@@ -223,7 +223,7 @@
       "emoji": "🦉",
       "tagline": "Scans every liquid Hyperliquid crypto perp (OI > $3M) for one-sided crowding — extreme funding + lopsided smart money + concentrated OI — then waits for that crowding to persist 1h+ AND for exhaustion to fire (volume fading, price stalling, RSI divergence) before fading the crowd. Conviction-scaled leverage (7/8/10) at 25% margin, with a macro gate that stands down when BTC's 4h move is extreme.",
       "belief_plain": "The crowd is wrong at the extremes. When everyone piles onto one side of a coin — paying punishing funding, the best traders all leaning the same way, open interest stacking up — and the price then stops following through, the unwind is violent and predictable. Owl waits for both the crowded state AND the exhaustion trigger, then takes the other side. Most days it does nothing, because most days one or both is missing.",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "thesis": "Crowded trades unwind violently. Owl's edge is timing the unwind: it never fades on crowding alone (which can persist and grind), it waits for crowding to PERSIST 1h+ and for distinct exhaustion signals (declining volume under extreme funding, price stalling against the crowd, capitulation wicks, 4h RSI divergence) to confirm the move is dying. Mean reversion fails in trending macro, so a hard gate blocks all fades when |BTC 4h| > 3%. XYZ banned — the funding/smart-money scoring is calibrated on crypto data shape.",
       "tags": [
         "contrarian",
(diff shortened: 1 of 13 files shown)


betashop (approved):
Approved — verified the diff.

- `position_notional` prefers `notional_size` (leaderboard USD field) with abs(), then size×entry_price → marginUsed → size; clearinghouse keys preserved so oxpecker's positionValue path is untouched.
- Case preserved at source, upper-cased at the comparison site (owl/otter sm_map, otter _is_cooled_down). The cooldown re-emit catch is the sharp one.
- PATCH bumps + catalog regen correct.

Follow-ups (non-blocking):
1. Close the loop on live — redeploy the 5 fixed strategies and review/close remora feca601f's 3 erroneous PUMP-SHORT positions.
2. Full per-strategy casing sweep across all 96 (this PR is a floor, not a ceiling) + a committed position_notional unit test. Starting the sweep now.

PR #503, opened 2026-07-30
fix(strategies): fleet casing sweep — 10 more case-sensitivity no-trades + ratchet guard

betashop:
Follow-up to #501. That PR fixed 5 strategies and noted the casing list was "a floor, not a ceiling." This is the sweep: the symbol-upper-casing idiom checked across **all 100 strategies**, the remaining emit-path bugs fixed, and a CI ratchet so it can't recur.

The #501 grep only covered the `coin`/`asset`/`name` idiom — it **missed the `token`/`tok` idiom**, which turns out to be the most common one. That gap is the headline finding here.

## AFFECTED — fixed

An upper-cased symbol reached an **emit** or a **Senpi tool call**, silently no-trading `kPEPE`/`kSHIB`/`kBONK` (and, where not banned, `xyz:` names). Fixed the same way #501 did: preserve case at the source, upper-case only at the comparison/join site.

| Strategy | Bug | Sink |
|---|---|---|
| **cuckoo** | `tally_consensus` re-uppered the asset **after** #501 fixed `position_asset` — defeating that fix downstream. **cuckoo still emitted `KPEPE`** (a #501 miss). | emit `scan.py:307` |
| **raptor** | Unfixed mirror-family sibling of remora. | emit `scan.py:314` + `fetch_current_px` `scan.py:206` |
| **condor** | `market_list_instruments` parse upper-cased (owl/otter class). | emit `scan.py:268` |
| **pangolin** | Instrument parse upper-cased. | `market_get_funding_history` **+** emit `scan.py:276` |
| **meerkat** | `event_asset()` upper-cased at the source (dynamic momentum feed). | emit `scan.py:323` + `market_get_asset_data` `scan.py:202` |

## LOW-RISK config hardening — fixed

Ship clean on majors, but a configured `k`/`xyz:` asset would silently no-trade. Raw-for-emit / upper-for-compare split applied: **stag, scorpion, beaver, hummingbird, koala**. (`koala` is the notable one — its own config invites `xyz:` assets yet upper-cased the HIP-3 prefix.)

## Guard + test (the repo had no CI)

- **`senpi-strategy-author/scripts/lint_casing.py`** — a **ratchet** lint. A regex can't distinguish an *emitted* upper-cased symbol (bug) from an upper-cased *comparison key* (fine) — both look like `x = y.upper()`. So it grandfathers the pre-existing sites (`casing-baseline.txt`) and **hard-fails only on NEW source-side symbol upper-casing.** `# casing-ok: <reason>` exempts a verified-safe line.
- **`.github/workflows/strategy-guards.yml`** — `py_compile` + casing guard + tests. (First CI in the repo; dependency-free.)
- **`strategies/cuckoo/tests/test_scoring_casing.py`** — `position_notional` both shapes (the #501 follow-up) + casing regression. 6/6 pass.

## Verification

All changed packages: `validate_strategy` ✓ · `py_compile` ✓ · casing guard ✓ (0 new) · tests ✓ (6/6). `catalog.json` regenerated to both locations (10 PATCH bumps). No live agent is redeployed by this PR.

## ⚠️ Outstanding — the token/tok second wave

The guard enumerated the true surface: **117 sites across 77 files**, now the baseline backlog. Most are the **safe comparison-key** pattern (e.g. `vulture`: `token_raw == tracked.upper()`), but a few need per-file data-flow tracing before they can be certified — **`osprey` and `piranha` look unresolved** and may be additional emit-path bugs. Those get a dedicated second-wave audit. The ratchet blocks any *new* instance meanwhile, so this backlog can only shrink.

## After merge

Redeploy the 10 fixed strategies (they're live silent-no-trades until redeployed). None hold wrong positions from this class — a silent no-trade is a *missed* trade, not a wrong one — so no position cleanup is needed beyond remora's already-known PUMP shorts from #501's notional bug.


betashop:

PR #512, opened 2026-08-04
docs: senpi validate is the authoring gate (+ per-signal sizing stated positively)

0xsarvesh:
## What

Three skills documented two per-signal sizing keys the runtime does not implement — 5 places in `senpi-trading-runtime` (SKILL.md + scan-contract.md), 2 in `senpi-strategy-author/references/creating-a-strategy.md`.

Worst case was `senpi-trading-runtime/SKILL.md`, which named a nonexistent field as *the* sizing key in the return-value contract and never mentioned `marginPct` at all.

## Why it matters

Failing this way is **silent**. An unrecognised top-level signal key is dropped with only a stderr warning, and sizing falls back to the recipe's configured margin. A scanner written from the old guidance keeps trading normally — just at a size its author never chose.

The repo already knows. Root `CLAUDE.md`, under *"Hard-won invariants (each failed silently in production once)"*, states the contract exactly. This PR makes three skills stop contradicting it.

## Approach — positive, not negative

The obvious fix is "there is no `<field>`". These files are **agent context**, and that phrasing is counterproductive: naming a phantom identifier in order to forbid it puts the token in front of a reader who had never encountered it, and negation is the weakest way to express a constraint.

So the replacement states the contract positively and exclusively:

> **Per-signal sizing is exactly two top-level keys: `marginPct` and `leverage`.** Any other top-level key is **dropped with only a stderr warning**, and sizing falls back to the recipe's configured margin — so an invented sizing field does not fail loudly, it quietly sizes the position differently than you intended.

The general rule — *any* unrecognised top-level key behaves this way — is also more useful than a warning about one particular invention, and it matches an idiom already used elsewhere in the same file (*"Signature is exactly two args"*).

Phrasing verified: the shipped skill text no longer contains the phantom identifiers anywhere.

## Fleet impact: none — this is preventive

Swept `strategies/` before writing:

- **no strategy emits either key at the top level.** The mentions under `strategies/` are v2 to 3.0 port comments and local dollar variables used for affordability maths.
- **`weight` appears only inside declared `data{}` payloads** — legitimate context, not sizing.

The docs pointed at a trap nobody has stepped in yet.

## Also

- Drops a dangling reference to `docs/error-code-taxonomy.md` in `liveness-verification.md` — that file does not exist. The error codes it names are real and stay; only the broken pointer goes.
- **Patch version bumps on all three skills.** A content change that does not bump its frontmatter version silently never ships.

## Test plan

Docs only; no code paths touched. Verified by grep that the shipped skill text contains no phantom identifiers and that the sizing contract is stated positively in all three places.


senpi-strategy-ops/references/liveness-verification.md
```diff
@@ -100,7 +100,7 @@ Failure signatures (**positive** evidence of breakage only — anything else is
 | Symptom | Field signature | Likely cause |
 |---|---|---|
 | Runtime says it's broken | `health === "unhealthy"` (in `status` or `state`) | The runtime's own verdict — trust it; read `lastError` |
-| Crash-looping | `health === "unhealthy"` with a restart count + cause on the scanner's own line in `status` | The supervisor is restarting a rapidly-failing scanner (it keeps retrying at capped backoff — restarts never stop on their own); fix the scanner code. Events carry `senpi.error.code: E_SCANNER_CRASH_LOOP` (tick failures: `E_SCANNER_TICK_ERROR` / `E_SCANNER_TICK_TIMEOUT`) — see `docs/error-code-taxonomy.md` |
+| Crash-looping | `health === "unhealthy"` with a restart count + cause on the scanner's own line in `status` | The supervisor is restarting a rapidly-failing scanner (it keeps retrying at capped backoff — restarts never stop on their own); fix the scanner code. Events carry `senpi.error.code: E_SCANNER_CRASH_LOOP` (tick failures: `E_SCANNER_TICK_ERROR` / `E_SCANNER_TICK_TIMEOUT`) |
```
danielmbirochi: Why this change ?

danielmbirochi (approved):

senpi-strategy-ops/references/liveness-verification.md
0xsarvesh: The path was dangling — `senpi-skills` has no `docs/` directory, so `docs/error-code-taxonomy.md` resolved to nothing for anyone following it. It looks like it was copied from `senpi-trading-runtime`, where that file does exist (it's generated from the code's own error table).

Dropped rather than repointed because the three codes it pointed at — `E_SCANNER_CRASH_LOOP`, `E_SCANNER_TICK_ERROR`, `E_SCANNER_TICK_TIMEOUT` — are already named inline in the same sentence, so the link added nothing the reader didn't already have. A relative path into another repo also can't resolve for an agent loading this skill at runtime.

If you'd rather keep a pointer, happy to make it a full URL to the file in the runtime repo instead.

0xsarvesh:
Pushed the Layer 13 doc half onto this branch — it touches the same six files, so a second branch would have conflicted with every one of them.

**What changed, and why here.** The author skill's definition of "done" was a grep-based linter reporting `0 errors`, followed by a **tiny deploy** to smoke-test. Both halves are backwards: a linter reads the package and never runs it, and funding a wallet to learn whether code executes spends real money to answer what `senpi validate` now answers for free.

So stages 8 and 9 become **advisory lints, then one gate**:

- the lints keep their value — instant, no credentials, catch the obvious before you wait for a tick — but they are explicitly not a verdict
- `senpi validate` is: **PASS** → hand to ops · **UNPROVEN** (exit 2, ran and established nothing — not a pass) · **FAIL** (findings carry what/why/fix against the actual package)

Also here:
- **`ctx.dry_run`** documented in the scan contract, with the reasoning — a scanner that returns early outside its session is right in production, but a tick that returns *without reading* proves nothing, which is how a fatal bug behind a session gate stayed invisible for 18h. Ignoring the flag stays safe; it just means the scanner can't be proven, and validation says so rather than passing it.
- **ops preflight** gains the runnability half beside the structural one
- **`liveness-verification.md`** gains the row it structurally could not have had: a scanner that runs perfectly and reads nothing emits *no failure signature at all*, so there is nothing in that table to look up. Catch it before the wallet exists.

The deeper guide's smoke-test section is rewritten around the same point — its own framing already made the case ("these surface only when the runtime itself executes your code, and they fail silently"), which is exactly what validate does now, minus the deploy.

Versions: author 2.12.0, ops 2.13.0, trading-runtime 3.2.0.

⚠️ **Merge order** — these tell agents to run `openclaw senpi validate`, which ships in [senpi-trading-runtime#292](https://github.com/Senpi-ai/senpi-trading-runtime/pull/292). Reviewable now, but please don't merge ahead of that release or the skills reference a command the host doesn't have.

@danielmbirochi — answered your question on the taxonomy path above; short version is the path was dangling (this repo has no `docs/` dir). Happy to make it a full URL into the runtime repo instead if you'd rather keep a pointer.

senpi-strategy-ops/references/liveness-verification.md
danielmbirochi: The file is here [docs/error-code-taxonomy.md](https://github.com/Senpi-ai/senpi-skills/blob/main/docs/error-code-taxonomy.md)

0xsarvesh: You're right, and I was wrong — sorry. Restored in `2e14866`.

The file is there, added in `19aa5f5` (#505), which is the commit this branch is based on. I checked in a worktree that was pinned to an older commit predating it, saw no `docs/` directory, and reported the path as dangling on that basis without re-checking against the remote. My mistake, and I stated it more confidently than the evidence supported.

Put back as a relative link (`../../docs/error-code-taxonomy.md`) so it resolves from the file that carries it rather than from the repo root — and the target does earn the pointer: it carries a full row per code, which is more than the inline names give.

0xsarvesh:
Widened, since @0xsarvesh is about to test end to end and these would bite. Two rounds of fixes since the approval — worth a re-look at the two ops files.

## 1. I documented enforcement that does not exist (`49de411`)

I had written, in **both** skills, that `deploy.py create` "refuses to fund a package that has not passed" validation. It does not — `deploy.py` has no reference to `senpi validate`, `verifyProof`, or the proof file. The deploy verb is being built toward that, but it is not there yet.

That is the worst shape the error could take: it does not merely misinform, it **licenses the skip**. An agent told the deploy step will re-run the check has a reason to treat the gate as optional — and the skip path ends with `create` funding a wallet and `verify` reporting `live` for a scanner that reads nothing.

Both now state the interim truth, which is a stronger instruction than the claim was: *nothing downstream re-checks this today — **you are the last check before real money***. The hard gate is noted in future tense, with the reason to validate anyway (a package that already passed carries its proof, so the transition costs nothing).

Also added, from reasoning through the unhappy path:
- **a stop condition** — re-running after an edit is not optional (the proof is tied to the exact bytes), but if the same code survives two attempts, stop and report rather than keep editing a package that is already unproven
- **the UNPROVEN dead end** — if it persists after adding `ctx.dry_run`, the finding distinguishes *reads attempted and failed* (auth/service) from *no read attempted* (the bypass is not on the fetching path). And "no setups right now" is explicitly ruled out — a tick that reads and finds nothing returns PASS with a no-signals warning. That is the rationalisation an agent reaches for.

## 2. `lifecycle.md` said the deploy gate was optional (`356376a`)

Pre-existing, found auditing for the same defect class. SKILL.md links to it as the mechanics reference, so it is where an agent goes for detail — and it contradicted SKILL.md on three points:

| | lifecycle.md said | reality |
|---|---|---|
| after `runtime` | *"deployment is DONE — the strategy is live"* | SKILL.md: *"Never report 'live' off `registered` alone"* |
| `verify` | *"**OPTIONAL** (only when the user asks…)"* | SKILL.md: *"the required gate"* |
| how verify judges | *"fast single check of `openclaw senpi state` for a completed tick"* | `deploy.py`: *"none of which depend on the flaky `status`/`state` JSON … does NOT wait for a scan tick"* |

`deploy.py`'s own comment settles it: *"A strategy is live only when every instance is runtime-running + scanner-active + DSL-wired + funded."* All three corrected, plus the header, which justified the three-step split partly by the first tick being a long wait — stale for the same reason.

Relevant beyond tidiness: a sibling doc saying *the other gate is optional* teaches that gates here are negotiable, which undercuts exactly what the validate changes establish.

## Also since approval

- Merged `main` — resolved the `$10`-minimum overlap (kept your floor, kept the validate line) and **a version collision git could not flag**: main's `$10` work took author 2.12.0 / ops 2.13.0, the same numbers this branch had used for different content. Skills auto-upgrade by version, so a host on 2.12.0 would never have seen these changes. Now author 2.13.0, ops 2.15.0, trading-runtime 3.2.0.
- Restored the taxonomy reference — @danielmbirochi was right, the file exists; I had checked a worktree pinned to an older commit.

**Merge order unchanged**: these tell agents to run `openclaw senpi validate`, which ships in [senpi-trading-runtime#292](https://github.com/Senpi-ai/senpi-trading-runtime/pull/292). Please don't land this ahead of that release.

141 skills tests pass.

senpi-strategy-ops/references/liveness-verification.md
danielmbirochi: Confirmed on the branch — the relative link is restored (`liveness-verification.md:105`) and resolves against `docs/error-code-taxonomy.md` on `main`. Resolving.

dmcr (approved):

0xsarvesh:
Pushed `fdd1b49` — four additions to `senpi-strategy-author/SKILL.md` (+25 lines, no deletions), all from field-testing this skill against a live dev agent box today.

## What the testing showed

Three authoring attempts, three claims of success, three packages that did not work:

| # | What the agent said | What was true |
|---|---|---|
| 1 | "Validation passed" | Every tick raised `AttributeError: 'ScanContext' object has no attribute 'call_tool'`. It had run `--stage import`. |
| 2 | "Ready to deploy when you want" | Package written as `runtime/recipe.yaml` — `E_VALIDATE_NO_RECIPE`, exit 3. Unloadable by anything. |
| 3 | PASS (correctly) | Passed with a Supertrend returning one direction for **every** input, so the strategy can never open a long. |

Every one was a single command away from being caught. When shown a real finding, the agent fixed it correctly and immediately — so this is a verification gap, not a code-generation gap.

## Why these edits are not "add another rule"

The rules were **already there and already correct**: FLAT layout in step 2, the gate in step 9, "never tell the user it's ready without PASS" right after it. All three were violated in the same session. A fourth rule in a document whose rules are being skipped changes nothing.

So each edit does one of three things:

1. **Removes a wrong option.** `--stage import` appears at line ~105 as a research aid — hundreds of lines before the gate, and it is the first `senpi validate` a reader meets. Now explicitly marked as not the gate.
2. **Forces evidence into the transcript.** The gate must be reported by quoting `✓ static` / `✓ import` / `✓ live` verbatim. An agent that cannot paste `✓ live` did not run it — now visible instead of invisible.
3. **Corrects a belief.** PASS means *runs*, not *right*. Named both passing-but-broken examples so a green gate does not read as a finish line.

Plus a one-liner mapping `E_VALIDATE_NO_RECIPE` (exit 3) to "your layout is wrong, not your code", pointing at the existing FLAT rule.

Each addition cites the failure that motivated it, so a later reader can tell a scar from a preference.

## Not included

No "check your indicator math" checklist. The correctness gap is real — two of two strategies examined passed the gate with genuine logic bugs — but prose will not close it. That one wants a backtest, which production telemetry already ranks as the #1 missing capability.

Raw artifacts (transcripts, generated packages, validate output) are available if useful for review.

PR #507, opened 2026-07-31
fix(strategies): spider marginPct is a PERCENT, not a fraction

danielmbirochi:
## The bug

The `spider` package on `main` is deploy-blocked. Both legs carry the fraction form of `marginPct` in their scanner inputs, and the percent guard refuses the package before funding:

```
$ python3 senpi-strategy-ops/scripts/deploy.py validate strategies/spider
instance swing: scanners[1].inputs.marginPct must be a PERCENT in (0,100] — set 28 (not 0.28)
instance scalp: scanners[1].inputs.marginPct must be a PERCENT in (0,100] — set 15 (not 0.15)
```

Confirmed live on a dev box today. Nobody can deploy spider from the catalog.

`marginPct` is a PERCENT of withdrawable margin in (0,100] everywhere in the runtime. The fraction form (`0.28` meaning 28%) is the recurring slip that sizes positions **100× too small** — a $5,000 wallet meant to open at $1,400 of margin opens at $14 instead, which then falls under the venue minimum notional and no-trades, or trades a position too small to matter. That's exactly what `margin_fraction_offenders` in `_pkg.py` and `validate_strategy.py` exist to catch pre-funding, and spider's own template tripped it.

## The fix

- `strategies/spider/swing/runtime.yaml`: `marginPct: 0.28` → `28`
- `strategies/spider/scalp/runtime.yaml`: `marginPct: 0.15` → `15`
- Both `scanners/scan.py`: read the input as a PERCENT. `margin_usd` (used only for the venue-min-notional check) is now `account_value * marginPct/100`, and the emitted top-level `marginPct` is the configured percent verbatim rather than being round-tripped back out of `margin_usd`. Added the standard `<=1.0 is a pasted fraction -> x100` defensive guard that marlin / bald-eagle / the other converted packages carry.
- Docstrings and the `strategy.margin_pct` fallback comments no longer teach the fraction form.
- `strategy.yaml` version `6.0.1` → `6.0.2`; `catalog.json` regenerated via `gen_catalog.py` into both locations (`strategies/` and `senpi-strategy-discover/`). Not hand-edited.

[...]

PR #518, opened 2026-08-04
docs+ci: sync README skill versions and auto-sync them going forward

betashop:
The README version table had drifted from each skill's `metadata.version` (the source of truth in every `SKILL.md` frontmatter). This fixes the drift **and** removes the manual step so it can't recur.

## 1 — Corrected the six stale versions
| Skill | README (old) | SKILL.md (actual) |
|---|---|---|
| `senpi-portfolio` | 1.7.1 | **1.10.0** |
| `senpi-improve-trades` | 1.1.1 | **1.7.0** |
| `senpi-strategy-discover` | 2.3.0 | **2.12.0** |
| `senpi-strategy-author` | 2.4.2 | **2.11.0** |
| `senpi-strategy-ops` | 2.2.1 | **2.12.0** |
| `senpi-trading-runtime` | 3.0.2 | **3.1.0** |

The other six already matched. (`improve-trades` had missed six minor releases — this table was clearly being forgotten on every bump.)

## 2 — Made it self-syncing (no more hand-editing)
- **`scripts/gen_readme_versions.py`** — reads every `senpi-*/SKILL.md` `metadata.version` and rewrites the README `Ver` column to match. `--check` mode exits non-zero on drift (for CI). Stdlib only, matches the repo's no-deps convention.
- **`.github/workflows/readme-versions.yml`** — on a PR that touches any `SKILL.md`, regenerates the table and **auto-commits the synced README back to the branch**; on `main` and fork PRs it verifies and fails on drift.
- A comment above the table marks the `Ver` column as generated — do not hand-edit.

Result: bump a skill's `metadata.version`, and the README updates itself. Docs/tooling only, no behavior change to any skill.

### DIFF
--- a/.github/workflows/readme-versions.yml
+++ b/.github/workflows/readme-versions.yml
@@ -1,0 +1,48 @@
+name: README versions
+
+# Keeps the README version table in sync with each skill's SKILL.md
+# metadata.version — the single source of truth. On a PR that bumps a skill
+# version, this regenerates the table and commits the fix back to the branch,
+# so the README is never edited by hand. On push to main it verifies sync.
+
+on:
+  pull_request:
+    paths:
+      - '**/SKILL.md'
+      - 'README.md'
+      - 'scripts/gen_readme_versions.py'
+      - '.github/workflows/readme-versions.yml'
+  push:
+    branches: [main]
+    paths:
+      - '**/SKILL.md'
+      - 'README.md'
+
+permissions:
+  contents: write
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+
+      # PRs from within the repo: auto-fix the README and push it back.
+      - name: Sync + auto-commit (same-repo PRs)
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+        run: |
+          python3 scripts/gen_readme_versions.py
+          if ! git diff --quiet -- README.md; then
+            git config user.name  "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git add README.md
+            git commit -m "docs: auto-sync README skill versions [skip ci]"
+            git push
+          fi
+
+      # Fork PRs (can't push) + push to main: verify only, fail if drifted.
+      - name: Verify in sync
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+        run: python3 scripts/gen_readme_versions.py --check
--- a/.github/workflows/readme-versions.yml
+++ b/.github/workflows/readme-versions.yml
@@ -1,48 +1,30 @@
 name: README versions
 
-# Keeps the README version table in sync with each skill's SKILL.md
-# metadata.version — the single source of truth. On a PR that bumps a skill
-# version, this regenerates the table and commits the fix back to the branch,
-# so the README is never edited by hand. On push to main it verifies sync.
+# The README skill-version table mirrors each SKILL.md metadata.version (the
+# single source of truth). This verifies the mirror. If it fails, run:
+#   python3 scripts/gen_readme_versions.py
+#
+# No `paths:` filter on purpose: the job takes ~2s, and a paths-filtered job can
+# never be a required check (PRs that don't touch a SKILL.md would sit forever
+# on "Expected — waiting for status").
 
 on:
   pull_request:
-    paths:
-      - '**/SKILL.md'
-      - 'README.md'
-      - 'scripts/gen_readme_versions.py'
-      - '.github/workflows/readme-versions.yml'
   push:
     branches: [main]
-    paths:
-      - '**/SKILL.md'
-      - 'README.md'
 
 permissions:
-  contents: write
+  contents: read
+
+concurrency:
+  group: readme-versions-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
 
 jobs:
-  sync:
+  verify:
     runs-on: ubuntu-latest
     steps:
+      # Plain checkout: on pull_request this resolves the merge ref, which works
+      # identically for same-repo and fork PRs.
       - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.head_ref }}
-
-      # PRs from within the repo: auto-fix the README and push it back.
-      - name: Sync + auto-commit (same-repo PRs)
-        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
-        run: |
-          python3 scripts/gen_readme_versions.py
-          if ! git diff --quiet -- README.md; then
-            git config user.name  "github-actions[bot]"
-            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-            git add README.md
-            git commit -m "docs: auto-sync README skill versions [skip ci]"
-            git push
-          fi
-
-      # Fork PRs (can't push) + push to main: verify only, fail if drifted.
-      - name: Verify in sync
-        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-        run: python3 scripts/gen_readme_versions.py --check
+      - run: python3 scripts/gen_readme_versions.py --check
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,6 +43,11 @@ strategies/<id>/                   # a strategy package (e.g. strategies/spider/
   places**: repo `strategies/catalog.json` (source of truth) AND
   `senpi-strategy-discover/catalog.json` (bundled so the catalog travels with the discover skill
   when installed standalone). Keep both in sync by re-running `gen_catalog.py`.
+- **The README skill-version table is GENERATED.** The `Ver` column mirrors each
+  `senpi-*/SKILL.md` `metadata.version` and is written by `scripts/gen_readme_versions.py`
+  (a repo-level generator, alongside `gen_catalog.py`) — never hand-edit it. Bump the version in
+  `SKILL.md` only, then run the script (or let CI do it). `--check` gates this in CI and also fails
+  if a skill has a `SKILL.md` but no README row.
 - **Validate** a package with `senpi-strategy-author/scripts/validate_strategy.py <dir>` and
   `senpi-strategy-ops/scripts/validate_universe.py <dir>` (every hardcoded ticker must be a live
   HL instrument — a dead name silently no-trades; `deploy.py create` runs this as a preflight and
--- a/README.md
+++ b/README.md
@@ -60,17 +60,17 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | Skill | Ver | Role |
 |---|---|---|
 | **Analyze** | | |
-| [`senpi-portfolio`](senpi-portfolio/) | 1.7.1 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
+| [`senpi-portfolio`](senpi-portfolio/) | 1.10.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
 | [`senpi-market-pulse`](senpi-market-pulse/) | 1.1.1 | Daily cross-asset market read (crypto, equities, commodities, macro, funding regime) |
 | [`senpi-smart-money`](senpi-smart-money/) | 1.1.1 | Where the most-profitable wallets are positioned vs. the crowd |
 | [`senpi-trader-research`](senpi-trader-research/) | 1.0.2 | Rank + vet Hyperliquid traders before copying them |
-| [`senpi-improve-trades`](senpi-improve-trades/) | 1.1.1 | Retrospective review + health checks off the **telemetry event log**: exit quality, missed signals, leaks, crashes, "if I'd held" counterfactual |
+| [`senpi-improve-trades`](senpi-improve-trades/) | 1.7.0 | Retrospective review + health checks off the **telemetry event log**: exit quality, missed signals, leaks, crashes, "if I'd held" counterfactual |
 | [`senpi-account-status`](senpi-account-status/) | 1.1.1 | Points, loyalty tier, fees, Arena standing, referrals |
 | **Run a strategy** | | |
-| [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.3.0 | Conversational picker — rank the catalog against your worldview |
-| [`senpi-strategy-author`](senpi-strategy-author/) | 2.4.2 | Build/edit a DSL-protected strategy package, one decision at a time |
-| [`senpi-strategy-ops`](senpi-strategy-ops/) | 2.2.1 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
-| [`senpi-trading-runtime`](senpi-trading-runtime/) | 3.0.2 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
+| [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.12.0 | Conversational picker — rank the catalog against your worldview |
+| [`senpi-strategy-author`](senpi-strategy-author/) | 2.11.0 | Build/edit a DSL-protected strategy package, one decision at a time |
+| [`senpi-strategy-ops`](senpi-strategy-ops/) | 2.12.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
+| [`senpi-trading-runtime`](senpi-trading-runtime/) | 3.1.0 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
 | **Move money / positioning** | | |
 | [`senpi-deposit-withdraw-transfer`](senpi-deposit-withdraw-transfer/) | 1.0.1 | The money-movement rails (funds in via embedded wallet; out via the app) |
 | [`senpi-why`](senpi-why/) | 1.0.3 | "Why Senpi / vs. other tools" — the positioning answer |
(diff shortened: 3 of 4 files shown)


shnoodles (commented):
## Review + proposed fixes

Reviewed against `main` @ `cb00f058`.

**Verdict: merge it, with fixes.** The premise is real and the six version corrections are all
correct — I reverted `README.md` to main's version, ran the new generator, and the resulting table
is byte-identical to the PR's. `improve-trades` genuinely had missed six minor releases.

Everything below is either a **fix to apply** or a **design call for the author**. All patches in
this doc were run and tested; results are in the test matrix at the bottom.

---

## Summary

| # | Item | Where | Severity |
|---|---|---|---|
| 1 | Fork PRs die at checkout → verify step never runs | workflow | **fix before merge** |
| 2 | Auto-commit doesn't re-trigger checks → can't be a required check | workflow | **fix before merge** |
| 3 | Auto-commit vs. `--check`-only | workflow | design call |
| 4 | Reverse drift unguarded (new skill missing from table passes CI) | script | fix |
| 5 | Non-triple versions silently truncated (`2.12.0-rc1` → `2.12.0`) | script | fix |
| 6 | `split("---", 2)` isn't line-anchored | script | fix (latent) |
| 7 | `CLAUDE.md` not updated | docs | fix — highest value/effort ratio |
| 8 | Nits: `push:` paths, `concurrency:`, root `scripts/` | both | nit |

What's already right and should not change: the script is **idempotent** (byte-compared across a
no-op run), it **fails closed** on a missing version (exit 2 in *both* modes, returning *before*
`write_text`, so it never leaves a half-synced README), `--check` prints a per-skill `old -> new`
plus the exact fix command, it's stdlib-only per repo convention, and the `<!-- do not hand-edit -->`
comment mirrors the existing `strategies/catalog.json` convention.

---

## 1. Fork PRs fail at checkout, so the fork path never runs

`.github/workflows/readme-versions.yml:28`

```yaml
- uses: actions/checkout@v4
  with:
    ref: ${{ github.head_ref }}
```

With no `repository:`, checkout defaults to the **base** repo. On a fork PR `github.head_ref` is
the branch name *in the fork*, which doesn't exist in the base repo — so checkout fails at step 1
and the `Verify in sync` step, the entire reason the fork branch of the `if:` exists, never
executes.

It fails closed (red X, nothing passes silently), but the error is unrelated to drift and it blocks
every fork PR that touches a `SKILL.md`. Fixed by the job split in §3.

## 2. The auto-commit can never be a required check

Pushes made with the default `GITHUB_TOKEN` **do not trigger workflow runs** (`[skip ci]` reinforces
it). So after the bot commits, the PR's new head SHA has no run for this workflow.

Harmless today. The moment this becomes a *required* check, every auto-synced PR parks forever on
"Expected — waiting for status". Either use a PAT/app token (§3, Option B) or drop the auto-commit
(§3, Option A).

## 3. Design call: auto-commit vs. `--check`-only

Auto-commit buys `contents: write` on a `pull_request`-triggered workflow that executes a script
from the PR head, plus push races (no `concurrency:` group, no rebase before `git push`), plus
non-fast-forward surprises for anyone mid-branch — to save one command.

To be fair on the security framing: **there is no privilege escalation here.** GitHub downgrades
fork tokens to read-only regardless of the `permissions:` block, and same-repo branch pushers
already have write. It's blast-radius hygiene, not a vulnerability.

Given this is the repo's first-ever workflow (`.github/` did not exist before this PR), I'd start
with Option A.

### Option A — recommended: verify only

Replace `.github/workflows/readme-versions.yml` with:

```yaml
name: README versions

# The README skill-version table mirrors each SKILL.md metadata.version (the
# single source of truth). This verifies the mirror. If it fails, run:
#   python3 scripts/gen_readme_versions.py
#
# No `paths:` filter on purpose: the job takes ~2s, and a paths-filtered job can
# never be a required check (PRs that don't touch a SKILL.md would sit forever
# on "Expected — waiting for status").

on:
  pull_request:
  push:
    branches: [main]

permissions:
  contents: read

concurrency:
  group: readme-versions-${{ github.head_ref || github.ref }}
  cancel-in-progress: true

jobs:
  verify:
    runs-on: ubuntu-latest
    steps:
      # Plain checkout: on pull_request this resolves the merge ref, which works
      # identically for same-repo and fork PRs.
      - uses: actions/checkout@v4
      - run: python3 scripts/gen_readme_versions.py --check
```

### Option B — keep auto-commit, made fork-safe

Two jobs, because same-repo and fork PRs need different checkouts:

```yaml
name: README versions

on:
  pull_request:
  push:
    branches: [main]

concurrency:
  group: readme-versions-${{ github.head_ref || github.ref }}
  # Not cancel-in-progress: never kill a job mid-push.
  cancel-in-progress: false

jobs:
  sync:
    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
    runs-on: ubuntu-latest
    permissions:
      contents: write
    steps:
      - uses: actions/checkout@v4
        with:
          ref: ${{ github.head_ref }}
          # A PAT/app token makes the auto-commit re-trigger checks; with the
          # default GITHUB_TOKEN the pushed commit triggers nothing, so this
          # workflow cannot be a required check. Converges either way — the
          # script is idempotent, so a re-triggered run finds no drift.
          token: ${{ secrets.README_SYNC_TOKEN || secrets.GITHUB_TOKEN }}
      - name: Sync + auto-commit
        run: |
          python3 scripts/gen_readme_versions.py
          git diff --quiet -- README.md && exit 0
          git config user.name  "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
          git add README.md
          git commit -m "docs: auto-sync README skill versions"
          # Rebase first: the branch may have moved since checkout.
          git pull --rebase --autostash origin "${{ github.head_ref }}"
          git push origin "HEAD:${{ github.head_ref }}"

  verify:
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
    runs-on: ubuntu-latest
    permissions:
      contents: read
    steps:
      - uses: actions/checkout@v4
      - run: python3 scripts/gen_readme_versions.py --check
```

Both files parse as valid workflow YAML, and the `run:` blocks pass `bash -n`. The
`git diff --quiet … && exit 0` guard is safe under Actions' default `bash -e`: a command failing
inside a `&&` list is exempt from `set -e` unless it follows the final `&&`, so a dirty README
falls through to the commit path rather than aborting the step. Verified both branches.

---

## 4–6. Script fixes

All three are silent-failure modes, which is the exact class `CLAUDE.md` is emphatic about. Applied
patch: **+41 / −10** on `scripts/gen_readme_versions.py`.

### 4. Reverse drift is unguarded

Verified: dropped in a `senpi-newthing/SKILL.md` with a version and no README row — `--check`
printed `README versions in sync ✓` and exited 0. The script only corrects rows that *already
exist*, so a brand-new skill silently never reaches the table. That's the same failure class the PR
is fixing, so "the table can never drift by hand again" is only half-true.

Fix: glob `senpi-*/SKILL.md` (not `senpi-*/`) and assert every hit has a row. Globbing on the
`SKILL.md` rather than the directory is deliberate — it means a dir without a `SKILL.md` (a local
untracked scratch dir, say) can't trip the check.

### 5. Non-triple versions truncate silently

Verified: set `metadata.version: "2.12.0-rc1"`; `--check` reported in sync while the README showed
`2.12.0`. `[0-9]+\.[0-9]+\.[0-9]+` grabs the numeric prefix and the trailing `["\']?` happily
matches nothing.

Two halves to this. In `VER`, capture the whole value and anchor to end-of-line. In `ROW`, match the
version cell as "anything but a pipe" instead of a semver triple — otherwise a malformed, empty, or
pre-release cell falls out of the regex entirely and the row is skipped with no error at all.

### 6. `split("---", 2)` isn't line-anchored

A `---` anywhere inside frontmatter truncates the parse, which is a live risk given several skills
have 10+ line folded `description: >-` prose blocks. I checked all 12 `SKILL.md` files: none
contain it today and all parse correctly, so this is latent, not broken. Line-anchoring costs
nothing.

### Patch

```diff
--- a/scripts/gen_readme_versions.py
+++ b/scripts/gen_readme_versions.py
@@
 the table can never drift by hand again.
 
+It fails loudly (never silently) in all four drift directions:
+  * README row version is stale                -> rewritten (--check exits 1)
+  * README row's SKILL.md has no version       -> exit 2
+  * README row for a skill dir that is gone    -> exit 2
+  * skill has a SKILL.md but no README row     -> exit 2
+
+Dirs without a SKILL.md are not distributed skills and are ignored.
+
 Usage:
@@
 # A skill row: | [`senpi-x`](senpi-x/) | 1.2.3 | description... |
-ROW = re.compile(r'^(\| \[`(senpi-[a-z0-9-]+)`\]\([^)]*\) \| )([0-9]+\.[0-9]+\.[0-9]+)( \| )')
-VER = re.compile(r'^\s*version:\s*["\']?([0-9]+\.[0-9]+\.[0-9]+)["\']?', re.M)
+# The version cell is "anything but a pipe" rather than a semver triple, so a
+# malformed / empty / pre-release cell gets corrected instead of silently
+# falling out of the regex and being skipped.
+ROW = re.compile(r'^(\| \[`(senpi-[a-z0-9-]+)`\]\([^)]*\) \| )([^|]*?)( \| )')
+
+# YAML frontmatter, line-anchored: a `---` inside a folded description block
+# cannot truncate the parse.
+FRONT = re.compile(r'\A---[ \t]*\r?\n(.*?)^---[ \t]*$', re.S | re.M)
+
+# Captures the whole value including any pre-release suffix, so `2.12.0-rc1` can
+# never be silently truncated to `2.12.0`.
+VER = re.compile(r'^[ \t]*version:[ \t]*["\']?([^"\'\s]+)["\']?[ \t]*$', re.M)
 
 
 def skill_version(skill_id: str):
     sk = ROOT / skill_id / "SKILL.md"
     if not sk.exists():
         return None
-    # only the YAML frontmatter (between the first two --- fences)
-    text = sk.read_text(encoding="utf-8")
-    parts = text.split("---", 2)
-    front = parts[1] if len(parts) >= 3 else text[:2000]
-    m = VER.search(front)
+    front = FRONT.match(sk.read_text(encoding="utf-8"))
+    if not front:
+        return None
+    m = VER.search(front.group(1))
     return m.group(1) if m else None
 
 
+def skills_on_disk() -> set:
+    """Every distributed skill = a senpi-*/ dir that has a SKILL.md."""
+    return {p.parent.name for p in ROOT.glob("senpi-*/SKILL.md")}
+
+
 def main() -> int:
     check = "--check" in sys.argv
     lines = README.read_text(encoding="utf-8").splitlines(keepends=True)
-    out, drifted, missing = [], [], []
+    out, drifted, missing, seen = [], [], [], set()
     for line in lines:
         m = ROW.match(line)
         if not m:
             out.append(line); continue
         skill_id, cur = m.group(2), m.group(3)
+        seen.add(skill_id)
         actual = skill_version(skill_id)
         if actual is None:
             missing.append(skill_id); out.append(line); continue
         if actual != cur:
-            drifted.append(f"{skill_id}: {cur} -> {actual}")
+            drifted.append(f"{skill_id}: {cur or '(empty)'} -> {actual}")
             line = ROW.sub(rf'\g<1>{actual}\g<4>', line, count=1)
         out.append(line)
 
     if missing:
-        print("ERROR: no metadata.version found for: " + ", ".join(missing), file=sys.stderr)
+        print("ERROR: no metadata.version found for: " + ", ".join(sorted(missing)),
+              file=sys.stderr)
         return 2
+    unlisted = skills_on_disk() - seen
+    if unlisted:
+        print("ERROR: skill(s) have a SKILL.md but no README table row: "
+              + ", ".join(sorted(unlisted)), file=sys.stderr)
+        print("Add a row to the skill table in README.md.", file=sys.stderr)
+        return 2
     if check:
```

---

## 7. `CLAUDE.md` isn't updated — highest value per line

`CLAUDE.md` exists precisely because "these rules have been forgotten or stripped during past
rewrites", and it already documents the other generated artifact by name
(`strategies/catalog.json` … `gen_catalog.py`). A new generated column plus the repo's first CI
belongs there, or the next AI editor hand-edits the table and undoes this PR. The README HTML
comment only helps someone already looking at the README.

Suggested addition, alongside the existing catalog bullet:

```markdown
- **The README skill-version table is GENERATED.** The `Ver` column mirrors each
  `senpi-*/SKILL.md` `metadata.version` and is written by
  `scripts/gen_readme_versions.py` — never hand-edit it. Bump the version in
  `SKILL.md` only; run the script (or let CI do it). `--check` mode gates this in
  CI, and also fails if a skill has a `SKILL.md` but no README row.
```

## 8. Nits

- **`push:` paths omit `scripts/gen_readme_versions.py`** — a script change landing on `main` isn't
  verified, asymmetric with the `pull_request` filter which does include it. Moot under Option A,
  which drops `paths:` entirely.
- **No `concurrency:` group** — two quick pushes race and one loses. Both options above add one.
- **Root `scripts/` is a new location.** Every other script in the repo lives in
  `<skill>/scripts/`, and `CLAUDE.md` names `senpi-trading-runtime/scripts/gen_catalog.py` as the
  repo-level generator. Root is defensible for a repo-level tool — just be deliberate, and say so
  in `CLAUDE.md` (§7) so the next generator lands in the same place.

## Out of scope, worth queueing

- **`strategies/catalog.json` is the generated artifact where drift is *functional*** — discover
  ranks it, and `CLAUDE.md` requires two copies stay in sync. I checked: both copies match and
  `gen_catalog.py` reproduces them cleanly, so nothing is broken today. But a `gen_catalog.py
  --check` in this same workflow would guard something that changes behavior, whereas a README
  column doesn't.
- **46 test files, zero CI before this PR.** Running pytest is arguably the more valuable first
  workflow.

---

## Test matrix

Run against the PR branch with the §4–6 patch applied. `main` @ `cb00f058`.

| # | Case | Old script | Patched |
|---|---|---|---|
| T1 | `--check` on PR branch as-is | pass | pass (exit 0) |
| T2 | Regenerate from main's stale README | matches PR table | matches PR table |
| T3 | Idempotent across a no-op run | yes | yes |
| T4 | `metadata.version: 2.12.0-rc1` | **silent pass (exit 0)** | caught, exit 1 |
| T5 | New skill w/ `SKILL.md`, no README row | **silent pass (exit 0)** | caught, exit 2 |
| T6 | `SKILL.md` missing `version:` key | exit 2, README untouched | exit 2, README untouched |
| T7 | Dir with no `SKILL.md` present | ignored | ignored (no false positive) |
| T8 | Empty version cell in README | **row silently skipped** | corrected → `1.0.3` |
| T9 | README row for a deleted skill dir | exit 2 | exit 2 |

Verified independently of the PR's own claims: all 12 rows checked against real `metadata.version`
values at latest `main`, and the six corrections (`portfolio` 1.10.0, `improve-trades` 1.7.0,
`strategy-discover` 2.12.0, `strategy-author` 2.11.0, `strategy-ops` 2.12.0, `trading-runtime`
3.1.0) are all accurate. The PR branch is already current with `main` — no rebase needed.

shnoodles:
## Follow-up on `afdc43ad` — satisfies the review

Re-verified against this head rather than taking the commit message's word for it. All eight items addressed:

| # | Item | Status |
|---|---|---|
| 1 | Fork-PR checkout failure | Fixed — plain `actions/checkout@v4`, no `ref:`; resolves the merge ref for forks and same-repo alike |
| 2 | Auto-commit can't be a required check | Fixed — auto-commit dropped |
| 3 | Auto-commit vs. verify-only | Verify-only adopted; `contents: read`, `concurrency:` group, no `paths:` filter |
| 4 | Reverse drift unguarded | Fixed — skill with a `SKILL.md` and no README row now exits 2 |
| 5 | Pre-release truncation | Fixed — `2.12.0-rc1` exits 1 instead of silently passing |
| 6 | `split("---", 2)` not line-anchored | Fixed — `FRONT` regex is line-anchored |
| 7 | `CLAUDE.md` not updated | Added, correctly placed next to the `gen_catalog.py` bullet |
| 8 | `push:` paths / `concurrency:` / root `scripts/` | Resolved; root location now documented |

**Nine-case matrix re-run against this head, all pass:** `--check` clean on the branch (exit 0) · regenerating from main's stale README still reproduces this table identically · idempotent across a no-op run · pre-release → exit 1 · unlisted skill → exit 2 · missing `version:` → exit 2 **with the README untouched** (no partial write) · empty version cell → corrected · deleted skill dir → exit 2.

CI green in 5s ([run](https://github.com/Senpi-ai/senpi-skills/actions/runs/30956194632/job/92149688494)) — the repo's first passing workflow. Versions still accurate; `main` hasn't moved (`cb00f058`), branch is 3 ahead / 0 behind.

### Three doc-accuracy nits left — none block merge

**1. `CLAUDE.md:49` — "then run the script *(or let CI do it)*."** This one's mine: I wrote that parenthetical while the auto-commit option was still on the table, and it got adopted verbatim. With verify-only CI, a forgotten bump now gets a red X, not a fix — which is exactly the wrong expectation to plant in the file future editors read.

```bash
sed -i '' 's/then run the script (or let CI do it)\./then run the script./' CLAUDE.md
```

**2. `README.md:60` — "The `Ver` column is *auto-synced* … (enforced in CI)."** "Enforced" is right; "auto-synced" now overpromises. `generated` is accurate.

**3. Script docstring** trimmed from four drift directions to three, dropping "README row for a skill dir that is gone → exit 2." The behavior still works (verified, exit 2) — it's just no longer documented.

---

### Separable: the `CLAUDE.md` header

The 5-line addition in this PR is the right call. Raising this here only because the PR touches the file — happy to split it into its own PR.

Lines 1–3 currently read:

> This repo is almost entirely AI-generated. The rules below exist because they have been forgotten or stripped during past rewrites. Preserve them on every edit.

The **rules** are load-bearing, and history proves it. `marginPct` is-a-percent-not-a-fraction was fixed in `885a9194`, tightened in `9de21f55`, landed as #482 — then recurred and was fixed again in `c344b23a` (#507), 25 PRs later. Coin casing / dex prefix took three waves: `3a545827` ("3 silent bugs"), `f1dd5773` ("10 more case-sensitivity no-trades"), `e2304124` ("casing sweep wave 2"). Every one is a strategy that deploys, ticks clean, and trades nothing. Those rules should stay exactly as they are.

The **preamble** is doing less work. The only operative words are "Preserve them on every edit" — the AI-generated claim is backstory, it changes no editor behavior, and on a public MIT repo of templates that move real money it's a disclosure choice rather than a requirement. Suggested replacement:

```markdown
# Senpi Skills — Repo Conventions

Preserve these on every edit. Each rule below corresponds to a bug that shipped silently — a
strategy that deploys, ticks clean, and trades nothing. They have been dropped in past rewrites
and had to be re-fixed.
```

Same priority signal and same "this recurs" weight, two lines shorter, no self-narration. Keeps all 178 lines of rules untouched.

betashop:

PR #185, opened 2026-04-15
changes to add skill attribution to all the skills

Hrishikesh-Thakkar:
> [!NOTE]
> **Low Risk**
> Low risk because changes are documentation-only and do not modify trading logic or runtime behavior; main risk is accidental mismatches between `skill_version` values across files.
> 
> **Overview**
> Introduces `CLAUDE.md` repo conventions requiring **strategy-representing** skills to pass `skill_name` and `skill_version` on all `strategy_create`/`strategy_create_custom_strategy` calls.
> 
> Backfills this convention across many strategy skills by adding a `## Skill Attribution` section to each `SKILL.md` and creating `references/skill-attribution.md` files containing the canonical JSON snippet with that skill’s name/version.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9eb9d9ee2549f0b6abb8b1deff708d485125ee48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #519, opened 2026-08-04
$10 minimum rollout: per-template calculated minimums [DO NOT MERGE until backend live]

betashop:
## ⏳ DO NOT MERGE until the backend `initialBudget` $100→$10 deploy is live (Vignesh, Wed)
Skills-side handling must follow the backend, never precede it. Pairs with **senpi-hyperliquid-mcp#185** (MCP) + **infra#1724** (prod env). Constants confirmed against #185: floor is `$10` (env `MINIMUM_STRATEGY_BUDGET` default), below-floor is `SERR037`.

## What
Platform floor **$100 → $10 per wallet**, the authored `min_budget` concept **deleted**, replaced by a **machine-computed minimum per template** — the smallest budget at which the design functions. Users size their own budget; we show the floor.

## The number
```
per_wallet_min = max($10, 12 / (marginPct/100 × min_leverage) + $1.50 fee)
strategy_min   = max over wallets( per_wallet_min ÷ funding_share ), floored $10×wallets, rounded up to a clean step
```
`min_budget.py` — **pure**, vendored **byte-identically** into gen_catalog + deploy, guarded by a parity checksum test. Platform constants ($10/$12/$1.50) consolidated there.

## Changes
- **gen_catalog** bakes `min_budget` + `wallet_count` + `binding_wallet` + per-wallet `breakdown` into catalog.json; drops `max(declared, 100×instances)`.
- **deploy.py** — `MIN_WALLET` 100→10 (imported); two-tier gate: hard block `< $10×wallets` (`E_FUNDS_BELOW_FLOOR`), **soft warn** `< computed min` (`E_BUDGET_BELOW_STRATEGY_MIN`, names the binding sleeve) → proceeds.
- **discover.py** — cards read the baked min → "Minimum $200 (3 wallets)"; above-budget strategies are **ranked + caveated, never filtered**; recommended-budget phrasing removed.
- **Deleted authored `min_budget`** from all strategy.yamls; optional `min_budget_floor` can only **raise** the computed value.
- **Golden-file test** locks all 101 minimums (a marginPct retune → visible diff) + guards re-authoring. Deploy-gate + discover tests re-parameterized to the $10 floor. Docs swept (SKILL.mds ×3, schema ×2, catalog-fields, creating-a-strategy, error-code-taxonomy + new code, CLAUDE.md).

## Where the 99 land (formula-faithful, per the decision)
14→13 run at $10 · 43 at $15–25 · 30 at $30–50 · rest $60–250. **Multi-sleeve minimums compute higher than the preview where the preview under-counted** (it didn't apply `÷funding_share` to the binding sleeve): **cub $200** (10%-share pre-IPO sleeve can't fund below), ox $250, lion/mongoose/hydra/asia-ai/elephant/spider/dragonfly up, ram $50. The 5 new templates: shadow/rotator/hare $20, kite $40, **mandate $250** (no-leverage 5% slots). All locked in `min_budget_golden.json`.

## Verification (offline, all green)
- gen_catalog regen **clean, 0 warnings**; **golden 101/101**; **vendor parity identical**; ops + discover suites **13/13**; all scripts compile.

## Runtime-team follow-up (not this repo)
The `effective_slots` slot-clamp for graceful below-min degradation lives in the **runtime plugin scaffold**. The hard gate at the computed min doesn't need it (at/above min every sleeve funds); it's what makes the *soft-warn* tier degrade cleanly instead of throwing `position_open_failed`. Flagging for the runtime team.


senpi-strategy-ops/scripts/min_budget.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def _f(x):
+    try:
+        return float(x)
+    except (TypeError, ValueError):
+        return None
+
+
+def _round_step(x):
```
0xsarvesh, line 52: **Blocking — this rounds DOWN above the cap.**

```
_round_step(271.67) -> 250
_round_step(1000)   -> 250
```

`ox` already hits it: raw **$271.67 → shipped $250**, so its `ballast` sleeve cannot fund at the advertised minimum. The docstring says "rounded up to a clean step", but past `STEPS[-1]` it clamps.

Suggest `ceil` to the next $25 rung beyond the last step instead of `return STEPS[-1]`.

_(Applies identically to the vendored copy in `senpi-trading-runtime/scripts/min_budget.py`.)_

senpi-strategy-ops/scripts/min_budget.py
```diff
@@ -0,0 +1,159 @@
...
+
+def _scanner_inputs(runtime):
+    """The first external_scanner's `inputs` map (where per-signal sizing tunables live)."""
+    for sc in (runtime.get("scanners") or []):
+        if isinstance(sc, dict) and sc.get("type") == "external_scanner" and isinstance(sc.get("inputs"), dict):
+            return sc["inputs"]
+    return {}
+
+
+def _margin_pct(strat, inp):
```
0xsarvesh: **Blocking — this takes the declared *baseline*, not the *smallest* slot.**

The module docstring commits to worst-case ("its smallest slot ... and its LOWEST leverage"). Leverage is worst-cased in `_min_leverage`; margin is not — `strategy.margin_pct` wins before `marginPctTiers` is ever consulted. Ten templates ship a minimum below their own smallest slot:

| id | used | smallest real slot | shipped | should be |
|---|---|---|---|---|
| shadow | 15% | `minMarginPct: 8` | **$20** | $40 |
| ant | 8% | `marginPctTiers.base: 6` | **$100** | $125 |
| gorilla, orangutan | 12% | `tiers.base: 8` | **$40** | $60 |
| chimp, gibbon, raven, starling, viper | 10% | `tiers.base: 7` | **$50** | $60 |

Verified in the scanners, not inferred: `strategies/ant/main/scanners/scoring.py:218` emits the per-signal `marginPct` straight from `marginPctTiers`, and `strategies/shadow/main/scanners/scoring.py:114` floors emitted margin at `minMarginPct`.

⚠️ Whoever fixes this: `sailfish`'s `leaderMarginPct: 1.5` looks like a hit but is **not** sizing — it's an RS-gap threshold in percentage points. A naive "min over every `*MarginPct*` key" drops sailfish $25 → $250. Needs an explicit key allowlist, not a regex.

senpi-strategy-ops/scripts/min_budget.py
```diff
@@ -0,0 +1,159 @@
...
+    per_wallet_over_share = []
+    breakdown = []
+    for inst in insts:
+        rt = runtimes.get(inst.get("name")) or {}
+        strat = rt.get("strategy") or {}
+        inp = _scanner_inputs(rt)
+        mp = _margin_pct(strat, inp)
+        lev = _min_leverage(strat, inp)
+        fs = _f(inst.get("funding_share")) or 1.0
+        pw = _per_wallet_min(mp, lev) if mp else WALLET_FLOOR
```
0xsarvesh: **Blocking — an unresolvable `marginPct` collapses to the $10 floor, silently.**

Three vol-parity sleeves resolve to `None` today — `caribou/long`, `caribou/short`, `hydra/hedge` — because their sizing lives in `baseRiskPct` / `minMarginPct` / `hedgeRiskPct`, none of which `_margin_pct` knows. **caribou ships `min_budget: 20`**; at its own `minMarginPct` (3%) x 3x the real number is ~$270.

`gen_catalog` warns on an unreadable `runtime.yaml` but is silent here — which is why the regen reports "0 warnings".

Blast radius today is nil (caribou + hydra are already blocked by an unrelated pre-existing `margin_fraction_offenders` validator failure), but this is the scaling failure mode: the next vol-parity template silently advertises $10/wallet and then no-trades.

Suggest returning the unresolved wallets from `strategy_min_budget` so `gen_catalog` can warn and `deploy` can refuse to assert a number it never computed.

0xsarvesh (commented):
## Deep review — scope: strategy-ops, discover, author

Reviewed `953c115` (the only commit over `origin/main`) in a clean worktree, ran every affected suite, and re-derived the minimums independently from the packages.

### Verdict

The architecture is right — one pure function, vendored with a checksum test, golden-locked, baked into the catalog, recomputed locally at deploy so "what the card promised == what deploy enforces." Regeneration is deterministic and both catalogs are byte-identical and in sync. All suites green locally: ops 72+15+10+28+11, discover 196, golden 101/101, vendor parity OK. Nothing here touches runtime rendering, `close.py`, `status.py`, or `verify`, so **no running strategy is at risk**.

But the central promise — *"the smallest budget where every wallet funds and its smallest slot clears the $12 bumped notional"* — is not what the code computes for a meaningful slice of the fleet, and it silently degrades rather than failing loudly. Plus a rollout blocker.

---

## Blocking

### 1. `_round_step` rounds *down* above $250 — the cap silently understates

```
_round_step(271.67) -> 250        _round_step(1000) -> 250
```

`ox` is already hitting it: **raw $271.67 → shipped $250**. Its `ballast` sleeve cannot fund at the advertised minimum — the exact failure this PR exists to prevent.

Worse, it neuters the documented escape hatch: `min_budget_floor: 500` is passed through `_round_step` and comes out **$250**. An author explicitly declaring a higher floor silently loses half of it.

Fix: past the last rung, `ceil` to the next $25 instead of clamping; and let `min_budget_floor` bypass rounding entirely.

### 2. `_margin_pct` takes the declared *baseline*, not the *smallest* slot

The module docstring commits to worst-case (*"its smallest slot … and its LOWEST leverage"*). Leverage **is** worst-cased; margin is not — `strategy.margin_pct` wins before `marginPctTiers` is ever consulted. Ten templates ship a minimum below their own smallest slot:

| id | used | smallest real slot | shipped | should be |
|---|---|---|---|---|
| shadow | 15% | `minMarginPct: 8` | **$20** | $40 |
| ant | 8% | `marginPctTiers.base: 6` | **$100** | $125 |
| gorilla, orangutan | 12% | `tiers.base: 8` | **$40** | $60 |
| chimp, gibbon, raven, starling, viper | 10% | `tiers.base: 7` | **$50** | $60 |
| armadillo | 6% | `tiers.base: 4` | $250 | $250 (capped, see #1) |

Verified against the scanners, not inferred: `strategies/ant/main/scanners/scoring.py:218` emits the per-signal `marginPct` straight from `marginPctTiers`, and `strategies/shadow/main/scanners/scoring.py:114` floors emitted margin at `minMarginPct`.

One caveat for whoever fixes this: `sailfish`'s `leaderMarginPct: 1.5` looks like a hit but is **not** sizing — it's an RS-gap threshold in percentage points. A naive "min over every `*MarginPct*` key" would drop sailfish from $25 to $250. The fix needs an explicit key allowlist, not a regex.

### 3. An unresolvable `marginPct` collapses to the $10 floor — with no warning

`pw = _per_wallet_min(mp, lev) if mp else WALLET_FLOOR`

Three vol-parity sleeves resolve to `None` today — `caribou/long`, `caribou/short`, `hydra/hedge` — because their sizing lives in `baseRiskPct` / `minMarginPct` / `hedgeRiskPct`, none of which the resolver knows. **caribou ships `min_budget: 20`**; at its own `minMarginPct` (3%) × 3x, the real number is ~$270.

`gen_catalog` warns on an unreadable `runtime.yaml`, but is silent here — which is why the regen reports "0 warnings."

Live blast radius today is nil (both packages are already blocked by an unrelated, pre-existing `margin_fraction_offenders` validator failure — see below), but this is the scaling failure mode: the next vol-parity template silently advertises $10/wallet and then no-trades. `strategy_min_budget` should return the unresolved wallets so `gen_catalog` can warn and `deploy` can refuse to assert a number it never computed.

<details><summary>pre-existing, not this PR: caribou + hydra are already undeployable</summary>

```
$ deploy.py validate strategies/caribou
✗ caribou: 6 issue(s) to fix before deploy:
    - instance long: `scanners[1].inputs.minMarginPct` must be a PERCENT in (0,100] — set 3 (not 0.03)
    ...
$ deploy.py validate strategies/hydra
✗ hydra: 2 issue(s) — instance hedge: `scanners[1].inputs.minMarginPct` ... set 2 (not 0.02)
```
Their vol-parity fractions are internal scanner math, not the runtime's `marginPct`, so this looks like a validator false positive worth its own issue.
</details>

### 4. No skill version bumps — installed users never receive this

`senpi-strategy-ops` 2.12.0, `senpi-strategy-discover` 2.12.0, `senpi-strategy-author` 2.11.0, `senpi-trading-runtime` 3.1.0 — **all unchanged**, while `deploy.py`, `discover.py`, `gen_catalog.py`, `catalog.json` and three SKILL.mds all changed. The precedent is explicit (37cc9e3: *"discover: bump 2.10.0 -> 2.11.0 so the regenerated catalog auto-upgrades"*).

On a backend-coupled change this is the worst possible skew: backend moves to $10, installed discover still cards "$100 minimum", installed ops still floors at $100.

---

## Should fix

### 5. `--json` swallows the entire soft-warn tier

`deploy.py`: `log = (lambda m: None) if a.json else (lambda m: print(m))`. The `[E_BUDGET_BELOW_STRATEGY_MIN]` warn is the **only** channel for the new advisory tier, and `report()` emits `{strategy, version, status, instances}` only — `st["below_min"]`, `st["min_budget"]`, and even `note` never enter the JSON payload. Anyone deploying with `--json` funds below the design minimum with zero signal. Add `min_budget` / `below_min` / `note` to the `report()` dict.

### 6. `senpi-strategy-author/SKILL.md` — partial sweep

Lines 115 and 117 still gate on `$102` (`**Balance ≥ $102, or unreadable**` / `**Balance < $102**`) while the prose immediately above now says `$10` / `$11.50`.

### 7. No CI in this repo

`.github/` does not exist. "guarded by a parity checksum test" / "fails CI if the two copies drift" is aspirational — the vendor-parity and golden tests only run when someone remembers to run them. Given the PR deliberately chose vendoring over a shared import, that guard *is* the safety argument for the choice; it needs an actual workflow.

---

## Minor

- **`catalog.json` +13%** (250,463 → 283,761 bytes) entirely from `min_budget_breakdown`, which no consumer reads — `discover.py` takes only `min_budget_binding_wallet`. Drop it, or keep it out of the bundled discover copy.
- **`funding_share: 0`** → `_f(0) or 1.0` silently treats a zero-share sleeve as 100%.
- **`_scanner_inputs`** returns only the *first* `external_scanner`; a multi-scanner runtime is sized off one of them.
- **`_min_leverage`'s `leverage_multipliers` branch is dead** — the field exists in `runtime-yaml.md:139` but in zero packages.

---

## On merge sequencing

The DO-NOT-MERGE gate is correct and necessary. `plan_funding` floors at $10, so merging ahead of the backend means every sub-$100/wallet `create` returns `SERR037` — which `deploy.py` does not translate, so the agent sees a raw backend error with no next-step rule. Worth adding a `SERR037` row to `docs/error-code-taxonomy.md` before this lands.

---

## What's solidly right

- Vendoring + byte-identical checksum test is the correct call across a skill boundary (a shared import isn't possible once these ship separately).
- The golden file is the right instrument — a marginPct retune moves a number and shows up in the diff.
- `discover` genuinely never filters on budget; verified there is no budget term anywhere in ranking.
- `gen_catalog` regen is deterministic and both catalog copies are already in sync at HEAD.
- Deploy-state gains only additive keys, so in-flight deploys and `status`/`close` are unaffected.

senpi-strategy-ops/scripts/min_budget.py
```diff
@@ -0,0 +1,159 @@
...
+            "funding_share": fs,
+            "per_wallet_min": round(pw, 2),
+        })
+    wallet_count = len(insts)
+    raw = max(per_wallet_over_share) if per_wallet_over_share else WALLET_FLOOR
+    floored = max(raw, WALLET_FLOOR * max(wallet_count, 1))
+    computed = _round_step(floored)
+    floor_override = _f(cat.get("min_budget_floor"))
+    if floor_override:                              # one-directional: can only RAISE the value
+        computed = max(computed, _round_step(floor_override))
```
0xsarvesh: **Blocking — the escape hatch is capped by the same rounding.**

`min_budget_floor: 500` -> `_round_step(500)` -> **250**. An author explicitly declaring a higher floor silently loses half of it, and the header comment ("a design that genuinely needs more declares `catalog.min_budget_floor`") promises otherwise.

The override should bypass `_round_step` entirely — it's already a deliberate, authored number.

senpi-trading-runtime/scripts/gen_catalog.py
```diff
@@ -215,7 +225,12 @@ def build(updated, branch):
         instance_count = len(instances)
 
         validate_declared(glossary, c, sid)
-        cadence = derive_cadence_seconds(instances, os.path.dirname(man_path))
+        pkg_dir = os.path.dirname(man_path)
+        cadence = derive_cadence_seconds(instances, pkg_dir)
+        mb = min_budget.strategy_min_budget(m, load_runtimes(instances, pkg_dir))
```
0xsarvesh, line 231: There's a warn for a declared-but-ignored `min_budget` right below, and `load_runtimes` warns on an unreadable runtime.yaml — but nothing warns when `min_budget.py` **could not resolve a wallet's marginPct** and fell back to the bare $10 floor. That's the case that silently ships a wrong number (`caribou` -> $20), and it's why this regen reports "0 warnings". Worth a warn here once `strategy_min_budget` surfaces the unresolved wallets.

senpi-strategy-author/SKILL.md
```diff
@@ -108,15 +108,15 @@ single most common way a strategy silently does nothing.
...
 `account_get_portfolio` → `data.portfolio.total_in_hyperliquid` (fall back to
-`total_withdrawable`). Deploy needs a little **over $100 USDC per wallet (~$102, to cover
+`total_withdrawable`). Deploy needs a little **over $10 USDC per wallet (~$11.50, to cover
 the ~$1.50 creation fee)** — `deploy.py create` reserves the fee first, so a wallet funded
-to exactly $100 still refuses with `[E_FUNDS_BELOW_FLOOR]`.
+to exactly $10 still refuses with `[E_FUNDS_BELOW_FLOOR]`.
 
 - **Balance ≥ $102, or unreadable** → say nothing about funding and move on. Unreadable
   means move on too — no retry loop, no blocking; funding is re-checked at deploy anyway.
 - **Balance < $102** → tell the user NOW, in one line, then keep building:
```
im-vignesh: The prose above was updated to "$10/wallet (~$11.50)" but these two thresholds still read "Balance ≥ $102" / "Balance < $102". As written, every user holding $12–$101 gets a funding warning that contradicts itself — it fires because they're "under $102" while the quoted message tells them they only need ~$11.50. Should be ~$11.50 × wallet-count (or just $12).

im-vignesh (changes requested):

betashop:
Thanks — this was a genuinely deep review; every blocking call was correct. Addressed in `b9b2241` (rebased onto main incl. #520):

**Blocking**
1. **`_round_step` clamp** → ceils to the next $25 past the last rung; **ox $271.67 → $275** (not $250). `min_budget_floor` now bypasses rounding entirely (500 stays 500).
2. **`_margin_pct` baseline → smallest slot** — min over an explicit key allowlist (`marginPct`/`marginPctBase`/`baseAllocationPct`/`minMarginPct` + `marginPctTiers`), rejecting `<=1` values so sailfish's `leaderMarginPct` and vol-parity risk-weights are never read as sizing. Now: shadow **$40**, ant **$125**, gorilla/orangutan **$60**, chimp/gibbon/raven/starling/viper **$60**, armadillo **$325** (its true 4%@1×, once the cap is fixed). sailfish stays **$25**.
3. **Silent $10 fallback** → `strategy_min_budget` returns `unresolved_wallets`; **gen_catalog now warns** (caribou `['long','short']`, hydra `['hedge']`) and deploy surfaces `[E_BUDGET_UNRESOLVED]`. The regen is no longer falsely "0 warnings".
4. **Version bumps** → ops **2.13.0**, discover **2.14.0**, author **2.12.0**, trading-runtime **3.1.2**.

**Should-fix**
5. `--json` now surfaces `min_budget` / `below_min` / `min_budget_note` / `min_budget_unresolved` via `report()`.
6. author SKILL.md 115/117 `$102` → `~$11.50/wallet`.
7. **CI added** — `.github/workflows/tests.yml` runs vendor-parity + golden + ops + discover on every PR/push. The guard is now real.

**Minor** — dropped the unused `min_budget_breakdown` from the baked catalog (**283K → 257K**); `_margin_pct`/`_min_leverage` now size across **all** external_scanners; `funding_share: 0` guarded. Added the **`SERR037`** row to the taxonomy for the merge-sequencing note. Left `_min_leverage`'s `leverage_multipliers` branch in (documented field, harmless).

Golden regenerated — 13 minimums moved up (worst-case), all locked. DO-NOT-MERGE-until-backend still stands.

im-vignesh (approved):

PR #527, opened 2026-08-06
senpi-portfolio: fix ghost (closed) strategies from a stale cross-run state cache

betashop:
**Bug:** `strategies`/`positions` reuse `strategies_full` from the shared state file (`<tempdir>/senpi-portfolio/state.json`) on presence alone, and that state persists across runs with no TTL. So a snapshot cached while a strategy was ACTIVE is served after it closes (a ghost), and misses strategies deployed since. `money` re-fetches fresh → the `9 vs 10` count mismatch + `reconciles:false` that led an agent to recommend closing an already-closed strategy. Both steps already filter `status=["ACTIVE"]` — this is caching, not filtering.

**Fix** (root cause, per @shnoodles' review): `step_money` (step 1) starts each turn from a clean state, and `_load_state` drops a state file older than `STATE_TTL_S` (90s). `money` and `strategies` can't disagree by construction; the standalone `strategies`/`positions` path self-heals past the TTL; no new MCP call. Intra-turn reuse preserved.

**Tests:** ghost dropped via the TTL (RED without it), `step_money` wipes a prior `strategies_full`, within-TTL cache still reused. 41/41.

**Also:** SKILL.md — stop + re-run on `reconciles:false` / a money-vs-strategies count mismatch before any close/rebalance rec. Bump 1.10.0 → 1.11.0.

### DIFF
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -16,7 +16,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.10.0"
+  version: "1.11.0"
   platform: senpi
   exchange: hyperliquid
 ---
@@ -317,11 +317,17 @@ directly.)
 - **Don't infer "wiped out" from a low balance.** Check `total_funded` / `total_withdrawn` — a
   strategy can show a small balance because profits were withdrawn (`netFunded` can be negative). That
   is not a loss.
-- **"Current / my strategies" = ACTIVE only — never CLOSED.** The engine already filters
-  `strategy_list(status=["ACTIVE"])`, so closed strategies are excluded by construction. If you ever
-  reach for `strategy_list` directly, pass `status: ["ACTIVE"]` — a bare call returns CLOSED/PAUSED too
-  and they must not be presented as current. Mention PAUSED strategies only if relevant, clearly
+- **"Current / my strategies" = ACTIVE only — never CLOSED.** The engine filters
+  `strategy_list(status=["ACTIVE"])` AND re-validates any cached strategy set against a fresh live
+  `strategy_list` before reusing it, so a strategy CLOSED since the last read can't linger as a ghost. If
+  you ever reach for `strategy_list` directly, pass `status: ["ACTIVE"]` — a bare call returns CLOSED/PAUSED
+  too and they must not be presented as current. Mention PAUSED strategies only if relevant, clearly
   labeled "paused," never as active.
+- **If the engine's own signals disagree, STOP and re-run — do NOT narrate through it.** A `reconciles:
+  false` in `totals`, or the `money` and `strategies` steps reporting a different strategy count/set, means
+  the numbers didn't tie out. Re-run the step fresh and reconcile BEFORE you say a word — above all before
+  any close / rebalance recommendation. Recommending action on a strategy that turns out to be already
+  closed is exactly the failure this guards against.
 - **The live clearinghouse is the source of truth for whether a strategy holds capital — over the `status`
   field AND over what anyone asserts about the wallet.** The engine reconciles this: a strategy whose live
   wallet holds **$0 account value, no positions, no idle** is flagged **`empty: true`** (`empty_reason`:
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -318,8 +318,8 @@ directly.)
   strategy can show a small balance because profits were withdrawn (`netFunded` can be negative). That
   is not a loss.
 - **"Current / my strategies" = ACTIVE only — never CLOSED.** The engine filters
-  `strategy_list(status=["ACTIVE"])` AND re-validates any cached strategy set against a fresh live
-  `strategy_list` before reusing it, so a strategy CLOSED since the last read can't linger as a ghost. If
+  `strategy_list(status=["ACTIVE"])`, starts each analysis turn from a clean state, and expires the shared
+  cache after a short window — so a strategy CLOSED since a prior run can't linger as a ghost. If
   you ever reach for `strategy_list` directly, pass `status: ["ACTIVE"]` — a bare call returns CLOSED/PAUSED
   too and they must not be presented as current. Mention PAUSED strategies only if relevant, clearly
   labeled "paused," never as active.
--- a/senpi-portfolio/scripts/portfolio.py
+++ b/senpi-portfolio/scripts/portfolio.py
@@ -32,6 +32,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import time
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 MARKET_ENRICH_CAP = 24      # cap the per-asset market pull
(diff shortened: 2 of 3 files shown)


senpi-portfolio/scripts/portfolio.py
```diff
@@ -1295,19 +1295,48 @@ def _money_totals(embedded, strategies, portfolio_totals):
 
 
 # ─────────────────────────────────────────────── self-heal: full strategies[] in state (for step 2 / 3)
+def _live_active_wallet_set(client, meta):
```
shnoodles: This is the first 8 lines of `fetch_strategies` (L557-565) copied — same `strategy_list(status=["ACTIVE"])` call, same `rows = sl if isinstance(sl, list) else _field(...)` unwrap, same `strategyWalletAddress` / `strategy_wallet_address` / `walletAddress` field list. Two parsers for one response shape that now have to be kept in sync; if the wallet field list ever grows, this one silently drops rows and the gate starts firing spurious re-fetches.

It also means an extra `strategy_list` round-trip on *every* `strategies` and `positions` step, not just the stale ones.

senpi-portfolio/scripts/portfolio.py
```diff
@@ -1295,19 +1295,48 @@ def _money_totals(embedded, strategies, portfolio_totals):
...
+    change — or an unreadable check — forces a full re-fetch. "Which strategies are live" drives
+    close/rebalance advice and is too consequential to answer from unvalidated cache."""
     embedded = state.get("embedded_wallet")
     portfolio_totals = state.get("portfolio_totals")
     strategies = state.get("strategies_full")
     if isinstance(embedded, dict) and isinstance(strategies, list) and isinstance(portfolio_totals, dict):
-        return embedded, strategies, portfolio_totals
-    # state absent/partial → recompute the full pull (embedded + fully-hydrated strategies). The market
-    # enrichment is the `positions` step's job — skip it here (want_market only gates step 3's fold).
+        live = _live_active_wallet_set(client, meta)
```
shnoodles: The gate is doing detection at the read site for something that can be prevented at the write site.

`step_money` (L1362) starts from `state = _load_state(state_path)` and never clears `strategies_full`, so last run's snapshot survives step 1 and reaches step 2 — that's the reported `9 vs 10`. `_all_and_persist` (L1476) already avoids this by building state from a fresh `{}` literal instead of merging.

Making `step_money` do the same is one line and removes the disagreement by construction, rather than catching it after the fact and paying an MCP call per step to do so. Detail in the review body.

senpi-portfolio/tests/test_portfolio.py
```diff
@@ -840,6 +840,92 @@ def test_unfunded_empty_strategy_reason():
...
+        ],
+    }
+    with tempfile.TemporaryDirectory() as td:
+        state_path = os.path.join(td, "state.json")
+        with open(state_path, "w") as fh:
+            json.dump(cached_state, fh)
+        res = portfolio.step_strategies(portfolio._FixtureClient(fixture), want_market=False,
+                                        state_path=state_path)
+    s = {x["name"]: x for x in res["strategies"]}["cub"]
+    assert s["account_value"] == 9999.0, "cache was re-fetched despite an unchanged active set (fast path lost)"
```
shnoodles: Worth sitting with this assertion: it pins **serving a stale `account_value` of $9,999 while the live clearinghouse reports $1,000** as correct behavior.

Within a single turn that's the intended fast path, no argument. But nothing here scopes it to a turn — the state file has no TTL, so the same assertion holds across runs hours apart. The docstring on the fix calls out "stale account value + positions" as the harm being fixed; this test locks in that exact harm for the (common) case where the wallet set happens to be unchanged.

A TTL would let you keep the fast path *and* bound this, without the test having to encode a stale value as the expectation.

senpi-portfolio/SKILL.md
```diff
@@ -317,11 +317,17 @@ directly.)
 - **Don't infer "wiped out" from a low balance.** Check `total_funded` / `total_withdrawn` — a
   strategy can show a small balance because profits were withdrawn (`netFunded` can be negative). That
   is not a loss.
-- **"Current / my strategies" = ACTIVE only — never CLOSED.** The engine already filters
-  `strategy_list(status=["ACTIVE"])`, so closed strategies are excluded by construction. If you ever
-  reach for `strategy_list` directly, pass `status: ["ACTIVE"]` — a bare call returns CLOSED/PAUSED too
-  and they must not be presented as current. Mention PAUSED strategies only if relevant, clearly
+- **"Current / my strategies" = ACTIVE only — never CLOSED.** The engine filters
+  `strategy_list(status=["ACTIVE"])` AND re-validates any cached strategy set against a fresh live
```
shnoodles: "so a strategy CLOSED since the last read can't linger as a ghost" is stronger than what the code delivers — the gate compares wallet *sets*, so it catches a close/deploy but not a stale snapshot of an unchanged set (stale account values, stale positions, stale unrealized PnL, all presented as live).

Since this text is what the agent reasons from, an overclaim here is the kind of thing that gets narrated to a user as certainty. Either soften it to what's actually guaranteed, or land a TTL so the sentence becomes true.

shnoodles (commented):
Real bug, correct diagnosis — the PR body names the root cause exactly: *"a shared state file that persists across runs with no TTL."* My one concern is that the fix then lands at the **read site** rather than at that root cause, and that costs more than it needs to while leaving most of the class open.

### What the fix doesn't reach

1. **An unchanged wallet set still serves a stale snapshot.** Run portfolio at 10am and again at 3pm with the same strategies, and the `strategies` step serves 10am account values and 10am positions as live. The new docstring names *"with its stale account value + positions"* as the harm — but that harm survives whenever the set happens to match, which is the common case. Only add/remove is caught.
2. **`meta_warnings` leaks across runs.** `step_strategies` / `step_positions` read it from state and carry it forward (L1393, L1423). When either runs standalone — which the docstrings explicitly promise they support — a previous run's warnings surface in this run's output. Unguarded by the new gate.
3. **Cost.** An extra `strategy_list` round-trip on every `strategies` and `positions` step, including the overwhelming majority where nothing is stale.

### The actual mechanism, and a smaller fix

`step_money` re-fetches its own data fresh, but starts from `state = _load_state(state_path)` (L1362) and **never clears `strategies_full`**. Last run's snapshot survives step 1 and gets served by step 2. That is precisely the reported `9 vs 10`.

Worth noting that `_all_and_persist` already gets this right — it builds state from a fresh `{}` literal rather than merging into the loaded one (L1476). The inconsistency between those two entry points *is* the bug.

Two changes, no new MCP call:

```diff
 # step_money — step 1 of the pipeline is by definition the start of a turn,
 # so derived snapshots from a previous run must not survive it.
-    state = _load_state(state_path)
+    state = {}
```

```diff
+STATE_TTL_S = 90        # covers a narrated money → strategies → positions turn; past that, recompute
+
 def _load_state(path):
     if not path or not os.path.isfile(path):
         return {}
     try:
+        if time.time() - os.path.getmtime(path) > STATE_TTL_S:
+            return {}       # cross-run staleness → recompute (the same fail-open self-heal as a corrupt file)
         with open(path) as fh:
```

(plus `import time`)

The first makes `money` and `strategies` unable to disagree **by construction** rather than detecting the disagreement afterward. The second covers the standalone-`strategies` path where `money` never ran, and bounds *every* stale key including `meta_warnings`. Both are safe: the re-fetch path at L1347–L1351 repopulates `registry_source` / `catalog_source` / `profile_source`, so resetting in `step_money` loses nothing.

That's ~5 lines against ~17, and it sits in `_load_state`, whose docstring already promises exactly this fail-open self-heal behavior.

### Keep

The SKILL.md guidance is worth keeping either way — it's the last line of defense when the engine's signals disagree, and it's independent of which layer the cache fix lands at. `test_stale_cached_strategy_set_is_refetched_not_served_as_ghost` should still pass against the alternative (the state file is written fresh in the test, so give it a backdated mtime, or drive it through `step_money` first).

Not blocking — the PR does fix the reported symptom. Mostly flagging that the cheaper fix is also the more complete one.

betashop:
Good catch on all three — the read-site set-check left value-staleness (unchanged set) and the `meta_warnings` cross-run leak open, and cost a `strategy_list` per step. Adopted your fix in bde7845:

- `step_money` starts from `{}` — money & strategies can't disagree by construction.
- `_load_state` drops state past a 90s TTL — covers the standalone `strategies`/`positions` path and bounds every stale key, no new call.

Updated the ghost test to backdate the mtime (per your note), added one asserting `step_money` can't preserve a prior `strategies_full`, kept the within-TTL reuse test. 41/41. Kept the SKILL.md guidance. Thanks for the sharper fix.

PR #526, opened 2026-08-06
Deploy-verb convergence: deploy.py as a thin wrapper over senpi deploy (Phase 2, release-gated)

danielmbirochi:
> **DO NOT MERGE until `@senpi-ai/runtime` ships the `senpi deploy` release** (Senpi-ai/senpi-trading-runtime#305). Skills `main` is cloned by every agent bootstrap immediately — merging this before the runtime release would teach agents a verb their runtime doesn't have.

Phase 2 of the D1 deploy plan: the runtime now owns the deploy lifecycle end-to-end (`senpi deploy` — reconcile → funds preflight → create → install → one observed fresh tick, as a detached job), so this PR converges the skills layer onto it.

Design doc (decisions, review threads): [Trading Runtime — senpi deploy CLI](https://app.notion.com/p/3b21feaac80681fcb44dfc0b675fa496)

## What changes

- **`deploy.py` becomes a thin wrapper** over `senpi deploy` / `senpi deploy status`: package resolution and attribution stamping stay, the state file is removed (the verb's journal is the record), create/fund/install/verify logic is deleted in favor of the verb.
- **`senpi-strategy-ops/SKILL.md`** teaches the verb's lifecycle: the detached start + status-poll loop, the exit-code contract (0 live · 2 refused · 3 failed · 4 installed-unobserved · 5 interrupted · 6 running · 1 error), the fresh-tick observe predicate (a stale snapshot is an error, not pending), deadline/grace semantics, the paused-match refusal beside the adopt-and-wait rule, and that there is no cancel verb.
- **Budget teaching, two tiers** (rebased onto main's $10-minimum rollout, #519/#521): below the $10/wallet hard floor refuses (`E_FUNDS_BELOW_FLOOR`); at/above the floor but below a wallet's own calculated sizing the deploy **proceeds and warns** (`W_BUDGET_BELOW_STRATEGY_MIN`/`W_BUDGET_UNRESOLVED`). The prefix now carries the verdict — `W_` codes are advisory warnings by construction, `E_` codes mean the operation was refused or failed — so the teaching block leans on the convention instead of naming exceptional codes: relay warnings, never undo, and some reports deliberately carry **no** teardown command (a stranded wallet gets read-only `status.py` triage, and `E_ROLLBACK_INCOMPLETE` owns its own cleanup when present).
[...]

PR #528, opened 2026-08-06
feat(strategy-ops): add `deploy.py upgrade` — apply an edit to a live strategy

betashop:
## Problem

Re-scoring / re-scanning / re-tuning a **live** strategy had no supported path — agents hand-sequenced close+recreate and hit the `./scanners` wrong-directory trap (a funded wallet that never trades), raw `strategy_create_custom_strategy` (naked wallet), and retry-loops on a nonexistent "update" verb. ~80 users hit the `./scanners` break over 4.5 weeks.

## Solution

One resumable, consent-gated verb: **`deploy.py upgrade <id> [--instance <arm>] --budget <usd>`**. It drives the tested `close → create → runtime → verify` path per arm — validates before closing, refuses to flatten without `--yes`, redeploys on a fresh full-budget wallet, siblings untouched. Applies an edit **authored in `senpi-strategy-author`** (ops applies, author edits — routing wired both ways). **Stopgap** for the runtime engine's in-place reload; same command surface when that lands.

## Money-path hardening (all reads that gate money fail **closed**)

- `strategies_for_or_none` — a failed `strategy_list` **refuses** (`blocked`, exit 2) rather than reading `[]` → funding a second wallet.
- Tri-state `_arm_wallet` (reuses `_recover_wallet`'s taxonomy) — refuses on `unreadable`/`unnamed`/`ambiguous`; arity twin fixed at `_recover_wallet` too.
- Circular-refusal loop broken; closing-wait polls the captured `strategyId`s; registered-stall re-renders the edit; scoped `delete_state` spares siblings; resume hints carry `--instance`.
- **Exit codes:** `0` done · `2` refused/failed · **`3` resumable** (`closing`/`closed` — so a `$?`/`&&` caller can't misread the "nothing deployed yet" state as done).

## Review

Three rounds addressed — @shnoodles (arity: the fix was a no-op on multi-arm; my test had stubbed the broken function), @danielmbirochi (8 fail-open money-path findings), @0xsarvesh (exit-3 for in-flight states + budget-continuity doc). **160 tests green**, incl. a real 3-arm fixture through `_scope_pkg` and RED→GREEN regressions.

## Coordination (not code blockers)

- **#526 merge order:** #526 rewrites `deploy.py` and deletes the machinery `upgrade` is built on; if it lands first it must absorb the `upgrade` surface (hand-off checklist posted). If #528 lands first, #526 rebases.
- **Naming:** `upgrade` (destructive close→redeploy) vs the engine's `update` (in-place). Proposal: keep `upgrade` as the stable surface, point it at the engine when it ships (SKILL already promises this).
- **Version reconcile with #512** (both bump strategy-ops SKILL): whoever merges second takes max+1.

### DIFF
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -308,9 +308,16 @@ exists to prevent).
      with `--yes`.
    - Each re-run reports where it is: `closing` (async flatten in flight — re-run) → `closed` → then the
      normal `wallets-ready` → `registered` → `live`. A transient `underfunded` right after close just
-     means the old funds are still returning — wait and re-run.
-3. **Tell the user what changed:** the wallet is NEW (old → new); funds moved main → new; a custom
-   ratchet/stop ladder on the old positions does NOT carry over (they were closed) — re-apply it if wanted.
+     means the old funds are still returning — wait and re-run. **Exit codes:** `0` = done (`live`), `2` =
+     refused / action-required (`needs-consent`, `blocked`, `failed`), **`3` = resumable, re-run**
+     (`closing`/`closed`). Don't treat `closed` as done — the old arm is gone and nothing is deployed yet;
+     keep re-running until `live`.
+   - **Budget continuity:** the fresh wallet is funded with `--budget` from the main wallet. If the old arm
+     held **more** than `--budget`, the surplus does NOT follow it across — it returns to main and stays
+     there. To carry the whole balance over, set `--budget` to the old arm's size (or higher).
+3. **Tell the user what changed:** the wallet is NEW (old → new); funds moved main → new (up to `--budget`
+   — any surplus stays in main); a custom ratchet/stop ladder on the old positions does NOT carry over
+   (they were closed) — re-apply it if wanted.
 
 **NEVER, during an upgrade:**
 - hand-render a `runtime.yaml` or run raw `openclaw senpi runtime create` on a hand-built file — the
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -277,6 +277,11 @@ all). **Redeploy / upgrade a deployed strategy — see the next section, never a
 
 ## Upgrade — apply an edited scan.py / scoring.py / runtime.yaml to a LIVE strategy
 
+> **STOPGAP.** The close→redeploy-on-a-fresh-wallet mechanics below exist only until the runtime engine
+> ships in-place scanner reload (`update`/`refresh`). When it lands, this whole section and the `upgrade`
+> verb collapse to a thin call to it — the command surface stays, the wallet-swap semantics go away. Treat
+> the wallet-swap as scheduled for deletion, not durable doctrine.
+
 The most common change request: the user asks to re-score / re-scan / re-tune a strategy (e.g. "make my
 live strategy more aggressive"). **The edit itself — changing `scoring.py` / `scan.py` / `runtime.yaml`,
 leverage, sizing, DSL — is authored in `senpi-strategy-author`** (the only skill that knows the scanner /
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -9,18 +9,19 @@ description: >-
   "stop/close/uninstall polar" — and for teardown like "close all strategies",
   "return funds to main", "tear everything down" (→ close.py --all). ALWAYS tear
   down via close.py, never a raw strategy_close (that strands the runtime). Also
-  use to APPLY AN EDIT to a strategy that is already live — "re-score / re-tune /
-  change the scanner on my live strategy", "apply my new scoring", "make it buy
-  more aggressively" — via deploy.py upgrade (there is no in-place reload yet, so
-  it closes the arm and redeploys on a fresh wallet; consent-gated, per arm).
-  Authoring the edited files themselves is senpi-strategy-author; APPLYING them to
-  a running strategy is upgrade here. A strategy is a PACKAGE (strategy.yaml + one
+  the skill that APPLIES an edit to an already-live strategy: the edit itself is
+  authored in senpi-strategy-author (the only skill that knows the scanner / yaml /
+  DSL schema — "make my live strategy more aggressive", change leverage/sizing/DSL
+  starts THERE); ops then applies it with deploy.py upgrade — no in-place reload
+  yet, so it closes the arm and redeploys on a FRESH wallet (consent-gated, per
+  arm). ops deploys / closes / applies; it does NOT author or edit strategy files.
+  A strategy is a PACKAGE (strategy.yaml + one
   runtime.yaml per instance + scanners/) the runtime supervises in-process — no
   scanner daemon. deploy.py runs three resumable steps (create→runtime→verify),
   plus upgrade (edit a live strategy); close.py tears down (stop runtime +
   strategy_close → flattens positions, returns funds). The id (spider, polar,
   kodiak) is the package folder. NOT for choosing WHICH strategy
-  (senpi-strategy-discover) or building/editing one (senpi-strategy-author).
+  (senpi-strategy-discover) or authoring / editing the strategy files themselves (senpi-strategy-author).
 license: Apache-2.0
 metadata:
   author: Senpi
@@ -107,7 +108,8 @@ wallet — every deploy gets a FRESH one.** If an existing `<id>` strategy is fo
 (funded but never got a runtime — the reuse trap an agent keeps landing back on) is **closed to recover its
 funds**, then a new wallet is created (prints **`closing-existing`**; re-run `create` once it's closed and
 funds are back); a **live, running** one is left untouched — `create` **refuses** so it can't silently
-flatten a real book (`close.py <id>` first to redeploy). The **`--budget` is a hard target**: create funds
+flatten a real book (to apply an edit to it, use **`deploy.py upgrade`** — see *Upgrade*; `close.py <id>`
+only to tear it down). The **`--budget` is a hard target**: create funds
 exactly what you ask (split by `funding_share`, $10/wallet floor); if your live balance can't cover it,
 create **HALTS with `underfunded`** and the exact shortfall — it will **NEVER silently fund less** (the
 "$1,000 → $10" failure). Fund/free USDC or confirm a smaller amount, then re-run. **Never hand-edit
@@ -165,7 +167,8 @@ scheduled/supervised scanner passes, so it does **not** wait for the first scan 
 > suggest a lower budget below the floor. Do not switch tools. If
 > `create` reports **`closing-existing`**, it's closing a runtime-less `<id>` wallet to recover funds so it
 > can deploy fresh — re-run `create` once it's closed. If it **refuses** "already deployed AND running", a
-> live `<id>` strategy exists — `close.py <id>` first to redeploy on a fresh wallet. If **`runtime`** lost its
+> live `<id>` strategy exists — to apply an edit use **`deploy.py upgrade <id> [--instance <arm>] --budget
+> <usd>`** (closes + redeploys fresh, consent-gated; see *Upgrade*), not a bare `close`+`create`. If **`runtime`** lost its
 > deploy state and can't safely resolve the fresh wallet, it refuses **split by cause** — and in **both**
 > cases you **never hand-register a runtime onto an old wallet** (no manual `runtime create`/`update` with a
 > wallet from a leftover yaml). **`[E_STATE_NO_WALLETS]`**: the backend has **zero** ACTIVE `<id>` wallets, so
@@ -274,18 +277,20 @@ all). **Redeploy / upgrade a deployed strategy — see the next section, never a
 
 ## Upgrade — apply an edited scan.py / scoring.py / runtime.yaml to a LIVE strategy
 
-The most common change request: the user asks to re-score / re-scan / re-tune a strategy, or you recommend
-it after analysis. **There is no in-place scanner reload yet** (a runtime-team `refresh` is coming). Until
-then, upgrading = **close the arm, then recreate it on a FRESH wallet with the edited files** — done **per
-arm**, so a multi-instance strategy's other sleeves keep running. **`deploy.py upgrade` does the whole
-thing** — it is the ONLY correct way to apply an edit to a live strategy. Do not hand-sequence it, and
-never hand-render a runtime.yaml (that's the `./scanners` "NO ENTRY SCANNERS" trap the verb exists to
-prevent).
+The most common change request: the user asks to re-score / re-scan / re-tune a strategy (e.g. "make my
+live strategy more aggressive"). **The edit itself — changing `scoring.py` / `scan.py` / `runtime.yaml`,
+leverage, sizing, DSL — is authored in `senpi-strategy-author`** (the only skill that knows the scanner /
+yaml / DSL schema). This skill's job is to **APPLY** that edited package to the live strategy. **There is no
+in-place scanner reload yet** (a runtime-team `refresh` is coming), so applying = **close the arm, then
+recreate it on a FRESH wallet with the edited files** — done **per arm**, so a multi-instance strategy's
+other sleeves keep running. **`deploy.py upgrade` does the whole apply** — the ONLY correct way. Do not
+hand-sequence it, and never hand-render a runtime.yaml (the `./scanners` "NO ENTRY SCANNERS" trap the verb
+exists to prevent).
 
 **Do this:**
-1. **Edit the on-disk package** in the durable root (`/data/workspace/strategies/<id>/[<instance>/]…`) —
-   `scoring.py` / `scan.py` / `runtime.yaml`. (`upgrade` re-runs the full `validate` preflight itself, so a
-   bad `./scanners` path is caught BEFORE anything closes.)
+1. **Confirm the edited package is on disk** in the durable root (`/data/workspace/strategies/<id>/[<instance>/]…`)
+   — authored via `senpi-strategy-author`, not hand-guessed here. (`upgrade` re-runs the full `validate`
+   preflight itself, so a bad `./scanners` path is caught BEFORE anything closes.)
 2. **Run `upgrade` and re-run it to advance** — one guided step per call (close → fund fresh wallet →
    register runtime → verify), exactly like `create` resumes. `--instance` is required for a multi-arm
    package (upgrades one sleeve, siblings untouched); the budget funds the fresh wallet:
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -8,17 +8,23 @@ description: >-
   "are my positions protected? / do they have a stop-loss (DSL)?",
   "stop/close/uninstall polar" — and for teardown like "close all strategies",
   "return funds to main", "tear everything down" (→ close.py --all). ALWAYS tear
-  down via close.py, never a raw strategy_close (that strands the runtime). A
-  strategy is a PACKAGE (strategy.yaml + one runtime.yaml per instance + scanners/)
-  the runtime supervises in-process — no scanner daemon. deploy.py runs three
-  resumable steps (create→runtime→verify); close.py tears down (stop runtime +
+  down via close.py, never a raw strategy_close (that strands the runtime). Also
+  use to APPLY AN EDIT to a strategy that is already live — "re-score / re-tune /
+  change the scanner on my live strategy", "apply my new scoring", "make it buy
+  more aggressively" — via deploy.py upgrade (there is no in-place reload yet, so
+  it closes the arm and redeploys on a fresh wallet; consent-gated, per arm).
+  Authoring the edited files themselves is senpi-strategy-author; APPLYING them to
+  a running strategy is upgrade here. A strategy is a PACKAGE (strategy.yaml + one
+  runtime.yaml per instance + scanners/) the runtime supervises in-process — no
+  scanner daemon. deploy.py runs three resumable steps (create→runtime→verify),
+  plus upgrade (edit a live strategy); close.py tears down (stop runtime +
   strategy_close → flattens positions, returns funds). The id (spider, polar,
   kodiak) is the package folder. NOT for choosing WHICH strategy
   (senpi-strategy-discover) or building/editing one (senpi-strategy-author).
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "2.13.0"
+  version: "2.14.0"
   platform: senpi
   exchange: hyperliquid
   requires:
@@ -39,6 +45,7 @@ python3 senpi-strategy-ops/scripts/deploy.py validate <id>                 # 0. 
 python3 senpi-strategy-ops/scripts/deploy.py create  <id> --budget <usd>   # 1. create wallets & fund them
 python3 senpi-strategy-ops/scripts/deploy.py runtime <id>                  # 2. register the runtime(s)
 python3 senpi-strategy-ops/scripts/deploy.py verify  <id>                  # 3. GATE — confirm LIVE (runtime+scanner+DSL+budget)
+python3 senpi-strategy-ops/scripts/deploy.py upgrade <id> --instance <arm> --budget <usd>  # apply an EDIT to a live strategy (close→redeploy fresh; resumable; consent-gated)
 python3 senpi-strategy-ops/scripts/status.py                               # what am I running? (+ health)
 python3 senpi-strategy-ops/scripts/close.py          <id>                  # teardown one strategy
 python3 senpi-strategy-ops/scripts/close.py          --all                 # teardown EVERY open strategy
@@ -263,7 +270,49 @@ it returns `closing` and hands polling to you: **re-run `close.py spider`** unti
 Re-runs are idempotent (runtime already gone → skip; already closing/closed → no re-submit). Strategies
 are discovered from `strategy_list` (`skillName==<id>`), so close also cleans up **orphaned** wallets
 that have no runtime. `--instance <name>` scopes an instance (needs its live runtime to map; else omit to close
-all). **Redeploy** = `close` then `create`/`runtime`/`verify`.
+all). **Redeploy / upgrade a deployed strategy — see the next section, never a bare `close`+`create`.**
+
+## Upgrade — apply an edited scan.py / scoring.py / runtime.yaml to a LIVE strategy
+
+The most common change request: the user asks to re-score / re-scan / re-tune a strategy, or you recommend
+it after analysis. **There is no in-place scanner reload yet** (a runtime-team `refresh` is coming). Until
+then, upgrading = **close the arm, then recreate it on a FRESH wallet with the edited files** — done **per
+arm**, so a multi-instance strategy's other sleeves keep running. **`deploy.py upgrade` does the whole
+thing** — it is the ONLY correct way to apply an edit to a live strategy. Do not hand-sequence it, and
+never hand-render a runtime.yaml (that's the `./scanners` "NO ENTRY SCANNERS" trap the verb exists to
+prevent).
+
+**Do this:**
+1. **Edit the on-disk package** in the durable root (`/data/workspace/strategies/<id>/[<instance>/]…`) —
+   `scoring.py` / `scan.py` / `runtime.yaml`. (`upgrade` re-runs the full `validate` preflight itself, so a
+   bad `./scanners` path is caught BEFORE anything closes.)
+2. **Run `upgrade` and re-run it to advance** — one guided step per call (close → fund fresh wallet →
+   register runtime → verify), exactly like `create` resumes. `--instance` is required for a multi-arm
+   package (upgrades one sleeve, siblings untouched); the budget funds the fresh wallet:
+   ```
+   python3 scripts/deploy.py upgrade <id> --instance <arm> --budget <usd>
+   ```
+   - **Flatten consent is built in.** If the arm holds an open position, the first call HALTS with
+     `needs-consent` — it will NOT silently market-exit a live book. Surface the flatten to the user
+     (positions close, funds return to main, redeploy on a NEW wallet), get an explicit yes, then re-run
+     with `--yes`.
+   - Each re-run reports where it is: `closing` (async flatten in flight — re-run) → `closed` → then the
+     normal `wallets-ready` → `registered` → `live`. A transient `underfunded` right after close just
+     means the old funds are still returning — wait and re-run.
+3. **Tell the user what changed:** the wallet is NEW (old → new); funds moved main → new; a custom
+   ratchet/stop ladder on the old positions does NOT carry over (they were closed) — re-apply it if wanted.
+
+**NEVER, during an upgrade:**
+- hand-render a `runtime.yaml` or run raw `openclaw senpi runtime create` on a hand-built file — the
+  `./scanners` trap. `upgrade` renders it INSIDE the instance dir for you.
+- raw `strategy_create_custom_strategy` — that's a naked wallet with no runtime.
+- `create` on top of a live strategy to "just re-fund it" — `create` refuses a running strategy; `upgrade`
+  is the path that closes-then-redeploys.
+- claim "upgraded / live" before `upgrade` reports `live`.
+
+*(Under the hood `upgrade` drives the same tested `close → create → runtime → verify` path on a fresh
+wallet. In-place `refresh` — reload the scanner with no close, no flatten, same wallet — is coming from the
+runtime team; the verb swaps to it when it lands, same command surface.)*
 
 ## Invariants
 
--- a/senpi-strategy-ops/scripts/deploy.py
+++ b/senpi-strategy-ops/scripts/deploy.py
@@ -1111,8 +1111,9 @@ def cmd_upgrade(pkg, a, log):
                                   "note": (f"Closing `{inst.runtime_name}` (flatten positions + return funds; "
                                            f"async). Re-run `{rerun}` to redeploy once it's closed.")})
 
-        # arm_kind == "none" (verified absent) — genuinely not deployed, or a runtime-less trap `create`
-        # itself clears → straight to redeploy.
+        # Nothing live to close, reached two ways: arm_kind == "none" (verified absent — genuinely not
+        # deployed), OR a wallet resolved but its strategy is already closed so `open_mine` is empty (a
+        # runtime-less trap `create`'s own live-guard then backstops). Either way → straight to redeploy.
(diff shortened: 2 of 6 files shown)


senpi-strategy-ops/scripts/deploy.py
```diff
@@ -351,6 +351,22 @@ def strategy_min(pkg):
...
+    entries are left untouched. Raises on an unknown instance name."""
+    names = [i.name for i in pkg.instances]
+    if instance_name not in names:
+        raise SystemExit(f"error: no instance {instance_name!r} in {pkg.id} (have: {', '.join(names)})")
+    kept = [i for i in pkg.instances if i.name == instance_name]
+    # Fund THIS arm with the full --budget: its fractional share of the whole package is irrelevant when
+    # it's the only arm being (re)deployed, and keeping it would scale the budget down (a 0.10-share sleeve
+    # funded at 10% of what the user asked). Treat the scoped arm as the whole.
+    kept[0].funding_share = 1.0
+    pkg.instances = kept
```
shnoodles, line 396: **Blocking — this is the single/multi bug.**

`_wallet_name` (line 337, outside this diff) keys off `len(pkg.instances) > 1`, and this function sets `pkg.instances` to a 1-element list. So every name derived after scoping — the PHASE-A fallback below, and the `strategyName` a scoped `create` assigns the fresh wallet — collapses to the bare `<id>`:

```
>>> deploy._wallet_name(pkg, preipo)   # 3-arm cub
'cub-preipo'
>>> deploy._scope_pkg(pkg, "preipo"); deploy._wallet_name(pkg, pkg.instances[0])
'cub'
```

Preserve the true arity here:

```suggestion
    kept[0].funding_share = 1.0
    pkg.full_instance_count = len(names)  # true arity survives the narrowing — `_wallet_name` needs it
    pkg.instances = kept
```

and read it in `_wallet_name` (one line, can't be a suggestion — it's outside the diff):

```python
    multi = getattr(pkg, "full_instance_count", len(pkg.instances)) > 1
```

The `getattr` default keeps every unscoped caller (`create`, `_recover_wallet`) on today's behavior.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -429,6 +445,14 @@ def cmd_create(pkg, a, log):
...
     runtimes = _cli.list_runtimes()
     existing_open = [s for s in _cli.strategies_for(mcp, skill_name=pkg.id) if _cli.strategy_open(s)]
+    if getattr(a, "instance", None):
+        # single-arm: the live-guard + close-existing must consider ONLY this arm's strategy (the one on
+        # this arm's runtime wallet), never its siblings — else a one-arm redeploy refuses/closes a healthy
+        # sibling. Once this arm is closed, its runtime is gone → no match → create proceeds on a fresh wallet.
+        _arm_rt = _cli.find_runtime(f"{pkg.id}-{a.instance}")
+        _arm_wallet = _cli.runtime_wallet(_arm_rt) if _arm_rt else None
+        existing_open = [s for s in existing_open
+                         if _arm_wallet and _cli.wallet_match(_cli.strategy_wallet(s), _arm_wallet)]
```
shnoodles: This resolves the arm **only** via its runtime, so when `_arm_wallet` is `None` the comprehension yields `[]` — which disables the live-guard *and* the runtime-less-trap close. A `create cub --instance preipo` on an arm whose runtime died but whose wallet is still open and funded now funds a second one; the unscoped `create` would have closed it and recovered the funds first.

The comment says "no match → create proceeds on a fresh wallet", which is right for a *closed* arm but wrong for a runtime-less *open* one. Route it through the same resolver PHASE A uses (see the `_arm_wallet` helper in the next comment) so both callers resolve the arm identically:

```suggestion
        # single-arm: the live-guard + close-existing must consider ONLY this arm's strategy, never its
        # siblings — else a one-arm redeploy refuses/closes a healthy sibling. `_arm_wallet` still resolves
        # a runtime-LESS open arm (by name), so the funded-but-stuck trap is closed here too rather than
        # double-funded; a genuinely absent arm resolves to None → no match → fresh wallet.
        w = _arm_wallet(pkg, pkg.instances[0], mcp)
        existing_open = [s for s in existing_open if _cli.wallet_match(_cli.strategy_wallet(s), w)]
```

(`wallet_match` already returns `False` on a falsy side, so the `_arm_wallet and` guard was redundant.)

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +915,118 @@ def cmd_verify(pkg, a, log):
...
+        # runtime-less-but-open arm looks "not deployed", so upgrade would skip consent + old-wallet
+        # cleanup and DOUBLE-FUND a fresh wallet next to the orphaned open one.
+        arm_rt = _cli.find_runtime(inst.runtime_name)
+        arm_wallet = _cli.runtime_wallet(arm_rt) if arm_rt else None
+        if not arm_wallet:
+            want = _wallet_name(pkg, inst)
+            cands = [s for s in _cli.strategies_for(mcp, skill_name=pkg.id, statuses=["ACTIVE"])
+                     if _cli.strategy_name(s) == want]
+            addrs = {str(_cli.strategy_wallet(s)).lower() for s in cands if _cli.strategy_wallet(s)}
+            arm_wallet = _cli.strategy_wallet(cands[0]) if len(addrs) == 1 else None  # unique match only
```
shnoodles: Lift this into a shared helper — `create --instance` needs the exact same resolution (comment above), and having it in one place is what stops the two from drifting:

```suggestion
        arm_wallet = _arm_wallet(pkg, inst, mcp)
```

with this next to `_scope_pkg`:

```python
def _arm_wallet(pkg, inst, mcp):
    """This arm's wallet address, or None. Prefers its live runtime; if the runtime is GONE (crashed, or
    the funded-but-never-registered trap) the strategy can still be OPEN, so fall back to the unique
    ACTIVE strategy carrying the name `create` gave this arm — the same key `_recover_wallet` matches on,
    and the backend holds it fixed while the strategy is active. Without the fallback a runtime-less-but-
    open arm reads as "not deployed", and a single-arm op would fund a fresh wallet NEXT TO the orphaned
    open one (double funding, un-consented flatten). Never guesses: anything but a unique name match
    stays None, so a sibling's wallet can't bind here."""
    rt = _cli.find_runtime(inst.runtime_name)
    wallet = _cli.runtime_wallet(rt) if rt else None
    if wallet:
        return wallet
    want = _wallet_name(pkg, inst)
    cands = [s for s in _cli.strategies_for(mcp, skill_name=pkg.id, statuses=["ACTIVE"])
             if _cli.strategy_name(s) == want]
    addrs = {str(_cli.strategy_wallet(s)).lower() for s in cands if _cli.strategy_wallet(s)}
    return _cli.strategy_wallet(cands[0]) if len(addrs) == 1 else None
```

Note the fallback only starts working once `_wallet_name` sees the true arity — with the current code it searches for `cub` and finds nothing, which is why this block is a no-op on multi-arm today.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +915,118 @@ def cmd_verify(pkg, a, log):
...
+                                f"retired, and a custom ratchet/stop ladder on the old positions does NOT "
+                                f"carry over — re-apply it after if wanted. Confirm with the user, then "
+                                f"re-run with --yes:\n      {rerun} --yes")}
+                log("\n  " + out["note"]); return out
+            # consent given → close THIS arm via the tested close primitive (same in-process call
+            # cmd_create uses for the runtime-less trap), then hand off to redeploy on a fresh wallet.
+            import close as _close  # noqa: E402 — sibling module, lazy import
+            runtimes = _cli.list_runtimes()
+            for s in open_mine:
+                _close.close_one(pkg.id, s, runtimes, False, log)
```
shnoodles: `close_one` returns `{"status": "failed", "error": ...}` when the runtime is still listed after two delete attempts — i.e. it may re-enter positions — and that's dropped here. State then advances to `closing`, and the `phase == "closing"` branch above only *polls*; it never re-attempts. So a stuck delete becomes a permanent "still flattening, re-run" loop with the real error never surfaced. (`cmd_create` re-attempts on every run, which is why it doesn't have this problem.)

```suggestion
            recs = [_close.close_one(pkg.id, s, runtimes, False, log) for s in open_mine]
            # A failed close (runtime still listed after delete → it may re-enter positions) must SURFACE,
            # not be swallowed. State stays put, so the next run re-enters this branch and re-attempts the
            # close — advancing to `closing` here would poll a strategy nothing is closing, forever.
            bad = [r for r in recs if r.get("status") == "failed"]
            if bad:
                return _emit(a, log, {
                    "strategy": pkg.id, "instance": inst.name, "status": "failed",
                    "note": "close FAILED, nothing redeployed: "
                            + "; ".join(str(r.get("error") or "?") for r in bad)
                            + f"\n      Resolve it, then re-run `{rerun} --yes`."})
```

Uses `_emit` from the next comment — apply them together. `failed` is already in main()'s non-zero exit set.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +915,118 @@ def cmd_verify(pkg, a, log):
...
+                out = {"strategy": pkg.id, "instance": inst.name, "status": "needs-consent",
+                       "wallet": arm_wallet,
+                       "note": (f"UPGRADE will CLOSE the live `{inst.runtime_name}` on wallet {arm_wallet}: "
+                                f"it market-exits any open position (often NONE if the strategy isn't "
+                                f"entering — that's the usual re-tune case), returns funds to your main "
+                                f"wallet, and redeploys the edited arm on a FRESH wallet. The old wallet is "
+                                f"retired, and a custom ratchet/stop ladder on the old positions does NOT "
+                                f"carry over — re-apply it after if wanted. Confirm with the user, then "
+                                f"re-run with --yes:\n      {rerun} --yes")}
+                log("\n  " + out["note"]); return out
```
shnoodles: **`upgrade --json` prints nothing on any PHASE-A status.** `log` is `lambda m: None` under `--json`, and these four sites only `log()` — they never `report()`. Confirmed:

```
status: needs-consent | stdout printed: ''
```

So in the mode an agent is most likely to run, the consent text — the one thing it has to relay to the user before a flatten — reaches an empty stdout and all it gets back is exit code 2.

A small helper next to `cmd_upgrade`, then each `log(...); return out` becomes `return _emit(a, log, {...})`:

```python
def _emit(a, log, out):
    """Emit an upgrade PHASE-A verdict. These aren't `report()` rows, so they need the same --json
    treatment: under --json `log` is a no-op, and without this the consent text — the one thing the
    agent must relay to the user — would leave stdout empty."""
    if a.json:
        print(json.dumps(out, indent=2))
    else:
        log("\n  " + out["note"])
    return out
```

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +915,118 @@ def cmd_verify(pkg, a, log):
...
+    mcp = MCPClient()
+    st = load_state(pkg)
+    up = st.get("_upgrade") or {}
+    rerun = (f"python3 {Path(__file__).name} upgrade {pkg.id} --instance {inst.name}"
+             + (f" --budget {budget_arg(a.budget)}" if a.budget is not None else ""))
+
+    # ---------- PHASE A: close the currently-live arm (once), consent-gated ----------
+    # Skipped on --dry-run: the preview must be side-effect + network free, so it routes straight to the
+    # create dry-run without probing the backend for a live arm.
+    if not a.dry_run and up.get("phase") != "redeploy":
```
shnoodles, line 1040: **Non-blocking, but worth knowing about:** `_upgrade.phase` is only cleared by `delete_state` on a `live` verify. If a redeploy stalls at `registered` (verify returns `not-live`), the marker persists — and a *second* `upgrade` invoked after further edits skips PHASE A entirely, lands in PHASE B at `registered`, and just re-verifies. `cmd_runtime` would idempotent-skip too (same wallet), so the new edits are never applied while the command reports progress.

Narrow window, and no fix in my branch — flagging so it's a known follow-up rather than a surprise.

senpi-strategy-ops/tests/test_deploy_gates.py
```diff
@@ -614,5 +614,202 @@ def test_empty_portfolio_is_zero_not_none(self):
...
+    _PATCH = ("_cli", "MCPClient", "load_state", "save_state", "_wallet_name", "cmd_create")
+
+    def setUp(self):
+        self._saved = {k: getattr(deploy, k) for k in self._PATCH}
+        self._closed = []
+        self._created = []
+        deploy.MCPClient = lambda *a, **k: object()
+        deploy.load_state = lambda pkg: {"id": pkg.id, "instances": {}}
+        deploy.save_state = lambda pkg, st: None
+        deploy._wallet_name = lambda pkg, inst: f"{pkg.id}-{inst.name}"
```
shnoodles: This is why CI was green. `_wallet_name` is the function that's broken, and the fixture stubs it with the answer it's supposed to produce — then builds an already-single-instance pkg, so `_scope_pkg` never runs and the arity trap can't fire.

Drop the stub and scope a real multi-arm fixture instead:

```suggestion
    # `_wallet_name` is deliberately NOT stubbed: the name it derives is the whole mechanism under test.
    _PATCH = ("_cli", "MCPClient", "load_state", "save_state", "cmd_create")

    def setUp(self):
        self._saved = {k: getattr(deploy, k) for k in self._PATCH}
        self._closed = []
        self._created = []
        deploy.MCPClient = lambda *a, **k: object()
        deploy.load_state = lambda pkg: {"id": pkg.id, "instances": {}}
        deploy.save_state = lambda pkg, st: None
```

and, in `_pkg()` just below:

```python
    def _pkg(self):
        """A 3-arm package narrowed by `_scope_pkg` — exactly what main() hands cmd_upgrade for
        `--instance preipo`. Built multi-arm on purpose: scoping a 1-arm fixture would hide the arity
        trap that made the name fallback resolve 'cub' instead of 'cub-preipo'."""
        def _arm(n):
            return types.SimpleNamespace(name=n, funding_share=1 / 3, runtime_name=f"cub-{n}")
        pkg = types.SimpleNamespace(id="cub", dir=Path("/x"), version="1",
                                    instances=[_arm("long"), _arm("short"), _arm("preipo")])
        deploy._scope_pkg(pkg, "preipo")
        return pkg
```

With just that change and no source fix, `test_runtime_gone_but_strategy_open_still_gated` fails `'wallets-ready' != 'needs-consent'` — it becomes a real regression guard. `close_one`'s stub also needs to return a dict for the failed-close check above.

My branch adds four more: arity preserved / bare name for a true 1-arm package / failed close surfaces / `--json` emits.

shnoodles (commented):
Read this closely and ran it offline against the real packages. **The design is right** — routing the redeploy through `create → runtime → verify` genuinely does make the `./scanners` and naked-wallet traps unreachable, the `full_validate` preflight lands before any close, and the dry-run is clean (no state written, plans only the named arm, good bad-name error). Suite is green.

**But it only works on single-instance packages.** `_wallet_name` derives the wallet's strategyName from package arity (`len(pkg.instances) > 1`), and `_scope_pkg` narrows that list to 1 — so for every multi-arm package, which is the only case `--instance` exists for, every name derived after scoping comes out as the bare `<id>` for a wallet the backend holds as `<id>-<arm>`. 21 of 104 catalog packages are multi-instance.

Reproduced against `strategies/cub`, driving it exactly the way `main()` does (`_scope_pkg` first, real `_wallet_name`):

```
name the fallback searches for : cub
name the wallet actually has   : cub-preipo

upgrade verdict     : wallets-ready
closed the old arm? : False
funded a NEW wallet?: True
```

No consent prompt, no close, straight to funding a second wallet — the double-funding / un-consented-flatten path a476ab50 was written to close. Same root cause has two more effects:

- **The async-flatten wait never waits.** `close_one` deletes the runtime, so the very next poll falls through to the broken name lookup and reports `closed` while the old strategy is still `ACTIVE` (verified: `reports: closed | expected: closing`). The next call then funds the fresh wallet alongside the still-open one.
- **The redeployed wallet is misnamed** `cub` instead of `cub-preipo`, breaking the convention `_recover_wallet` matches on — a later lost-state `runtime` on the full package hits the "unnamed" branch and refuses `E_STATE_AMBIGUOUS`.

CI missed all three because `UpgradeConsentPhase` stubs out `_wallet_name` — the one function that's wrong — and never calls `_scope_pkg`.

Fixes are in the line comments. All of them together (plus tests) are on **`review/upgrade-arity-fix`** off this head: 154 green, and the two new guards fail on the current code. Say the word and I'll push it.

Not line-anchorable, worth a follow-up either way:
- `references/lifecycle.md:70` still says "redeploy = `close` then `create`/`runtime`/`verify`", which now directly contradicts SKILL.md:273 ("never a bare `close`+`create`").
- The PR body says "+8 unit tests, 145 green" — actual is +14 tests, 150 passing on this head.

senpi-strategy-ops/SKILL.md
```diff
@@ -263,7 +270,49 @@ it returns `closing` and hands polling to you: **re-run `close.py spider`** unti
 Re-runs are idempotent (runtime already gone → skip; already closing/closed → no re-submit). Strategies
 are discovered from `strategy_list` (`skillName==<id>`), so close also cleans up **orphaned** wallets
 that have no runtime. `--instance <name>` scopes an instance (needs its live runtime to map; else omit to close
-all). **Redeploy** = `close` then `create`/`runtime`/`verify`.
+all). **Redeploy / upgrade a deployed strategy — see the next section, never a bare `close`+`create`.**
```
shnoodles, line 296: This rule is right, but the three places an agent actually reads at the moment it's about to hand-sequence still point the old way:

- **`deploy.py`'s own live-guard** — `create` refusing "already deployed AND running" prints *"To redeploy on a fresh wallet, close it first: `close.py <id>`"*. That's the highest-leverage line in the codebase for this PR and it never mentions `upgrade`.
- **SKILL.md:110** — *"a live, running one is left untouched — `create` refuses (`close.py <id>` first to redeploy)"*.
- **SKILL.md:168** — the error-recovery playbook: *"If it refuses 'already deployed AND running' … `close.py <id>` first to redeploy on a fresh wallet."*

And `senpi-strategy-author`'s handoff loop (SKILL.md:290-311) — the likely entry point for *"make my live strategy more aggressive"* — only knows `create → runtime → verify`, so an agent that edits a deployed package there gets routed into the refusal above rather than to `upgrade`.

Left these out of my branch to keep the diff on the correctness bug, but the verb won't get picked up much until the error strings name it.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +945,124 @@ def cmd_verify(pkg, a, log):
...
+        # Resolve the arm's wallet via the shared resolver — its live runtime, or (runtime gone) the
+        # unique ACTIVE strategy carrying the arm's name. Without the fallback a runtime-less-but-open arm
+        # reads as "not deployed", so upgrade would skip consent + old-wallet cleanup and DOUBLE-FUND a
+        # fresh wallet next to the orphaned open one. Same resolver cmd_create's live-guard uses.
+        arm_wallet = _arm_wallet(pkg, inst, mcp)
+        open_mine = ([s for s in _cli.strategies_for(mcp, skill_name=pkg.id)
+                      if _cli.strategy_open(s)
+                      and _cli.wallet_match(_cli.strategy_wallet(s), arm_wallet)]
+                     if arm_wallet else [])
```
danielmbirochi: `strategies_for` rides on `list_strategies`, which swallows transport errors and returns `[]` (`_cli.py:374-377`). One failed/timed-out `strategy_list` here makes a live arm read as "nothing to close": PHASE A skips `needs-consent` and the close, persists `phase=redeploy`, and falls into `cmd_create` — whose scoped guard filters `existing_open` against `wallet_match(s, None)` (always False) — so a fresh wallet is funded next to the still-live arm: un-consented orphan + double-fund, the exact trap this verb exists to close. Same fail-open class as the `listStrategies → []` finding in #526. The fix pattern already exists: `list_runtimes_or_none` (`_cli.py:190`) was added precisely so money paths don't mistake "unreadable" for "gone" — this path needs the `strategy_list` analogue and a hard refusal when the read fails.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +945,124 @@ def cmd_verify(pkg, a, log):
...
+            st["instances"][inst.name] = {"status": "pending"}  # forget the old id → create makes a FRESH wallet
+            st["_upgrade"] = {"phase": "closing"}
+            save_state(pkg, st)
+            return _emit(a, log, {"strategy": pkg.id, "instance": inst.name, "status": "closing",
+                                  "note": (f"Closing `{inst.runtime_name}` (flatten positions + return funds; "
+                                           f"async). Re-run `{rerun}` to redeploy once it's closed.")})
+
+        # nothing live to close (arm not deployed, or a runtime-less trap that `create` itself clears) →
+        # straight to redeploy.
+        up["phase"] = "redeploy"; st["_upgrade"] = up
```
danielmbirochi, line 1117: Once `phase=redeploy` is persisted, PHASE A is skipped forever (`:992`). If this branch was reached on a bad read while the arm is actually live, `cmd_create`'s live-guard then refuses with "use `upgrade`" — which is the command being run. Circular refusal loop; the only escape is hand-editing `.deploy-state.json`, which nothing teaches. Either don't persist the phase on this no-op branch (recompute liveness each call), or have the create live-guard clear `_upgrade` when it fires.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -351,6 +354,46 @@ def strategy_min(pkg):
...
+    cmd_upgrade's PHASE A so the two resolve the arm identically and can't drift."""
+    rt = _cli.find_runtime(inst.runtime_name)
+    wallet = _cli.runtime_wallet(rt) if rt else None
+    if wallet:
+        return wallet
+    want = _wallet_name(pkg, inst)
+    cands = [s for s in _cli.strategies_for(mcp, skill_name=pkg.id, statuses=["ACTIVE"])
+             if _cli.strategy_name(s) == want]
+    addrs = {str(_cli.strategy_wallet(s)).lower() for s in cands if _cli.strategy_wallet(s)}
+    return _cli.strategy_wallet(cands[0]) if len(addrs) == 1 else None
```
danielmbirochi: The docstring says "NEVER guesses", but both callers read `None` as "arm absent → safe to fund fresh". Two ACTIVE strategies carrying the arm's name (residue of a prior double-fund), or a name-rejection fallback wallet (the `SERR055` path at `:548` funds with *no* custom name), both resolve to `None` → the consent gate and the scoped create live-guard disarm together → another wallet funded, no consent. `_recover_wallet` (`:598`) already models the correct behavior — tri-state `none`/`unnamed`/`ambiguous`/`unreadable`, refusing on everything but a clean match. `_arm_wallet` should return the same taxonomy and callers refuse (pointing at read-only `status.py` triage) on anything except a verified clean `none`.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +945,124 @@ def cmd_verify(pkg, a, log):
...
+        arm_wallet = _arm_wallet(pkg, inst, mcp)
+        open_mine = ([s for s in _cli.strategies_for(mcp, skill_name=pkg.id)
+                      if _cli.strategy_open(s)
+                      and _cli.wallet_match(_cli.strategy_wallet(s), arm_wallet)]
+                     if arm_wallet else [])
+
+        if up.get("phase") == "closing":
+            # a close was triggered on a prior call — strategy_close is async, so wait for the flatten
+            # to finish (the arm's open strategy disappears) before minting the fresh wallet.
+            if open_mine:
```
danielmbirochi: This wait is effectively a no-op: after `close_one` the runtime is deleted, so `_arm_wallet` falls to its name fallback, which filters `statuses=["ACTIVE"]` (`:391`). A closing strategy has already left ACTIVE (but is still open, CLOSING isn't in the dead set) → `arm_wallet=None` → `open_mine=[]` → this immediately reports `closed` / "funds returning" while the flatten may still be in flight. Remember the `strategyId` at close time and poll it directly — strictly better than name-matching here.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +945,124 @@ def cmd_verify(pkg, a, log):
...
+    # ---------- PHASE B: redeploy the arm on a fresh wallet — one resumable step per call ----------
+    s = inst_state(st, inst.name)
+    status = s.get("status")
+    if a.dry_run or not s.get("strategyId") or status in (None, "pending", "creating"):
+        return cmd_create(pkg, a, log)
+    if status == "active":
+        return cmd_runtime(pkg, a, log)
+    # registered → a fast single check (max_wait=0); verify deletes state on `live`, clearing _upgrade too.
+    av = argparse.Namespace(**{**vars(a), "max_wait": 0})
+    return cmd_verify(pkg, av, log)
```
danielmbirochi: If the redeploy stalls at `registered`/not-live (e.g. the new scanner is broken — exactly the case where the user edits again and re-runs `upgrade`), every subsequent run lands here: verify-only, `max_wait=0`, forever. The re-edited `runtime.yaml`/scanner is never re-rendered — there's no route back to `cmd_runtime`, and its idempotent skip would no-op anyway — so the status trail keeps judging the OLD deployment while reading as if the edit applied. `registered` under an in-flight `_upgrade` should force a runtime re-create (or at least diff the rendered config and re-create on change).

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -334,7 +334,10 @@ def _wallet_name(pkg, inst):
     """The strategyName `create` assigns this instance's wallet: `<id>-<instance>` for a multi-instance
     package, else the bare `<id>` — sanitized. Recovery re-derives the SAME name to match a wallet back
     to its instance, so a lost-state redeploy binds to the right wallet instead of guessing."""
-    multi = len(pkg.instances) > 1
+    # Arity from the ORIGINAL package, not the possibly-scoped `pkg.instances` — `_scope_pkg` narrows that
+    # list to 1 for a single-arm op, and a multi-arm arm must still be named `<id>-<arm>`, not the bare
+    # `<id>`. The getattr default leaves every unscoped caller (create, _recover_wallet) on today's path.
+    multi = getattr(pkg, "full_instance_count", len(pkg.instances)) > 1
```
danielmbirochi, line 362: Right fix — but `_recover_wallet` (`:614`) still branches on `len(pkg.instances) > 1`. Under `_scope_pkg` that's 1, so a scoped `runtime <id> --instance <arm>` with lost state takes the single-instance branch (`cands = list(active)`) and can bind the arm to a lone ACTIVE *sibling's* wallet — two runtimes trading one wallet, the reuse trap the refusal split exists to prevent. Same `getattr(pkg, "full_instance_count", …)` treatment needed at `:614`.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +945,124 @@ def cmd_verify(pkg, a, log):
...
+        save_state(pkg, st)
+
+    # ---------- PHASE B: redeploy the arm on a fresh wallet — one resumable step per call ----------
+    s = inst_state(st, inst.name)
+    status = s.get("status")
+    if a.dry_run or not s.get("strategyId") or status in (None, "pending", "creating"):
+        return cmd_create(pkg, a, log)
+    if status == "active":
+        return cmd_runtime(pkg, a, log)
+    # registered → a fast single check (max_wait=0); verify deletes state on `live`, clearing _upgrade too.
```
danielmbirochi, line 1127: `delete_state` unlinks the whole `.deploy-state.json`, so a scoped verify going `live` also discards the *siblings'* entries (a mid-deploy sibling's `strategyId`/wallet included) — contradicting `_scope_pkg`'s "the siblings' deploy-state entries are left untouched" (`:361`). Under `--instance`, delete only this arm's entry and remove the file when no entries remain.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -891,6 +945,124 @@ def cmd_verify(pkg, a, log):
...
+        # nothing live to close (arm not deployed, or a runtime-less trap that `create` itself clears) →
+        # straight to redeploy.
+        up["phase"] = "redeploy"; st["_upgrade"] = up
+        save_state(pkg, st)
+
+    # ---------- PHASE B: redeploy the arm on a fresh wallet — one resumable step per call ----------
+    s = inst_state(st, inst.name)
+    status = s.get("status")
+    if a.dry_run or not s.get("strategyId") or status in (None, "pending", "creating"):
+        return cmd_create(pkg, a, log)
```
danielmbirochi, line 1124: The resume hints emitted downstream of this hand-off drop both the `--instance` scope and the `upgrade` frame — `"Re-run deploy.py create <id>"` (`:590`), `"Next: deploy.py runtime <id>"` (`:587`), verify's re-run note (`:938`). An agent following them mid-upgrade on a multi-arm package runs *unscoped* `create`, which refuses on live siblings at best — and at `:499` closes a runtime-less sibling **without consent**. Thread the scope + verb into those notes (hints must be computed against actual state).

danielmbirochi (commented):
Reviewed as a stopgap (close→redeploy on a fresh wallet) pending the engine-level `update` verb — not judging it against that ticket's surface. The design and routing are right, the consent text is excellent, and the arity fix + tests are real (suite passes offline). But the two guards the verb exists for — the consent gate and the double-fund guard — currently rest on fail-open reads, the same class as the `listStrategies → []` finding in #526. Inline comments on the specific paths.

Two PR-level items:

1. **Merge order with #526.** #526 rewrites `deploy.py` as a thin wrapper over `senpi deploy` and deletes `.deploy-state.json`, `cmd_create/runtime/verify`, `_wallet_name`, `--instance` — everything `upgrade` is built on. 4 of this PR's 5 files are also in #526's set. If this lands first (likely, since #526 is release-gated), #526 must absorb the `upgrade` surface before it merges, or a fleet taught `deploy.py upgrade` at 2.14.0 hits a verb that no longer exists. Worth recording that dependency on #526 now.
2. **Verb naming.** The permanent ticket's verb is `update` (non-destructive, in-place); this teaches `upgrade` (destructive close+redeploy). If both ever coexist, that's a destructive-vs-non-destructive routing hazard on near-synonyms. Renaming now is nearly free; un-teaching the fleet later is not — either align on one name or state the mapping explicitly in SKILL.md.

Smaller notes, no inline needed: the new refusals (`needs-consent`, close-failed) are the only agent-facing refusals in the file without `[E_*]` codes and the taxonomy doc gets no row; `upgrade` emits `closing` while `create` emits `closing-existing` for a different thing, and PHASE-A JSON has a different shape from `report()` under the same `--json` flag; the `## Upgrade` SKILL block teaches wallet-swap semantics as durable doctrine — worth an explicit STOPGAP marker so it's clearly scheduled for deletion when in-place refresh lands.

Verified as claimed: preflight validate strictly before any close; scoped arm funds at full `--budget` (share=1.0) with single-arm min recomputed; wallet naming survives scoping (real 3-arm fixture, `_wallet_name` unstubbed); failed close surfaces and doesn't advance state; `--json` consent emission fixed; `needs-consent` exits 2; dry-run network-free; no catalog regen needed (no strategy.yaml touched); version bumps correct.

0xsarvesh:
Read this closely because it moves money. **It's good code** — the fail-closed discipline is the right instinct throughout, and it's better tested than most of what's landing this week. One thing I'd want changed before merge, one wording nit, one question.

## What's right

- **Every read that gates money fails closed.** `strategies_for_or_none` → `None` refuses rather than proceeds — *"never fund on a blind read."* Same for the tri-state wallet resolver refusing on `unreadable` / `unnamed` / `ambiguous`.
- **The close-poll uses the captured `strategyIds`, not name matching.** With the reason written down: `close_one` deletes the runtime and a CLOSING strategy has already left ACTIVE, so a name lookup would return nothing and report `closed` while the flatten is still in flight. That's a real bug avoided deliberately.
- **A failed close surfaces instead of advancing state**, so the next run retries rather than polling a strategy nothing is closing.
- **The consent text is honest** — it names the market-exit, the wallet retirement, *and* that a custom ratchet/stop ladder does not carry over. That last one is the kind of thing users discover afterwards.
- **Refusals exit 2**, so an agent can branch on them.
- 157 tests green, including `test_live_arm_without_yes_needs_consent` and a test that the consent text reaches stdout under `--json`.

## The one I'd fix before merge: `closed` exits 0

Line 1243 lists the non-zero statuses. `closing` and `closed` aren't among them.

`closed` is the most dangerous state this verb can be in: **the old arm is gone, positions flattened, funds returned to main, and nothing is deployed.** An agent — or a `&&` chain, or a human reading `$?` — that treats exit 0 as "step succeeded" stops exactly there, leaving the user with no running strategy and their capital sitting in main.

The JSON carries `status: "closed"` plus a clear re-run instruction, so a caller reading output is fine. But the exit code says *success* at the one moment it most needs to say *not finished*. Given the verb is built entirely around resumability, I'd make in-flight states exit non-zero — or a distinct code (3) meaning "resumable, re-run" — so "keep going" can't be misread as "done".

## Nit

The fallthrough comment reads `arm_kind == "none" (verified absent)`, but that branch is also reached when a wallet *did* resolve and `open_mine` is empty. The second clause — *"or a runtime-less trap `create` itself clears"* — does cover it, and `cmd_create`'s own live-guard backstops it, so this is wording rather than a hole. On a money path I'd still have the comment name both ways in.

## Question

`--budget` funds the fresh wallet from main while the old arm's funds are still returning (handled by the `underfunded` retry). If the old arm held **more** than `--budget`, the remainder stays in main rather than following the strategy across.

Intended? Defensible either way, but worth stating explicitly in the SKILL text — "upgrade" implies continuity, and a user may reasonably expect the same capital to carry over.

betashop:
Thanks @0xsarvesh — all three addressed in `eb641e1`.

1. **`closed` exits 0 → fixed.** Extracted `_exit_code()`: `0` done/info · `2` refused/failed · **`3` resumable-re-run**, and `closing`/`closed` now return 3. You're right that `closed` is the acute one — old arm gone, funds in main, nothing deployed — so a `$?`/`&&` caller that stops on exit 0 strands the user. Scoped it to `closing`/`closed` only (upgrade-exclusive statuses) so the standalone `create`/`runtime` steps keep their exit-0 done-for-this-step semantics and existing `&&` chains don't break. +3 unit tests pin the mapping.
2. **Fallthrough comment → fixed.** Now names both ways in: `arm_kind == "none"` (verified absent) **or** a wallet resolved whose strategy is already closed (`open_mine` empty).
3. **Budget continuity → intended, now documented.** Yes — the fresh wallet is funded with `--budget` from main; if the old arm held more, the surplus stays in main rather than following the strategy across. SKILL now states it explicitly and tells the user to set `--budget` to the old arm's size to carry the whole balance. (The `underfunded` retry handles the reverse — funding before the old funds have finished returning.)

160 green. Ready for a re-look.

PR #531, opened 2026-08-12
skills context reduction: stop mirroring the producer

danielmbirochi:
Stops the skills layer from mirroring state the runtime already renders, and fixes five claims that
were wrong about their own code. Base is `feat/deploy-verb-convergence` (#526), so this PR is the
reduction on top of the convergence, not a re-do of it.

## What changes

**The reduction.** `senpi-strategy-ops/SKILL.md` 626 -> 329 lines (8.3k -> 3.9k words), below even
`main`'s 371. The deleted prose was a second copy of a producer the runtime owns — `buildBudgetEscape`
and `buildSoftBudgetNote` already render the budget warns, computed against terminal state, including
a close command scoped to the sleeves that actually exist. The skill restated all of it, which is not
just expensive but unsafe: the runtime sometimes emits no command at all, while the prose asserted
what the escape "is". Depth that survived moved behind trigger pointers into `references/`
(`refusal-playbook.md`, `editing-a-live-strategy.md`, an expanded `lifecycle.md`).

**Five correctness fixes**, from the cross-check of #526 against the runtime side: the proof gate
being skipped by the two most-copied surfaces; `validate` labelled "no side effects" while it writes
a fetched catalog id to disk; `[INVALID_REQUEST]` taught as a refusal code with no taxonomy row; the
unreadable-universe case exiting 3 rather than 2; and the stale-proof repair being able to outrun the
harness timeout.

**`cli_json` stops discarding a refusal's own JSON on exit code alone.** Some verbs exit with the
job's verdict code, not a transport code, so a refusal still prints a complete report and the gate
threw it away unread.

## Field verification

Run on Box A (dev channel, runtime 3.0.69) as tests 3.2 and 3.3 of the deploy pre-release plan.
Full record: `docs/superpowers/reports/2026-08-12-skills-context-reduction-3.2-3.3.md` in the
personal-agents workspace. Cost $0 — no wallet was created and no funds moved.

- Suites green in a clean checkout: ops 308 passed / 1 skipped, author 6, portfolio 51,
  improve-trades 71, runtime help skill 2.
- `cli_json` proven on live output: the same real `deploy status` call (rc=3, 8809 bytes of valid
  report) returns `None` under the old gate and a parsed report under the new one.
- Exit-code parity holds on every arm reached (0 live / 2 refused / 3 failed); `--max-wait` forwards;
  refusal text is relayed verbatim, including values the wrapper cannot compute (`$403.79` largest
  fundable budget, `short by $999,595.21`).
- Money guard held under `--budget 999999` on both the wrapper and the raw verb — adopted the
  existing wallet, funded nothing.
- Every error code the skills teach exists in the runtime dist (24 taught, 0 missing).
- Agent-in-the-loop: with the playbook prose deleted, the agent hit the wrapper's refusal, applied the
  fix it named, recognised that the edit invalidated the recorded proof, re-validated, and stopped at
  the money boundary to ask. The reduction did not cost steering.

## Review notes

Inline comments below carry six findings from that run. Note that four of them live in files this PR
does not touch (`status.py`, `portfolio.py`, `liveness-verification.md`) — they are anchored to the
diff lines that route agents to those surfaces, and none of them is a regression introduced here.

## Verification trap worth knowing

The five skill versions are byte-identical to #526's (`3.6.10 / 3.0.0 / 1.12.0 / 4.0.0 / 1.8.0`), so
a version check cannot tell the two branches apart on a box. Verify by content marker
(`references/refusal-playbook.md` exists only here) or by `diff -r`.


senpi-strategy-ops/SKILL.md
```diff
@@ -490,137 +254,76 @@ The user just funded a strategy; the last thing they see must explain **how the
...
-"broken"** — it's just not autonomous, and `status.py` says how it's managed: **copy** (follows a
-`traderAddress`, run by Senpi's copy engine) or **manual** (you manage it in the app). The *only*
-no-runtime anomaly is an **autonomous package strategy** (skillName, no trader) missing its runtime →
-flagged **no-runtime** with the fix (likely an interrupted deploy). Also **runtime-stopped**, plus a list
-of **orphan runtimes**. On a host **without openclaw** the registry isn't visible, so package strategies
-report **runtime-unknown** — not a diagnosis; check from the runtime host, and never read it as an
-interrupted deploy. `--fast` skips the per-runtime health call; `--json` for machine output. **Tell the
-user the management mode for off-runtime strategies — do not call them idle.** Don't hand-compose
-`strategy_list` — use `status.py`.
+`python3 scripts/status.py` (`<id>` filters, `--fast` skips the per-runtime health call, `--json` for
```
danielmbirochi, line 260: **Finding 1 — `status.py` fails both 3.3 name clauses (not a regression here; flagging because this PR routes agents to it).**

Verified on Box A against a live `strategy_list` (6 non-CLOSED wallets, runtime 3.0.69):

* It renders `(none)` for wallet `0x5d04bef2…`, whose payload `strategyName` **is** `dev-wiring-test-0728`.
* The `--json` row keys are `package / is_pkg / strategyId / wallet / status / funded / positions / runtime / health` — there is **no `name` and no `name_source`**. `strategyName` is never read.
* `package: "(not on runtime)"` is a display string sitting in a data field.

`status.py` is byte-identical between this branch and `feat/deploy-verb-convergence`, so this is inherited, not introduced. Worth recording against this line because the sentence directly above sends the agent here as "the single source of truth", and `portfolio.py` — the surface that *does* carry `name_source` on every row — passed the same clauses cleanly.

Credit where due: the grouping change means `owl-main` is no longer presented *as* wallet `0x4f9ab6d6…`'s name. The sharpest half of the old naming trap is fixed; the missing-name half is not.

senpi-strategy-ops/scripts/_cli.py
```diff
@@ -63,9 +63,15 @@ def _extract_json(text):
...
-    rc, out, _err = run_cli(args, timeout)
-    if rc != 0 or not out.strip():
+    """Run a CLI command expected to emit JSON on stdout; return the parsed object or None.
+
+    Does NOT gate on `rc`: some verbs (`deploy status`) exit with the JOB's own verdict code, not
+    a transport code, so a refusal still prints a complete report. Discarding it on `rc != 0` threw
+    away the one payload callers most need to read. Mirrors `deploy.py`'s `read_status`, which
+    already ignores rc for the same reason. A transport failure still degrades to None here because
+    it also leaves `out` empty or unparseable — the sentinel survives on those grounds, not on rc."""
+    _rc, out, _err = run_cli(args, timeout)
```
danielmbirochi, line 82: **Finding 2 — the vendored `portfolio.py` registry leg is dead on every real host, and it is three defects rather than one.**

Anchored here because this docstring is the tolerance that `portfolio.py`'s copy never adopted. Proven by A/B on Box A:

| run | `registry_source` | `runtime_registered` | `profile_source` |
|---|---|---|---|
| default | `null` | `null` ×6 | `catalog` |
| `--state /data/.openclaw/senpi-state` | `null` | `null` ×6 | `catalog` |
| `SENPI_STATE_DIR=/data/.openclaw/senpi-state` | **`registry`** | **`true` ×5, `false` ×1** | **`mixed`** |

**(a)** `DEFAULT_STATE_DIR = os.path.expanduser("~/.openclaw/senpi-state")` resolves to `/root/…` while OpenClaw lives under `/data/…`. `/root/.openclaw` exists but holds no `senpi-state`, so nothing looks wrong and `meta.warnings` stays `[]`.

**(b)** The `--state` flag is parsed but **ignored** — `load_runtime_registry` reads `os.environ.get(STATE_DIR_ENV) or DEFAULT_STATE_DIR`. A flag that silently does nothing is worse than an absent one: it reads as an escape hatch and isn't.

**(c)** Telemetry does `json.loads(proc.stdout or "null")` — a *strict* parse of stdout that carries `[plugins]` banners — and does not set `SENPI_LOG_LEVEL=error` in the child env. Measured on the box: the strict parse **fails**, `_cli._extract_json` on the same bytes **succeeds**. So `runtime_health` stays `"unknown"` for every runtime **even after (a) is fixed**.

The sibling `senpi-improve-trades/scripts/review.py` already has the robust `_resolve_state_dir` that walks up from `__file__`, and this file already has the tolerant reader. Both fixes exist in-repo; the portfolio copy has neither.

senpi-strategy-ops/SKILL.md
```diff
@@ -490,137 +254,76 @@ The user just funded a strategy; the last thing they see must explain **how the
...
+machine output). It is the single source of truth — live `strategy_list` ∪ `runtime list`, never the
+ephemeral deploy state — so **don't hand-compose `strategy_list`**. A strategy with **no runtime is not
+"broken"**: it is just not autonomous, and `status.py` labels how each one is managed (copy, manual, …).
```
danielmbirochi: **Finding 3 — the two read surfaces disagree about liveness on the same box, and the agent believed the broken one.**

Live agent run on Box A, read-only prompt ("how are my strategies doing?"). The agent reported:

> *"Runtime liveness: unverified for all strategies. The runtime registry isn't visible from this host, so I can't confirm any scanner is actively ticking… Don't assume they're live-and-ticking just because status says ACTIVE."*

That is exemplary fail-closed phrasing and it is **false**. `status.py`, on the same box in the same minute, reports all five runtimes `✅ healthy`. The agent also reported `hype-momentum`'s mandate as "unknown — no `runtime.yaml` profile found" when that strategy's full runtime.yaml is sitting in the registry the reader failed to open (finding 2).

The failure is silent (`meta.warnings: []`), so nothing prompts the cross-check that this line's own routing would have resolved. Net effect on a real user: the agent under-reports health and disclaims knowledge it demonstrably has on that host.

senpi-strategy-ops/SKILL.md
```diff
@@ -490,137 +254,76 @@ The user just funded a strategy; the last thing they see must explain **how the
...
+**Tell the user the management mode for off-runtime strategies — do not call them idle.** The one real
+anomaly is an autonomous package strategy missing its runtime (**no-runtime**). The full label set:
+[`references/lifecycle.md`](references/lifecycle.md).
 
 **"Are my open positions protected? / do they have a stop-loss?"** → the DSL coverage verdict
 (PROTECTED / UNPROTECTED / STOP-NOT-ON-VENUE). Key trap: an unprotected position shows up as an
-**absence** in `senpi dsl positions`, so you must reconcile open positions against the tracked set — full
-procedure in [`senpi-trading-runtime/references/dsl-protection-check.md`](../senpi-trading-runtime/references/dsl-protection-check.md).
+**absence** in `senpi dsl positions`, so reconcile open positions against the tracked set — procedure:
+[`senpi-trading-runtime/references/dsl-protection-check.md`](../senpi-trading-runtime/references/dsl-protection-check.md).
```
danielmbirochi, line 271: **Finding 4 — `protected` is an attribution stamp, and the agent asserts it to the user as DSL protection.**

With the registry unreadable (finding 2), `has_exit` is permanently `False` and the `not_running` override can never fire, so:

```python
protected = False if not_running else bool(has_exit or skill_name)
```

collapses to `bool(skill_name)` — the presence of a `skillName` attribution field, not any exit that was read.

Confirmed reaching the user in the live agent run: it stated *"Protection: ✅ Template DSL (hard stop + profit ratchet)"* for `owl`, and `⛔ NOT PROTECTED` for the one wallet with no `skillName` — both derived from attribution alone. The code comment beside this logic names the exact trap it was written to close ("the trap that let a user think a funded-but-never-registered strategy was live and protected"); the state-dir bug re-opens it silently.

Note the internal inconsistency this produces in one reply: the agent asserted DSL protection *and*, three paragraphs later, that it could not confirm any scanner was running.

senpi-strategy-ops/scripts/deploy.py
```diff
@@ -1531,7 +1565,14 @@ def common(p):
                 for e in errors:
                     print(f"    - {e}", file=sys.stderr)
             else:
-                print(f"✓ {pkg.id}: deploy-ready ({len(pkg.instances)} instance(s))")
+                proven = ", ".join(f"{n} ✓" for n, s, _ in proofs if s == "proven")
+                print(f"✓ {pkg.id}: structurally deploy-ready ({len(pkg.instances)} instance(s))"
+                      + (f" — proven: {proven}" if proven else ""))
+            # The proof gate is the runtime's, and it refuses PRE-MONEY. Naming it here is the whole
+            # point: an unproven instance makes the NEXT command fail, not this one.
```
danielmbirochi, line 1572: **Finding 5 — the package-identity gate exists only in this wrapper, not in the verb.**

The comment right here draws the correct line for the proof gate ("The proof gate is the runtime's, and it refuses PRE-MONEY"). The identity gate below it is the opposite case, and worth stating explicitly because agents call the verb directly.

`deploy.py create` correctly refused a package whose `runtime.yaml` still carried the source package's identity — pre-money and precise:

```
✗ probefs: 2 issue(s) to fix before deploy — nothing was started:
    - instance main: set runtime `group: probefs` (found 'owl')
    - instance main: set runtime `name: probefs-main` (found 'owl-main')
```

The **raw verb accepted the same package and started a job** (`dpl-37dbd056`). It was stopped — but by `E_VALIDATE_NO_PROOF`, and after a proof was recorded, by `E_FUNDS_SHORT` (`dpl-2ac5350e`). Never by identity. `openclaw senpi validate` also passes such a package and records a proof; it reads the instance dir only.

**Not executed** (it costs money and points a runtime named `owl-main` at a live strategy): with an affordable budget and a valid proof, nothing in the verb's pre-money path examines package identity, so the mismatch would first surface at install — after the wallet is created and funded. `E_RUNTIME_ID_TAKEN` exists in the runtime dist. That is the funded-with-no-runtime shape, reached through a package this wrapper rejects for free.

Given the standing doctrine that guards outside the backend are advisory, this one belongs below the skill. Filing as a runtime-side ticket rather than a change to this PR.

docs/error-code-taxonomy.md
```diff
@@ -27,32 +27,76 @@ failed; `W_` = an advisory warning, and the operation PROCEEDED. Rules:
...
+| `E_VALIDATE_NO_RECIPE` | `senpi-trading-runtime` (`senpi validate`, target resolution — `validate/codes.ts:37`) | The validate target is a directory holding no recipe of its own — typically a package root whose `strategy.yaml` lists instance directories. Validation runs against one runtime, so the root is not a target at all | Not a package defect and not a money path. Point `openclaw senpi validate` at the dir holding that instance's `runtime.yaml`; the refusal lists the instances to pick from. One run per instance — which is also how each instance gets the proof `create` refuses without |
 
-2026-08-05 (D1 Convergence): `deploy.py` is now a thin wrapper over `senpi deploy`, so the
+2026-08-05 (D1 Convergence): the converged `deploy.py` is a thin wrapper over `senpi deploy`, so the
```
danielmbirochi, line 55: **Finding 7 — this file is never installed on a box, and a skill reference links to it as if it were.**

`senpi-strategy-ops/references/liveness-verification.md:113` ends its crash-loop row with:

```
see [`docs/error-code-taxonomy.md`](../../docs/error-code-taxonomy.md)
```

From the installed layout that resolves to `/data/.openclaw/skills/docs/error-code-taxonomy.md`. Skills land as bare copied directories under `/data/.openclaw/skills/<skill>/` — the repo-root `docs/` tree is never installed, so the path does not exist on any agent box (verified on Box A).

Swept every markdown pointer across the five skills in this PR: **36 of 37 resolve on the box; this is the one that does not.** Inherited from `feat/deploy-verb-convergence`, not introduced here.

The same install-layout assumption is why this branch's new `senpi-strategy-ops/tests/test_skill_surface.py::test_every_code_the_ops_skill_names_has_a_taxonomy_row` cannot pass on a box — it reads `TAXONOMY` at the repo root and raises `FileNotFoundError` there, while passing in a clean checkout. Not a code defect, but it does mean "run the suite on the box" is not a usable release gate as written.

Cheapest fix is probably to inline the handful of codes that row needs, or ship the taxonomy inside the skill that cites it.

danielmbirochi (commented):
Field findings from the Box A run of plan items 3.2 and 3.3 (dev channel, runtime 3.0.69, $0 spent). Six comments below.

Four of them (1, 2, 3, 4) live in `status.py` / `portfolio.py`, which this PR does not touch — they are byte-identical to `feat/deploy-verb-convergence`. They are anchored to the diff lines that route agents to those surfaces. **None is a regression introduced by this PR**, and none needs to block it; findings 2 and 5 are the two I would file as tickets before release.

The reduction itself came through clean: exit-code parity on every arm reached, verbatim refusal relay including engine-computed values, and an agent that navigated a refusal, a proof-invalidation rule and a money gate with the playbook prose deleted.

senpi-strategy-ops/scripts/_cli.py
danielmbirochi: This seems to be already broken in production ;s 

I'm fixing it to stop querying directly the file system, and use the runtime interface for getting the data needed.

senpi-strategy-ops/SKILL.md
danielmbirochi: **Fixed in [#532](https://github.com/Senpi-ai/senpi-skills/pull/532)** (`ff8ff636`, `ba6b7a5f`).

Both clauses now pass. `--json` rows carry `name` and `name_source`, sourced from the payload's `strategyName` with the `strategyName → tradingStrategyName → name` fallback and `name_source` recording which field answered — the same honest shape `portfolio.py` already exposed, quoted rather than reinvented. `package` is `None` in the data; `"(not on runtime)"` moved to the render site where the other display fallbacks live.

**The text surface got it too, on a second pass.** The first commit fixed only `--json`, which review caught: the human render still printed the *runtime* name as the unlabelled identity string — exactly the shape your evidence describes, where `owl-main` appears once in the listing attached to the wallet that is not called `owl-main`. The text table now leads with the backend name and labels the runtime as `· runtime <id>`. A name that came from a fallback field is marked and explained in a legend, so a reader can tell "this is its name" from "this is the package id standing in."

Your closing observation — that `portfolio.py` passed the same clauses cleanly and was the surface to copy — is what the fix did, and it went one step further: `_cli.strategy_name_and_source` and `portfolio._strategy_name_and_source` are now held to the same *answers* by a test, with the five genuine divergences (case-insensitive key match, whitespace-only strings, dict/bool/int values) pinned explicitly rather than left uncovered. Without that, the fix would have created a third copy of the name reader — the same back door that produced the original disagreement.

Skill version `3.6.10` → `3.6.15` across this branch.

senpi-strategy-ops/scripts/_cli.py
danielmbirochi: **Fixed in [#532](https://github.com/Senpi-ai/senpi-skills/pull/532)** (`7b82050b`, `596ab408`), with the producer half in [senpi-trading-runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330).

Your A/B is what moved the fix off the filed ticket. Ticket 24 prescribed adopting `review.py`'s `_resolve_state_dir()` — locate the private file more cleverly. That is superseded: `portfolio.py` no longer reads the file at all. It calls `openclaw senpi runtime list --json`, and the runtime renders a descriptor carrying the description, `hasExit` and the DSL ladder. `DEFAULT_STATE_DIR` and its `~` expansion are **deleted** — the guess is removed rather than improved — and `_profile_from_runtime_yaml`, `_dsl_preset_summary`, `_dsl_ladder` and `scripts/_yaml.py` (271 lines) go with them. The skill stops interpreting a schema it does not own.

Against your three defects:

**(a)** resolved by deletion. There is no state dir to resolve; the process that *writes* the registry answers instead, using its own notion of where state lives.

**(b) — one correction to the characterisation.** `--state` was never wired to `load_runtime_registry`, and was never meant to be: it is the **step state file** (`<tempdir>/senpi-portfolio/state.json`, `portfolio.py:1563`), a different concept that happens to share a word. "Parsed but ignored" implies a broken wire; it is a naming collision. It resolved by deletion too — with no runtime state dir left in the file, nothing collides, and the flag's help text now says what it is. Ticket 24 has been corrected on this point.

**(c)** confirmed and fixed. Both CLI reads now go through the ops skill's tolerant `run_cli` + `_extract_json` (`SENPI_LOG_LEVEL=error` in the child env, first-`{` scan over banner-polluted stdout). You were right that this was independently load-bearing: fixing the path alone would have left `runtime_health` at `"unknown"` on every host.

Two things this turned up that your comment did not have visibility into:

- The vendored helpers are now held by a parity test that compares **body and signature** — a `timeout=60` → `timeout=30` drift passed under the body-only comparison the old test used. Your point about parity certifying the wrong thing was the more durable half of this finding.
- Widening the CLI read surfaced a crash path: `_run_cli` caught only `FileNotFoundError` and `TimeoutExpired`, so an `OSError` from `subprocess.run` (a fork failure on a memory-constrained box) took out the whole engine. Widened at the parity origin and identically in the vendored copy, with `SPAWN_FAILED_PREFIX`'s never-ran vs stopped-waiting distinction intact — `deploy.py` branches on it on the money path.

senpi-strategy-ops/SKILL.md
danielmbirochi: **Fixed in [#532](https://github.com/Senpi-ai/senpi-skills/pull/532)** (`7b82050b`, `6fdabb75`).

The two surfaces can no longer disagree, because they now quote the same producer: `status.py` and `portfolio.py` both read `openclaw senpi runtime list --json`. The ops routing line says so explicitly and drops any implication that one independently confirms the other.

The silence is gone too. A failed read yields `null` on every affected field and a `meta.warnings` entry naming the command that failed — never `False`, never empty. `runtime_health` gains `"unverified"` for exactly this case, and the narration in both skills learns a third phrase, **"could not verify on this host"**, so `null` renders as neither protection nor its absence. Without that phrase `null` collapses into one of the two halves of the reply you quoted.

**Your finding got larger under review, in a way worth recording here.** Turning the registry read on activated a dormant fail-open one layer down: `_liveness_from_status` searched only the wrapper of `senpi status --json`, but the verdict lives inside `statuses[]`, and the mapper **fell back to `live`**. Measured on the producer's real shape:

```
{'statuses':[{'health':'unhealthy','lastError':'scan threw'}]}  ->  live
{'statuses':[{'health':'unknown'}]}                             ->  live
{'statuses':[]}   # no running runtime at all                   ->  live
```

End-to-end that gave `runtime_health: live`, `protected: True`, and no warning. It had never run against a production payload before — the id map came from the dead registry read, so the field was pinned `"unknown"` on every host. The branch is what made it load-bearing, and it would have shipped as the mirror image of your finding: the agent over-reporting health instead of under-reporting it.

Both halves are fixed — the verdict is read from `statuses[]`, and only a real health key may promote to `live`; a run state (`running`, `up`, `active`) may only downgrade, per the split `_cli.py` already made after an earlier incident. The suite had been green over it because every status fixture was a bare `{"overallHealth": …}`, a shape the CLI never emits; the fixtures now use the producer's real document, including the `unhealthy` and empty-`statuses` cases.

senpi-strategy-ops/SKILL.md
danielmbirochi: **Fixed in [#532](https://github.com/Senpi-ai/senpi-skills/pull/532)** (`7b82050b`) plus the producer half in [senpi-trading-runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330) (`a115c464`, `95a83eec`).

`protected` no longer derives from attribution. It reads `descriptor.hasExit` — an exit block the **engine** parsed from the deployed `runtime.yaml` it registered — and the expression you quoted is gone. `skill_name` is not part of it in any branch.

The tri-state matters as much as the source: when the read fails, `protected` is `null`, not `False`. "We could not check" and "we checked and there is none" are different claims, and the narration now has a phrase for the first.

Two consequences worth flagging, because both are user-visible:

**`hasExit` is stricter than the `has_exit` it replaces.** `portfolio.py:192` treated *any* `exit:` block as protection. The engine's predicate is its own funding gate (`instanceHasDslExit`: `isTruthyPreset(dsl_preset) || engine === "dsl"`), so a strategy whose `exit:` block carries neither will flip from `protected: true` to `protected: false`. That is a correction, not a regression — `package.ts:331` states that keeping those two predicates identical is why author-green stays deploy-green — but it will change verdicts on real strategies.

**The ladder had a second reader, and it had already drifted.** The descriptor's first cut hand-parsed `dsl_preset` alongside the engine's own parser and diverged on three supported shapes: root `dsl_preset.max_loss_pct` (which the documented resolution order puts first), legacy root `tiers:`, and camelCase `triggerPct`/`lockHwPct`. Each rendered a real-looking ladder full of nulls for a funded, protected position — your finding's failure mode, moved one layer down rather than closed. It now derives from `parseDslPluginConfig(mapRuntimeExitToDslConfig(exit))`, so there is one interpretation again.

The code comment you quoted — the trap about a funded-but-never-registered strategy reading as live and protected — is now enforced by the thing it describes rather than re-derived beside it.

senpi-strategy-ops/scripts/deploy.py
danielmbirochi: **Not fixed, and deliberately so — filed as Ticket 27** in `docs/superpowers/followups/2026-08-11-deploy-followup-tickets.md`. (26 was taken by the `close.py` addressing gap.)

Your reasoning is the reason: a guard that lives only in this wrapper is advisory, because agents call the verb directly. Moving the identity check into the verb's pre-money path is a `senpi-trading-runtime` change, and putting it in a skills PR would leave the real hole open while looking closed.

The ticket records the trace as you measured it: `deploy.py create` refuses pre-money and precisely (`set runtime group: probefs (found 'owl')`), the raw verb accepts the same package and starts a job, and what actually stops it is `E_VALIDATE_NO_PROOF` and then `E_FUNDS_SHORT` — never identity. Also that `openclaw senpi validate` passes such a package and records a proof, because it reads the instance dir only, so the proof gate cannot stand in for the identity gate.

The unexecuted arm is written down rather than left implied: with an affordable budget and a valid proof, nothing in the verb's pre-money path examines package identity, so the mismatch first surfaces at install — after the wallet is created and funded. That is the funded-with-no-runtime shape, reached through a package this wrapper rejects for free. `E_RUNTIME_ID_TAKEN` exists in the dist, which makes it a placement problem rather than a missing-error problem.

Related on the ticket: 19 (a funded wallet stranded with no runtime is the same shape) and 25 (the doctrine that real invariants live below the skill).

docs/error-code-taxonomy.md
danielmbirochi: **Fixed in `12e559bf`, on this branch** — this one lands in #531 itself, since #531 is where the unrunnable test arrived.

The crash-loop row now names its three codes inline (`E_SCANNER_CRASH_LOOP`, `E_SCANNER_TICK_ERROR`, `E_SCANNER_TICK_TIMEOUT`) and ends with `read lastError for the cause`, carrying no cross-tree link. Nothing in the installed skill now points outside its own directory, which takes the pointer sweep to **37/37**.

`test_skill_surface.py::test_every_code_the_ops_skill_names_has_a_taxonomy_row` skips with an explicit reason when the repo-root taxonomy is absent, instead of raising `FileNotFoundError`. A new test patches `TAXONOMY` at a nonexistent path and asserts the target test *skips rather than errors*, so the guard cannot be removed silently. "Run the suite on the box" is a usable release gate again.

Skill version `3.6.10` → `3.6.11`.

One thing your sweep found that this fix does **not** close, worth knowing: `test_every_relative_md_link_exists` scans only the two `SKILL.md` bodies, not `references/*.md` — so it never covered this link, before or after. All relative links across the skills resolve today, but nothing guards against a new dead one inside a reference file. Left as a follow-up rather than widened here.

danielmbirochi:
**On the DSL-positions case, recorded here because it shaped how `protected` is sourced in [#532](https://github.com/Senpi-ai/senpi-skills/pull/532).**

The runtime already exposes live DSL state, and it is already the production verdict procedure — not something to add. `openclaw senpi dsl positions --json [-a <addr>]` (gateway `senpi.listDslPositions`) returns `phase`, `currentTierIndex`, `floorPrice`, `currentBreachCount`/`breachesRequired`, `currentROE` and `elapsedMinutes` per tracked position, and `senpi dsl inspect <asset> --json` gives the full state for one. The verdict procedure built on them is `senpi-trading-runtime/references/dsl-protection-check.md`, routed from `senpi-strategy-ops/SKILL.md:532` — both shipping on boxes today.

So `portfolio.py` must not re-derive it. That is the same rule as the rest of this PR.

The limit is what decided the design: **`dsl positions` lists positions that *are* tracked, so a flat strategy has none.** `portfolio.py:910` already names that case — *"NO record (sub-Tier-1) → armed: false … EXPECTED, not a gap"* — which means an empty list cannot distinguish "has a DSL exit, currently flat" from "has no DSL at all". `protected` is a **config-posture** question, so it is sourced from the engine-rendered `descriptor.hasExit` (the deployed `runtime.yaml`'s exit block, read by the engine that owns the schema), while live per-position state stays with MCP `ratchet_stop_list` at `portfolio.py:926`.

Worth noting the two DSL sources answer different questions and both are legitimate: the local CLI is the *runtime's* view of what it is tracking, `ratchet_stop_list` is the *backend's*. Nothing in this PR collapses them.

danielmbirochi:
**Where the six findings landed.**

| # | Finding | Resolution |
|---|---|---|
| 1 | `status.py` renders no name / no `name_source` | [#532](https://github.com/Senpi-ai/senpi-skills/pull/532) — `--json` **and** the text surface |
| 2 | the registry leg is dead on every real host | [#532](https://github.com/Senpi-ai/senpi-skills/pull/532) + [runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330) — the guess is deleted, not repaired |
| 3 | the two read surfaces disagree about liveness | [#532](https://github.com/Senpi-ai/senpi-skills/pull/532) — both now quote the same producer |
| 4 | `protected` is an attribution stamp | [#532](https://github.com/Senpi-ai/senpi-skills/pull/532) + [runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330) — sourced from an exit the engine read |
| 5 | the identity gate lives only in the wrapper | **not fixed** — Ticket 27, runtime-side by design |
| 7 | the taxonomy pointer resolves on no box | **this PR**, `12e559bf` |

Each thread above carries the detail and the commit.

Two things reviewers of the stack should know:

**The `hasExit` tightening changes verdicts on real strategies.** `portfolio.py` treated any `exit:` block as protection; the descriptor uses the engine's own funding-gate predicate, so a strategy whose `exit:` carries neither `dsl_preset` nor `engine: dsl` flips from `protected: true` to `protected: false`. A correction, but a user-visible one.

**Ship order.** `runtime list --json` exists only on [runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330); `origin/main`, `release` (3.0.32) and `dev` (3.0.71) all still carry the option-less command. Until #330 ships, every runtime field in #532 is `null` on every box — which is not a regression, since the read it replaces was already dead on every real host. It converts a silent reassuring falsehood into a loud null. Box A verification is scheduled once #330 reaches the dev channel; nothing in this stack has been verified on a box yet.

senpi-strategy-ops/scripts/_cli.py
```diff
@@ -208,9 +223,11 @@ def _parse_runtime_list(out):
 
 def list_runtimes():
     """All runtimes (running AND stopped) by parsing `runtime list` text. NOTE on runtime v3: `runtime
-    list` has no --json (human text only), and `status --json` is *flaky* — it transiently returns an
-    empty `statuses[]` even while runtimes are running — so it is NOT a reliable inventory. The text
-    table (id / wallet / source / status) is authoritative; use `status -r <id>` only for health.
+    list --json` exists only on newer builds (see `senpi-trading-runtime/references/runtime-cli.md`) and
```
0xsarvesh, line 226: **MUST FIX** — `runtime list` has no `--json` on any current build, and this reverses a correct fact from this PR's own base.

The new docstring says:

> `runtime list --json` exists only on newer builds (see `senpi-trading-runtime/references/runtime-cli.md`) and this reader must keep working on the older ones

I checked the runtime source on both branches:

| build | definition |
|---|---|
| runtime `main` | `senpi-commands.ts:138` — `.command("list").description(...).action(async () => {...})` |
| runtime `feat/senpi-deploy` (#305) | `senpi-commands.ts:156` — identical, no options |

Neither has a single `.option()` on `runtime list`. The `--json` you may have seen at `senpi-commands.ts:430` / `:448` belongs to **`action list`** — a different command, with `-r/--runtime`, `-a/--address` and `--json`.

Two things make this worth fixing rather than leaving:

**1. It reverses `821133d` in this PR's own base (#526)** — *"lifecycle: the runtime-CLI table matches the verb — no runtime create, **list takes no `--json`**"*. That commit was right.

**2. The citation is circular.** This docstring cites `senpi-trading-runtime/references/runtime-cli.md` as its authority — and this same PR changes that file's flags column for `runtime list` from `—` to `--json`. So the code cites the doc, the doc was edited in the same commit range, and neither matches the runtime source.

**Why it bites later:** the docstring reads as an invitation to modernise `list_runtimes()` onto `--json` once "older builds" age out. That parse fails, the reader falls to its `[]` path — and this same docstring notes that `[]` conflates *none* with *unreadable*, which is precisely the distinction `list_runtimes_or_none` exists for on teardown's money path.

Fix is two lines: restore the docstring to "`runtime list` has no `--json` (human text only)" and revert the `runtime-cli.md` flags cell to `—`.

0xsarvesh (commented):
Reviewed the reduction and the code changes. One thing to fix, one framing note, and two changes worth calling out as good.

## Fix before merge

The `runtime list --json` claim in `_cli.py` (inline below) is factually wrong on every current build and reverses a correct fact established in this PR's own base. It also propagated into `senpi-trading-runtime/references/runtime-cli.md`, whose flags cell for `runtime list` changed from `—` to `--json` — and the docstring cites that doc as its authority, so the two now corroborate each other rather than the runtime source.

This is the failure mode the root `CLAUDE.md` names directly: *"Copy identifiers from their source, never from memory — a plausible field name, enum, or unit compiles fine, ticks clean, and trades nothing."*

## Framing: this is relocation, not net reduction

The headline is accurate — `senpi-strategy-ops/SKILL.md` goes **626 → 329**, genuinely below `main`'s 367. But the total ops doc surface **grows**:

| | SKILL.md + references/ |
|---|---|
| #526 base | 1,417 lines |
| #531 | 1,494 lines |

That is not a criticism of the design — moving depth out of the always-loaded hot path into on-demand `references/` is exactly what the repo `CLAUDE.md` prescribes, and the always-loaded surface is the one that costs on every invocation. But "context reduction" invites a reader to assume the total shrank, and someone measuring later will find it didn't. Worth one sentence in the PR body so the number isn't a surprise.

I checked the safety invariants survived the cut — `close.py`, `strategy_close`, `skillName`, `marginPct`, `E_FUNDS`, `@senpi-ai/runtime` are all still present. The frequency drops (`W_BUDGET` 10 → 2, `close.py` 24 → 10) read as removed *restatement*, not removed rules, which is the stated intent.

## Two changes worth calling out

**The `run_cli` `OSError` widening is the best change here.** Catching only `FileNotFoundError` let an ENOMEM or EAGAIN fork failure propagate out of a read and kill the caller mid-run — on a strained box, the exact moment you least want to lose the script's output. Widening to every spawn-side `OSError` while keeping timeout separate is right, and the reasoning in the comment is the correct one: never-ran vs stopped-waiting is the distinction the money path makes, and only a timeout is the latter.

**The `cli_json` fix is a real bug fixed.** Discarding a complete, parseable report because the exit code was non-zero — when the verb's contract is that the code carries the *job's* verdict, not the transport's — is a genuine defect, and the field evidence (same live call, rc=3, 8809 bytes of valid report, `None` under the old gate vs a parsed report under the new one) is the right way to demonstrate it.

## Not reviewed

I've deliberately stopped short of a full pass on this one. It sits on `feat/deploy-verb-convergence` (#526), which is gated on runtime #305 — where I've just posted 17 release blockers including four security findings, several of which change the report shape and refusal text this skill relays. A deep review of the relaying layer is worth more once the thing being relayed settles.

Happy to do that pass whenever the stack below is stable.

senpi-strategy-ops/scripts/_cli.py
danielmbirochi: It has, it was implemented in the https://github.com/Senpi-ai/senpi-trading-runtime/pull/305 . Already tested in dev environment.

0xsarvesh: **Withdrawing this — I was wrong, and the docstring is correct.**

I read `runtime list` out of a stale worktree pinned at `6f2cd917`. The branch head is `3609dff2`, and at that head `senpi-commands.ts:169` is:

```ts
.command("list")
.description("List all installed runtimes ...")
.option("--json", "Print raw JSON from the gateway")
.action(async (opts: { json?: boolean }) => {
```

So the picture is exactly what the docstring says:

| build | `runtime list --json` |
|---|---|
| runtime `main` (shipped 3.0.61) | **absent** — bare `.action(async () => {...})` |
| runtime `feat/senpi-deploy` (#305, current head) | **present** |

That also resolves the "reversal" I claimed: `821133d` in #526 was correct *when it was written*, the runtime PR added `--json` afterwards, and this PR is updating for it. Same for the `runtime-cli.md` flags cell — the doc changed because the CLI changed, which is the right practice, not a circular citation. Both edits stand.

Apologies for the noise. The irony of quoting "copy identifiers from their source, never from memory" while reading a week-old checkout is not lost on me.

PR #532, opened 2026-08-13
portfolio reads the producer; status prints the name it read

danielmbirochi:
Resolves findings 1, 2, 3 and 4 from the Box A review on #531. Base is `feat/skills-context-reduction`, so
this is the fix on top of the reduction, not part of it. Finding 7 landed in #531 itself; finding 5 is filed
as Ticket 27 because it belongs below the skill.

**Depends on [senpi-trading-runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330)** for
`senpi runtime list --json`. See "Ship order" at the bottom — it matters.

## The defect

`portfolio.py` is the engine behind "how are my strategies doing?". It located the runtime's registry with
`os.path.expanduser("~/.openclaw/senpi-state")`, which resolves to `/root/…` on a box where OpenClaw lives
under `/data/…`. The read failed on **every real host**, and its guard treated an absent registry as normal, so
`meta.warnings` stayed empty. Everything derived from it degraded to a reassuring value:

- `protected` collapsed to `bool(skill_name)` — the presence of an attribution stamp, rendered to the user as
  *"Protection: ✅ Template DSL (hard stop + profit ratchet)"* for a strategy whose exit block was never read.
- `not_running` — the field written to catch "funded and ACTIVE with nothing behind it" — was structurally
  incapable of ever being `True`.
- `runtime_health` was `"unknown"` for everything, so in one reply the agent asserted DSL protection *and*
  said it could not confirm any scanner was ticking, while `status.py` reported all five runtimes healthy in
  the same minute.

## The fix: ask the producer

`portfolio.py` now calls `openclaw senpi runtime list --json` and consumes the engine-rendered descriptor.
`DEFAULT_STATE_DIR` and its `~` expansion are **deleted** — the guess is removed, not improved. Gone with it:
`_profile_from_runtime_yaml`, `_dsl_preset_summary`, `_dsl_ladder`, and `scripts/_yaml.py` (271 lines) — the
skill stops interpreting a schema it does not own.

Every field that could degrade is now tri-state, and `null` means one concrete thing: **the read did not
happen.**

| field | values |
|---|---|
| `runtime_registered` / `not_running` / `protected` | `true` / `false` / `null` |
| `running_blind` | `true` / `false` / `null` — **new**; surfaces `running — NO ENTRY SCANNERS`, invisible before |
| `runtime_health` | `live` / `degraded` / `not_running` / `unknown` / `unverified` |

`meta.warnings` names the command that failed. Silence was the defect.

`status.py` (finding 1) now prints the backend `strategyName` in both `--json` and the text surface, labels the
runtime name as the runtime's, and stops parking `"(not on runtime)"` in a data field. The original evidence
was two wallets — `owl` (runtime `owl-main`) and a hand-made `owl-main` — where the string `owl-main` appeared
once in the listing, attached to the wallet that was *not* called that.

## Changes with no review thread, called out because they are invisible otherwise

**The vendored-helper parity test now covers what it claims to.** `run_cli` and `_extract_json` had to be
vendored into `portfolio.py` — skills install as bare sibling directories with no cross-skill import path. That
is the same mechanism that let a state-dir resolver drift twenty lines from the one function parity actually
pinned. The test now compares every shared helper, body **and signature** (a `timeout=60` → `timeout=30` drift
passed under body-only comparison), and a second test holds `_cli.strategy_name_and_source` and
`portfolio._strategy_name_and_source` to the same *answers*, with the five real divergences pinned explicitly
rather than left uncovered.

**A crash path introduced mid-branch, and closed.** Shelling out meant `_run_cli` caught only
`FileNotFoundError` and `TimeoutExpired`; an `OSError` from `subprocess.run` — a fork failure on a
memory-constrained box — took out the whole surface. Widened at the parity origin (`_cli.py`) and identically
in the vendored copy, with `SPAWN_FAILED_PREFIX`'s never-ran vs stopped-waiting distinction intact, since
`deploy.py` branches on it on the money path.

**A health mapper that had never run against a real payload.** Turning the registry read on activated a
dormant fail-open: `_liveness_from_status` searched only the wrapper of `senpi status --json`, whose verdict
actually lives inside `statuses[]`, and **fell back to `live`**. So `{'statuses':[{'health':'unhealthy'}]}`,
`{'statuses':[{'health':'unknown'}]}` and an *empty* `statuses` all mapped to `live` — with `protected: True`
and no warning. It also promoted run states (`running`, `up`, `active`) to a health verdict, which `_cli.py`
separates deliberately after an earlier incident. Both are fixed: only a real health key may promote to `live`,
a run state may only downgrade, and a successful read with no recognizable verdict is `unknown`.

The suite had been green over that because **every status fixture was a bare `{"overallHealth": …}` — a shape
the CLI never emits.** They now use the producer's real document, including the `unhealthy` and empty-`statuses`
cases.

## Verification

`python3 -m pytest senpi-portfolio senpi-strategy-ops -q` → **382 passed, 1 skipped**, in both file orderings
(order matters: an earlier revision leaked an empty `PATH` across test modules and broke 5 tests that shell out
to `bash`). The other eight skill suites pass unchanged.

Not yet verified on a box. Box A verification is scheduled once #330 reaches the dev channel.

## Ship order — read before merging

`runtime list --json` exists only on #330. `origin/main`, `origin/release` (3.0.32) and `origin/dev` (3.0.71)
all still carry the option-less command. **Until #330 ships, every runtime field here is `null` on every box.**

That is not a regression: the read it replaces was *already* dead on every real host. This converts a silent
reassuring falsehood into a loud, correctly-null answer — which is the whole point. But do not expect runtime
fields to populate on merging this alone. `senpi-portfolio/SKILL.md` now declares
`requires: senpi-trading-runtime` and describes exactly what a too-old box looks like, so the agent reads it as
"this box's runtime predates the feature" rather than "the strategies are broken."


danielmbirochi:
Producer half: [senpi-trading-runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330). Findings and evidence: [#531](https://github.com/Senpi-ai/senpi-skills/pull/531), which this branches from.

**Merging this alone leaves every runtime field `null` on every box** until #330 ships — see "Ship order" in the description. That is a loud null replacing a silent falsehood, not a regression, but it is not the finished state either.

danielmbirochi:
## Box A verification — PASS, $0

Verified on Box A (dev channel) against `@senpi/runtime@3.0.72`, which carries [runtime#330](https://github.com/Senpi-ai/senpi-trading-runtime/pull/330) (merged to `dev`). Skills overlay at `834ad9c`, confirmed by version (portfolio `1.13.2`, ops `3.6.15`) **and** content (`DEFAULT_STATE_DIR` absent, `_yaml.py` absent, `requires:` present). No wallet created, no funds moved; baseline and final state identical — 6 ACTIVE, **$297.50** net funded.

Full record: `docs/superpowers/reports/2026-08-13-pr531-findings-verification.md`.

**The headline A/B**, with no `SENPI_STATE_DIR` set — the previous pass could only reach these by forcing that variable by hand:

| field | before | now |
|---|---|---|
| `registry_source` | `null` | **`runtime-cli`** |
| `runtime_registered` | `null` ×6 | **`True` ×5, `False` ×1** |
| `runtime_health` | `unknown` ×6 | **`live` ×5, `unknown` ×1** |

**`protected` is earned, not stamped.** The stop-ship risk was the `hasExit` tightening flipping a genuinely protected strategy to unprotected. It did not — and the ladders are *distinct per strategy*, which is what proves they were read rather than derived:

| runtime | hard stop | arms at | tiers |
|---|---|---|---|
| `hype-momentum-main` (live $150) | −20% | +10% | 6 |
| `terrapin-u1` / `u2` / `u3` | −25% / −22% / −18% | +15% / +12% / +10% | 5 |
| `owl-main` | −35% | +5% | 6 |

Under the pre-fix code all five read `protected: true` from `bool(skill_name)` with no ladder at all.

**Finding 3 is closed:** `status.py` and `portfolio.py`, same box same minute, **6/6 agreement** — previously healthy ×5 against unknown ×6.

**The governing rule holds in both directions.** With `openclaw` unreachable and a fresh state file, every one of `runtime_registered` / `not_running` / `running_blind` / `protected` is `null`, `runtime_health` is `unverified`, and the warning names the command: `` RUNTIME STATE UNVERIFIED — `openclaw senpi runtime list --json` failed (command not found: openclaw) `` — "never report a strategy as running, protected, or not-running from this run."

**The agent used it correctly, unprompted.** On the unattributed wallet:

> **How I know:** The runtime registry was successfully read (`runtime_read_ok: true`, `registry_source: "runtime-cli"`) — it answered, and this strategy simply isn't in it. So the "not registered" verdict is definitive, not a failed read.

That distinction — *the read succeeded and the answer is no* vs *the read failed* — is what findings 3 and 4 were about. Compare the previous pass: *"the runtime registry isn't visible from this host, so I can't confirm any scanner is actively ticking"* (false, while `status.py` showed five healthy) alongside *"Protection: ✅ Template DSL"* (unearned).

### Two things reviewers should know

**A new minor finding.** The step state cache (`STATE_TTL_S = 90`) can serve a runtime read up to 90s old — a warm run reported `runtime-cli` / `runtime_read_ok: true` / `protected: true` while `openclaw` was genuinely unreachable. Bounded and documented, and the values were true when read, so it is a snapshot ageing rather than a fabrication. Recording it because it nearly produced a false result in this very run: **any failure-path test must force a fresh state path.**

**One test file still cannot run on a box.** `test_dex_awareness.py` reads the repo-root `strategies/` tree (`FileNotFoundError: /data/.openclaw/skills/strategies/bobcat/main/scanners/scan.py`) — the same install-layout class Task 3 fixed for the taxonomy, different file, out of scope here. Three of the four files from the previous pass's F8 are now clean.

### Promotion

Dev-proven ≠ released. #330 is on `dev` only; it reaches production via `release` → `@senpi-ai/runtime`. **This PR should not merge to a production-bound branch until that ships** — until then its runtime fields are null on any prod box, loudly and by design.

PR #534, opened 2026-08-13
fix(condor,piranha): unreachable entry gates blocked 100% of signals

shnoodles:
## What

condor and piranha were **structurally unable to ever open a position**. Both are fixed here, with regression tests.

Found while investigating a live account (M176334) where **condor ($4,000) and piranha ($1,500) sat `ACTIVE` and fully funded for 15 days — ~7,000 scanner ticks each, zero signals, zero fills.**

Nothing alerted, because nothing was broken in the usual sense: both scanners spawned, ticked on cadence, exited clean, and logged

```
[condor.scan]  WAITING — no apex setup >= MIN_SCORE=12 (scanned 50)
[piranha.scan] WAITING — no forced-flow / liquidation-unwind signature (min score 5); scanned=4 held=[]
```

which is indistinguishable from a selective strategy correctly standing aside.

## condor — the smart-money gate can never be satisfied

`fetch_sm_map` mapped `pct_of_top_traders_gain` onto a field named `consensus_pct`, and `evaluate_trend_continuation` gated it at `>= 70` as if it were a per-market *"% of top traders leaning this way"*.

It isn't. It's a share of **total** top-trader gain, distributed across the whole leaderboard. From a live `leaderboard_get_markets` snapshot (2026-08-13, 271 markets):

| | value |
|---|---:|
| max | **25.63** |
| p90 | 0.92 |
| median | 0.01 |
| **sum across all markets** | **125.2** |
| markets ≥ 70 | **0** |

A market clearing 70 would have to account for 70% of all top-trader gain board-wide. The live gate funnel:

```
valid direction     271
SM consensus >= 70    0   <-- every signal died here
traders >= 50         0
3TF alignment         0
macro gate            0
```

Drop that single gate and the rest of the chain works: **271 → 12 → 4**.

There is no per-market consensus percentage anywhere in the payload — the closest field is the boolean `is_dominant_direction`. The gate was written against a field shape that does not exist. Condor's own catalog copy advertised *"smart money is >=70% leaning the same way"*, so this was intent, not a typo.

**Fix:** directional agreement comes from `is_dominant_direction`. The gain share is kept as a separate *concentration* filter on its own scale, and the `+4/+3/+2` scoring tiers are rescaled onto that scale — they were pinned at 80/75, so even a passing candidate would always have landed in the base `+2` bucket. Tier magnitudes are unchanged, so the maximum score is still 19.

The legacy `minSmConsensusPct` input is **deliberately ignored**: it was set on the wrong scale, and honouring a stale `runtime.yaml` that still carries `70` would silently re-brick the strategy.

## piranha — gate 1 can't resolve by either route

Gate 1 requires OI unwinding ≥3% over 1h (`if oi_pct > -3.0: return None`). There are two ways to get `oi_pct`, and **both were broken**:

1. **Primary source — key-shape mismatch.** `oi_velocity_1h` did `oiv.get("oi_change_pct")` and required a nested dict with a `"1h"` key. Live `market_get_asset_data` returns **flat** keys:
   ```
   "oi_velocity": { "oi_change_pct_5m": …, "oi_change_pct_1h": 1.2772…, "oi_change_pct_4h": … }
   ```
   so it resolved to `None` and the real value was never read. (The tool docs describe it as "`oi_change_pct` (5m/15m/1h/4h)", which reads like a nested object — likely where this came from.) Both shapes are accepted now.

2. **Fallback — cache overwritten before it's read.** `scan.py` wrote the current OI into `oi_cache[cu]`, then passed `_prev_oi(oi_cache, coin)` — reading back the value it had just written. `prev_oi == cur_oi`, so the computed delta was **exactly +0.000000%** every tick. The read now happens before the refresh.

`minScore: 5` was never even consulted. (`scanned=4` is by design — the universe is `[BTC, ETH, SOL, HYPE]`.)

## Verification

Against live data, on this branch:

- **condor** emits `HYPE SHORT score=13` (≥ `MIN_SCORE` 12) — `['4H_TREND_LIGHT -1.0%', '1H_CONFIRMS -0.81%', '15M_SPIKE +6.69', '3TF_ALIGNED_SHORT', 'SM_STRONGLY_TILTED 25.6%', 'DEEP_CONSENSUS (212t)', 'PEAK_SESSION_17UTC']`
- **piranha** produces a `SHORT` thesis `score=6` (≥ `minScore` 5) on a forced-flow signature, and still returns `None` on a quiet tape.

New: `senpi-strategy-ops/tests/test_entry_gate_reachability.py` — **11 tests, 7 of which fail against `origin/main`** (verified in a throwaway worktree). The other 4 are deliberate controls — non-dominant market, quiet tape, nested `oi_velocity` shape, fallback given a genuine previous OI — so "make it trade again" can't be satisfied by deleting the gates.

Full suite: **217 passed** (206 before + 11), plus `validate_strategy`, `validate_universe`, vendor parity and the min_budget golden. `catalog.json` regenerated to both locations via `gen_catalog.py`.

Versions: condor `1.0.0 → 1.1.0` (gate semantics changed), piranha `1.0.1 → 1.0.2` (pure defect fix, no tunables touched). Condor's tagline/thesis updated — they advertised the `>=70%` threshold that never existed.

## ⚠️ Needs an owner decision

Condor's new thresholds — `minSmGainSharePct: 5.0`, tiers `15`/`8` — are **derived from a single live snapshot and are not backtested**. They restore signal flow and are deliberately conservative (5.0 ≈ the median gain share among markets with ≥50 traders; it yields 3–4 candidates per tick, and condor takes only the single highest-scoring one at `maxPositions: 1`). **Someone who owns this strategy should tune them.** Everything is exposed as `runtime.yaml` inputs.

## Follow-ups not in this PR

- **Two more packages show the identical zero-signal shape** on the same account: `albatross` (6,680 ticks / 0 signals — already `catalog.status: blocked`) and `kestrel` (766 ticks / 0 signals). Same class suspected, **not verified**. Worth a fleet-wide sweep for other hard gates written against a misread field.
- `senpi-strategy-ops/tests/test_dex_awareness.py` fixtures use `pct_of_top_traders_gain: 71.0` / `63.0` — values that can't occur in production, and part of why this gate looked plausible. Harmless for what that test asserts (dex handling), but worth rebasing onto the real scale.
- **A `WAITING` tick that never varies is undetectable today.** Consider having the scaffold log the best score seen (`best=9 vs min=12`) and/or the per-gate rejection counts, so "structurally unreachable" is distinguishable from "selective" without reading the recipe.

### DIFF
--- a/senpi-strategy-discover/catalog.json
+++ b/senpi-strategy-discover/catalog.json
@@ -1,6 +1,6 @@
 {
   "_version": "2.1",
-  "_updated": "2026-08-12",
+  "_updated": "2026-08-13",
   "_generated": true,
   "_instructions": "GENERATED from each strategy package's strategy.yaml — do not hand-edit; run senpi-trading-runtime/scripts/gen_catalog.py. Agents read this via senpi-strategy-discover (scripts/discover.py). DECLARED fields (archetype, sub_style, asset_classes, asset_scope, risk_level, tier, belief_plain, direction) are author-set in strategy.yaml `catalog:`; DERIVED fields (assets, leverage_max, funding_split, instance_count, cadence_seconds, time_horizon, max_slots) are computed from instances/params. Labels/glosses come from senpi-strategy-discover/references/glossary.yaml. 'min_budget' is the machine-COMPUTED minimum total budget at which the design functions (min_budget.py) — the smallest budget where every wallet funds and its smallest slot clears the $12 bumped notional; 'wallet_count' + 'min_budget_binding_wallet' explain it. It is NOT authored — deleting an authored min_budget is a no-op; use min_budget_floor to RAISE it. Positions scale with budget above the min. A strategy is a deployable package; install via senpi-strategy-ops.",
   "skills": [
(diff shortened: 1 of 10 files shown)


betashop (approved):
Verified end-to-end. Both are real silent-no-trade defects: condor's `pct_of_top_traders_gain >= 70` gate is unreachable (that field is a board-wide gain *share*, observed max ~25 across 271 markets), and piranha read OI-velocity from a nested key the API returns flat (`oi_change_pct_1h`) plus wrote the OI cache before reading it, so the fallback delta was always 0.00%. Confirmed the code matches the diagnosis — condor moves direction to `is_dominant_direction` and rescales the gain-share tiers onto the real scale; piranha reads the flat key and reads prev_oi before the refresh. The reachability test is exactly right (7 fail on main, 4 deliberate controls stay green, so 'make it trade' can't be faked by deleting gates). CI green, 217 pass, catalog regen + versions (condor 1.1.0 / piranha 1.0.2) correct. Shipping condor's conservative-but-unbacktested thresholds as-is per owner call — strictly better than never-trading, and every threshold is a live-tunable input.

PR #536, opened 2026-08-13
Make caribou, dire and hydra deployable again: scope the marginPct guard, drop caribou's stray telegram token

danielmbirochi:
Three shipped catalog packages cannot be funded on `main`. Neither cause is a bug in the packages' math; both were found during dev-box testing of the deploy verb (scenario S3), where caribou was refused pre-money with 6 issues.

## 1. The marginPct fraction guard refuses correct packages

`margin_fraction_offenders` (added in #482, duplicated in `senpi-strategy-ops/scripts/_pkg.py` and `senpi-strategy-author/scripts/validate_strategy.py`) walked the **whole** runtime document and flagged any key matching `/margin_?pct/i` holding a value in (0,1].

Scanner `inputs` are private tunables in whatever units their scanner defines, and three packages legitimately use the fraction form there:

| Package | Key | Converted where |
|---|---|---|
| caribou (both sleeves) | `inputs.minMarginPct: 0.03` / `maxMarginPct: 0.15` | `scoring.py:262` returns `pct * 100.0`; `scan.py:352` emits the percent |
| hydra (`hedge`) | `inputs.minMarginPct: 0.02` / `maxMarginPct: 0.12` | same vol-parity shape |
| dire (`main`) | `inputs.sizingTiers[].marginPct: 0.20…0.30` | `scan.py` multiplies by 100 at emit |

All three were refused pre-funding — and the refusal is prescriptive ("must be a PERCENT in (0,100] — set 3 (not 0.03)"), so it reads as an instruction to break a working package by sizing it 100× larger.

**The runtime had already resolved this exact question and gone the other way**, in two places, before this guard was written:

- `src/validate/recipe-checks.ts` `findMarginPctFraction` reads **only** `strategy.margin_pct`, commented: *"Deliberately not a walk of the whole document. Scanner `inputs` are private tunables… at least one production strategy carries fractional tier values there on purpose, converting them before emitting. Flagging those would accuse a correct strategy of a bug."*
- `src/deploy/min-budget.ts`: *"marginPct is read from an EXPLICIT allowlist of sizing keys — NOT a `*MarginPct*` match. A pattern would wrongly pick up non-sizing thresholds."*

So the skills side was refusing to fund what the runtime proves fine. This narrows the check to the keys that really are slot sizes — `strategy.margin_pct`, plus `inputs.marginPct` / `inputs.marginPctBase`, which a scanner passes through to its emitted signal — and only at those positions. Every case #482 was built for still fires (its tests are unchanged and still pass); the emitted per-signal value keeps being checked at the live stage, where the real number is visible rather than inferred.

Both copies stay byte-identical, preserving author-green == deploy-green.

## 2. caribou carries a `${TELEGRAM_CHAT_ID}` no box sets

`notifications.telegram_chat_id: "${TELEGRAM_CHAT_ID}"` in both sleeves, a v2 port leftover. `deploy` substitutes only the wallet env (and the decision model where a recipe needs one), so the token survives render and both gates refuse it. caribou is the **only** package of 129 that has one; removing the line makes its notifications block identical to the 51 others that ship one.

This half is not fixable from the lint: the runtime enforces it independently via `LEFTOVER_TOKEN_RE` in `src/deploy/package.ts`, so the token had to go.

## Regression guard

Added `test_every_catalog_package_validates` — every shipped package must pass the gate that funds it. Nothing asserted that, which is why this sat on `main` until someone tried to deploy caribou on a box. It now covers all 104 packages.

## Verification

- `python3 -m pytest senpi-strategy-ops/tests/ senpi-strategy-author/tests/ -q` → **188 passed**
- Full repo suite → **360 passed, 1 failed**; the failure (`test_status_health.py::test_every_health_class_is_rendered`) is **pre-existing on `main`** — verified by running the same suite on a pristine checkout (357 passed, 1 failed, same test). It is `sys.modules` shadowing between `senpi-account-status/scripts/status.py` and `senpi-strategy-ops/scripts/status.py` when both test dirs run in one process; it passes in isolation. Untouched here, out of scope.
- `validate_strategy.py` and `_pkg.validate` both clean across all 104 packages (3 packages / 10 errors before).
- caribou renders with zero leftover placeholders on both sleeves.
- `gen_catalog.py` produces no content drift (timestamp only), so the generated catalogs are deliberately not in this diff.

Skill versions bumped: `senpi-strategy-ops` 2.16.1 → 2.16.2, `senpi-strategy-author` 2.14.0 → 2.14.1.

## Not fixed here, worth its own ticket

`gen_catalog.py` warns that caribou's and hydra's `min_budget` is **understated** — `min-budget`'s `asPercent()` rejects values ≤ 1, so fraction-form sleeves fall back to the $10 floor. caribou advertises **$20** against a real floor near **$275** (at $10/wallet its smallest slot is 3% × $10 × 3x = $0.90 notional against a $12 bumped minimum). Scoping the guard deliberately leaves that untouched — fixing it moves published catalog numbers and is a separate decision.

Also unrelated and still open: `octopus` is undeployable for a genuine bug in its own scanner (its universe builder keeps the synthetic `.upper()` alias `KPEPE` and drops the real `kPEPE`), being fixed separately.


betashop (commented):
Reviewed this end-to-end, and I checked the load-bearing claim rather than taking it on faith. Clean fix.

**The scoping matches the runtime's own check.** I pulled `findMarginPctFraction` in `recipe-checks.ts` to confirm — it reads only `config.strategy.margin_pct` and never walks scanner `inputs`. So the #482 skill-side guard was genuinely *stricter than the engine it's meant to mirror*. Scoping it to `_SIZING_KEYS` under `strategy`/`inputs` puts the two back in agreement, which is the right north star for this check.

**Premise verified, not just the diff.** caribou's `minMarginPct`/`maxMarginPct` are documented fractions and `scoring.py` does `round(pct * 100.0, 4) → PERCENT` before emit; caribou/dire/hydra are all shipped. So the old whole-document `*marginPct*` walk was refusing correct, live packages pre-money — with a prescriptive "set 3, not 0.03" message that would have 100×-oversized them if an author followed it. Real bug, real consequence.

**What's good:**
- Twin guard fixed **identically** in `_pkg.py` and `validate_strategy.py` — no half-fix drift between the two copies.
- Tests assert **both** directions: the private tunables (`minMarginPct`, `sizingTiers[].marginPct`) clear, while the exact `inputs.marginPct` slot-size key still flags. That's the correct invariant, not just "make caribou pass."
- `test_every_catalog_package_validates` is the regression guard that was *missing* — the reason this sat on `main` unseen until an S3 deploy attempt tripped it. Glad it's in now.

**Scope is exactly as your body frames it** — this is the pre-deploy / author-time gate (`validate_strategy.py` exit-1 + `_pkg.py` deploy refusal). Running instances and the runtime engine were never touched, so nothing live was ever at risk; it's *fresh* deploys through the skill path that were blocked. Appreciated keeping the writeup precise on that — it's an easy thing to overstate.

**One residual, non-blocking:** the guard still flags a bare `inputs.marginPct` / `marginPctBase` fraction (your own test locks that in). Correct today — that's the passthrough convention — but a future scanner that keeps a bare `marginPct` fraction and ×100's at emit would trip it. Might be worth one line in the strategy-author SKILL so authors know the convention is now "keep a fraction tunable under a `min`/`max`/tier name, or write the percent," and don't rediscover this the hard way.

Telegram line is fine to drop — it's a dead `${TELEGRAM_CHAT_ID}` placeholder with no value, nothing to rotate.

LGTM.

betashop (approved):
Resolved the merge against main's 3.x author/ops rewrite and verified locally — author 10 passed, ops 341 passed plus the golden every-package-validates over all 104 shipped packages (caribou/dire/hydra included). SKILLs took main's content + a patch bump (3.0.3 / 3.6.23); test_pkg.py is a clean union with main's DSL-exit coverage; the guard rescope and main's new wallet check coexist. Approving.

PR #15, opened 2026-02-27
dsl improvements v1

garg-prashant:
> [!NOTE]
> **High Risk**
> Introduces a new strategy-scoped DSL execution model that can close positions and delete state based on MCP strategy/clearinghouse responses; mistakes here could cause missed protection or unintended closes/cleanup. Also changes tier-floor math in `dsl-v4.py`, affecting live stop levels for existing users.
> 
> **Overview**
> Upgrades the `dsl-dynamic-stop-loss` skill to **v5**, shifting from *per-position* cron/state to a **per-strategy** cron (`DSL_STATE_DIR` + `DSL_STRATEGY_ID`) that queries MCP for strategy status and clearinghouse positions, reconciles/deletes orphan state files, emits **ndjson per position**, and **deletes state on successful close**.
> 
> Adds `scripts/dsl-cleanup.py` and new `references/cleanup.md` to perform strategy-level directory cleanup (delete-only, blocked if any state is still `active`), and updates docs/schemas/output guidance for xyz assets and the new lifecycle.
> 
> Adjusts `scripts/dsl-v4.py` tier-floor computation to lock a fraction of the `entry→highWater` move (with ratchet against stored floors) instead of the prior leverage/ROE-based price formula.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 598b13d9091a9dc0e18ad031599c3bf62e8b0a1b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #13, opened 2026-02-27
New skill: Fortress — Multi-Agent Trading Fortress for Hyperliquid

Collectorscanyon:
## Summary
Adds the `fortress/` skill package for Senpi marketplace submission.

Fortress is a multi-agent trading consensus skill for Hyperliquid with four pillars:
- Smart Money Oracle
- TA Engine
- Volatility Engine
- Risk Arbiter

A trade signal is emitted **only** when all four agents unanimously vote GO (3/3 consensus required).

## Included
- `fortress/SKILL.md` with full frontmatter and operating spec
- `fortress/scripts/*` pillar scripts + consensus aggregator + shared helpers
- `fortress/references/cron-templates.md` with ready-to-use keepalive crons
- `fortress/references/state-schema.md` with state contract
- `fortress/README.md` quick start
- Row added to the main README.md skills table

## Compliance Notes
- MCP-only external calls via `mcporter`
- No direct HTTP/curl exchange calls
- Atomic state writes with `os.replace()`
- HEARTBEAT_OK when idle
- Structured JSON output contract

## Live Results
- 78% win rate on consensus trades
- 62% drawdown reduction
- Captured +60% on HYPE SHORT

## Pricing
- Free Basic: consensus alerts + baseline controls
- Paid Premium: $29/mo (persistent agents + advanced reporting)
- Senpi auto-billing from trading account or trade profits

PR #537, opened 2026-08-13
Mirror / copy trading: new senpi-trade skill + mirror-aware trader-research

betashop:
## Mirror / copy trading — a complete, runtime-honest experience

**What & why.** Mirror/copy trading is the highest-demand thing users ask agents for, and agents had zero training on it — they couldn't reliably find or vet traders, sized copies badly, mis-explained the mechanics, and gave false protection assurances. This ships the whole journey — **find → vet → decide → execute → protect → monitor** — plus the onboarding routing into it.

## What's in it

**New skill — `senpi-trade` `1.0.0`** — the direct-trade skill (manual position **or** mirror a specific trader), one decision at a time. Ships two references: the single-source mirror-mechanics explainer + mirror monitoring.

**Mirror-aware find engine — `senpi-trader-research` `1.0.2 → 1.3.0`** (`scripts/research.py` is the core logic):
- Ranks by **copyability** — can you actually mirror them *now* (price distance from entry) — not by ROI.
- Default **blend** of three complementary views (7d ROI + 30d ROI + 30d realized PnL), ordered by the consistency score — the user never picks a sort or window.
- Per-trader **min budget**, live **book** (bias + top names), and honest **flags** (blowup / idle / dormant / high-turnover / roi-pnl-conflict / no-open-positions).
- Data-integrity hardening: 0-as-missing sentinels coerced to \`None\` + cross-view backfill; the pick is reconciled with the shortlist order (⭐, market-fit explained when it isn't row #1); live **stderr progress** streaming.

**Onboarding routing — `senpi-strategy-discover` `2.16.0 → 2.17.0`** — the copy branch now hands off to the find engine (kills the "proven vs hot" window-pick) and surfaces **both** template flavors: copy-a-trader **and** smart-money-by-signal.

**Cross-skill CTAs + market-read streaming — `senpi-smart-money` `1.1.1 → 1.2.0`, `senpi-market-pulse` `1.1.1 → 1.2.0`** — "mirror the smart money" CTAs; \`pulse.py\` streams progress during the market cross-reference (no silent gap).

## Where to review
- **Logic** lives in `senpi-trader-research/scripts/research.py` (the engine) and `senpi-market-pulse/scripts/pulse.py` (progress). Everything else is `SKILL.md` **narration contracts**.
- **Tests**: 21 engine + 23 discoverability + market-pulse suites green; `research.py` adds ~190 lines of tests covering the blend, sentinel coercion, the flags, and the copyability sort.
- README `Ver` column synced to every touched skill.

## Validated
- Iterated dozens of times on live prod, including a full end-to-end that funded a **real \$400 mirror** (find → close a managed strategy → sim → fund → verify → protect).
- The ranking path was independently reviewed against the live discovery MCP; every finding is fixed here — the Gain-to-Pain view returned wiped/micro-volume/days-old accounts, several numeric fields used \`0\` as a missing sentinel, ROI/PnL sign-conflicts weren't flagged, and the API's own quality filters informed the fix.
- A **second review** then diffed the engine's live output against the platform's own \`execution_estimate_position_opening\` and caught judgment-layer defects the data-only pass couldn't — symmetric (non-directional) slippage, a fabricated trade count, and budget/attribution edges. **All addressed** (17 of 20 findings fixed, 3 deferred with reasons — see the review thread). Two independent reviewers, every finding applied.

## Out of scope — follow-ups, non-blocking
- \`ratchet_stop_add\` silently drops the Phase-1 max-loss floor (false-protection) — MCP-side fix.
- \`strategy_create\` tool description ("do NOT chain / do NOT poll") contradicts this skill's research + sim + verify flow — a product call on which side is authoritative, tracked separately.
- Momentum lazy-enrich tidy (perf: enrich momentum only for the post-sort top-N).


senpi-trader-research/scripts/research.py
```diff
@@ -25,6 +30,25 @@
...
+NEAR_ENTRY_BAND_PCT = 5.0     # a position within this PRICE distance of the trader's entry is a fresh mirror entry
+ENRICH_TOP_DEFAULT = 20       # how many of the blended pool to mirror-enrich — a wide net so genuinely-mirrorable options surface
+MOMENTUM_TOP = 10             # pull the 4h-momentum leaderboard call only for the top N (it's a tiebreak, not part of the copyability rank) — bounds MCP calls on a wide pool
+MIN_NOTIONAL_USD = 12.0       # HL per-position minimum (the $10 floor, auto-bumped to ~$12) — a copy below this is skipped
+MIRROR_DUST_FRAC = 0.01       # positions below this share of notional are dust — excluded from the whole-book budget so a residual tail can't explode it
+# "Who should I copy?" — no single sort is smart enough, so the default FIND blends complementary views and
+# lets a trader seen in more than one (proven AND currently performing) rank higher. The user never picks a sort.
+# GAIN_TO_PAIN_RATIO = risk-adjusted (steady winners, not big-drawdown gamblers); WEEKLY ROI = hot now; MONTHLY ROI = proven return.
+MIRROR_VIEWS = [
```
shnoodles, line 52: **This view is the dirtiest axis in the API, not the steady one.** The comment above calls `GAIN_TO_PAIN_RATIO` "risk-adjusted (steady winners, not big-drawdown gamblers)" — live data says the reverse. Sorting by it collapses the denominator and returns artifacts:

| # by G2P | gainToPainRatio | maxDrawdown | age | total volume |
|---|---|---|---|---|
| 1 `0x62cd…fd51` | **20,615,487** | **−100%** | 197d | $449k |
| 3 `0x13d0…fa8a` | 2,370,428 | **−100%** | 601d | **$1.07** |
| 5 `0xf1df…844a` | 847,972 | **−100%** | **9.4d** | $70k |
| 8 `0xb771…afd6` | 235,570 | **−100%** | **7.6d** | $32k |

Four of eight are wiped accounts; one has $1.07 of lifetime volume; two are barely a week old. And in the *other* two views G2P is `0` (missing) on 7/10 and 6/6 rows — so the field is either absent or nonsense, never usable.

That's a third of the blend's input, and a third of the bounded 20-slot `ENRICH_TOP_DEFAULT` budget, spent on noise.

**Suggested fix:** drop the G2P view and rank the "steady" axis on `tcsValue` instead — it was present and plausible on **all 30 rows** I pulled (0/8/17/22/25/42/50/58/75/83/100). It's the number behind the `consistency` label you already trust; right now you read the label and throw away the score, which is why ordering *within* a bucket falls back to G2P.

senpi-trader-research/scripts/research.py
```diff
@@ -125,35 +151,278 @@ def _candidate(t):
...
             c["trades"] = round(tpd * c["active_days"])
+    if c["trades_per_day"] is None and c["trades"] and c["active_days"]:
+        c["trades_per_day"] = round(c["trades"] / c["active_days"], 2)
+    ts = c.get("last_trade_ts")
+    if ts:
+        ts = ts / 1000.0 if ts > 1e11 else ts   # lastTradeTimestamp seen in both seconds and ms — normalize to s
+        c["last_trade_days_ago"] = round((datetime.datetime.now(datetime.timezone.utc).timestamp() - ts) / 86400.0, 1)
+    else:
+        c["last_trade_days_ago"] = None
     return c
```
shnoodles, line 191: **Sentinels get read as real values in this builder.** `_f()` passes `0` straight through, but the live payload uses `0` for "not computed":

- `gainToPainRatio: 0` on a trader with **$703,585** PnL (`0x9c16…4a6f`)
- `returnOnInvestment: 0` on traders with **$5,062** and **$12,410** realized PnL
- `maxDrawdown: 0` alongside `−100` elsewhere in the same response
- `averageLeverageUsed: 0` **and** `averageMarginUsed: 0` on the same rows — that pair feeds `min_mirror_budget`
- `averageHoldTimeSeconds` is null-or-0 on **10/10** rows — the field is unusable, don't surface it

A `maxDrawdown: 0` reads as a flawless record and skips `blowup_risk` entirely; a `roi_pct: 0` silently sinks a good trader in any ROI-ordered view.

**Suggested fix:** coerce to `None` in `_candidate` before anything downstream sees them — `roi == 0 and pnl not in (None, 0)`, `max_drawdown == 0`, `gain_to_pain == 0`, `avg_leverage == 0`, `avg_margin == 0`. `None` already flows correctly through `_reliability`, `_flags`, and `_mirror_sort_key`; `0` does not.

senpi-trader-research/scripts/research.py
```diff
@@ -125,35 +151,278 @@ def _candidate(t):
...
+def _flags(c, positions=None, net_upnl=None, margin_pct=None):
+    """The analyst's anchor list — surfaced verbatim by the skill. Track-record + book risks together."""
+    flags = []
+    if c.get("reliability") == "thin":
+        flags.append("thin_track_record")     # < 5 trades or < 7 active days — not yet trustworthy
+    if c.get("consistency") == "CHOPPY":
+        flags.append("choppy_consistency")
+    dd = c.get("max_drawdown_pct")
+    if dd is not None and dd <= CATASTROPHIC_DD:
+        flags.append("blowup_risk")            # ≤ -83% max drawdown — near-liquidation even by perps standards
```
shnoodles, line 337: **Missing the one flag that would have caught the worst row in the dataset.**

The `WEEKLY`/`RETURN_ON_INVESTMENT` view — `MIRROR_VIEWS[1]`, "7d hot" — returns this at **rank #1**:

```
0x08c1…1135
  returnOnInvestment:        +295.47
  profitAndLoss:          −1,584,318.20
  unRealizedProfitAndLoss: −2,128,460.31
  winRate:                       100
```

ROI says +295%, PnL says **down $1.58M**, `winRate` says perfect. The skill instructs showing ROI beside PnL and nothing reconciles them, so a user reads "+295% this week" and mirrors a $2.1M unrealized loss.

Current flags don't cover it: drawdown is −34.7% (above `CATASTROPHIC_DD`), the record isn't thin, and `currently_in_drawdown` only fires post-enrichment.

**Suggested fix:** add a sign-coherence flag right here —
```python
roi, pnl = c.get("roi_pct"), c.get("pnl_usd")
if roi is not None and pnl is not None and roi > 0 and pnl < 0:
    flags.append("roi_pnl_conflict")   # headline ROI contradicts actual PnL — do not lead with ROI
```
Worth treating as a hard reject rather than a flag: if the two headline numbers disagree in sign, neither should be shown as a recommendation.

shnoodles (commented):
Reviewed the ranking path against the **live discovery MCP** — the gap the PR flags as unverified ("app-scoped token"). I pulled 30 rows across four views. The design is right (copyability over ROI, `blowup_risk` demoted first, per-trader `min_budget_usd`, the flag taxonomy) but it's ranking on fields the live payload can't support.

The core problem: **several numeric fields use `0` as a "missing" sentinel**, and nothing distinguishes that from a real zero. `gainToPainRatio` is `0` for 7/10 rows in MONTHLY/ROI and **6/6 — every row — in WEEKLY**, including a trader with $703,585 PnL. Four inline comments cover that and the fixes; a fifth is a routing conflict with the `strategy_create` tool contract that needs a product call, not a patch.

One correction to something I raised earlier: the engine correctly emits only a follower **count**, never the raw `followers[]` MIDs. Not an issue.

I didn't exercise the mirror execution path, DSL routing, or the 38 tests — this is the ranking path only.

senpi-trader-research/scripts/research.py
```diff
@@ -125,35 +151,278 @@ def _candidate(t):
...
+            concerns,
+            _REL_RANK.get(c.get("reliability"), 4),
+            -len(c.get("seen_in") or []),                     # cross-window confirmation (proven AND hot)
+            -(m.get("fresh_entry_surface_pct") or 0))
+
+
+def _fetch_view(client, meta, time_frame, sort_by, limit):
     try:
         resp = client.mcp_call("discovery_get_top_traders", time_frame=time_frame, sort_by=sort_by,
                                limit=limit, timeout=20)
```
shnoodles, line 467: **The API's own quality filters are never passed.** This is the only call site and it sends `time_frame`/`sort_by`/`limit` only — `consistency`, `open_position_filter`, `risk_labels`, and `activity_labels` all go unused, so the engine pays to filter client-side afterwards (hence the 20-deep enrichment and the ~90s timeout).

I ran the same G2P sort with `consistency: ["ELITE","RELIABLE"]` + `open_position_filter: true`. Every wiped account, the $1.07-volume trader, and both week-old accounts disappear; what comes back is `tcsValue` 58/75/58/**100**/50/75 with real ages and real books — including one ELITE trader at $2.64M PnL that the unfiltered call never surfaced.

**Suggested fix:** pass both. For mirroring, `open_position_filter: true` isn't optional — a trader with no open book gives a fresh mirror nothing to copy on day one, which is exactly the "looks broken" failure `DORMANT_DAYS` is trying to catch after the fact.

This one change does most of the work of comment #1 on its own.

senpi-trade/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+doesn't transfer), **protection is optional and stacks** (default: follow their exits). Never paraphrase a
+different version of this anywhere.
+
+### Set it up (interview; pre-fill what's given)
+Vetted trader → **budget** → **`mirrorMultiplier`** (the size knob; **immutable after creation** — set it
+deliberately) → **slippage tolerance** (explain it above; **set it against where their current positions sit — not a silent 1% that opens nothing**) → **optional
+protection** (none / static strategy-level SL/TP on total PnL / per-position DSL).
+
+### The hero check — simulate BEFORE funding
+Run `execution_estimate_position_opening` at the user's budget × multiplier × slippage **before** creating
```
shnoodles, line 198: **This contradicts the `strategy_create` tool contract**, and the agent gets both in the same turn. The MCP description says, verbatim:

> You MAY do at most ONE `discovery_get_trader_state` to sanity-check the trader, then call `strategy_create` — do NOT also chain `discovery_get_trader_history` / `discovery_get_top_traders` / `leaderboard` / `execution_estimate_position_opening`, and do NOT re-validate. […] do NOT poll `strategy_list`.

The skill mandates all three: the sim here (and as a stop-condition at :269), the whole `senpi-trader-research` engine upstream, and *"Poll `strategy_list` to ACTIVE"* at :201 / :116 / :222.

Which instruction wins is undefined, so mirror behaviour won't be reproducible between runs.

Not something to patch in this PR — it's a product call. My read is the **tool descriptions are the stale side**: they look written to stop wasteful chaining on a bare *"copy this OG"*, and this PR's whole premise is that unresearched mirroring gives bad advice. But it can't ship unreconciled. If the descriptions change, worth doing before merge so the skill and the tool agree on day one.

0xsarvesh:
## Review: run the engine against prod, then diff its model against the platform's own dry-run

**Method matters for this one.** I reviewed the diff, then ran `research.py` live against `mcp.prod.senpi.ai` and compared every judgment it produces against `execution_estimate_position_opening` — the platform's own answer to the same question. Two defects only show up that way, and they're the two I'd block on.

**What's solid first:** the plumbing is genuinely well built. The engine ran `--limit 12 --enrich-top 8` in 9.4s with `meta.warnings: []`, pure JSON on stdout and progress on stderr exactly as designed. It fails open on **every** edge case I threw at it — garbage address, well-formed-but-nonexistent address, `limit 0`, `limit 9999`, invalid token — all five return valid JSON with the cause in `meta.warnings`. `_positions_from_state` is correct: I recomputed TAO's `moved_from_entry_pct` independently from `returnOnEquity`/leverage and it matches to the decimal. The 0-sentinel diagnosis behind this PR is real and well-founded — live rows do come back `winRate: 100, maxDrawdown: 0, gainToPainRatio: 0`.

The problem is the **judgment layer**. The four numbers `SKILL.md` instructs the LLM to decide on — `trades`, `mirror_fit`, `min_budget_usd`, `momentum` — are wrong on all four, and they compose into a shortlist whose #1 pick is the worst available mirror.

---

## 🔴 Blockers

### 1. `_mirrorability` models slippage as symmetric; the platform gates directionally

`_mirrorability` buckets on `abs(moved_from_entry_pct) <= NEAR_ENTRY_BAND_PCT`. The platform's actual rule, verbatim from two dry-runs:

```
"Current price 86.358000 exceeds slippage price 84.768000 for BUY"    → longs gated UPWARD only
"Current price 0.055726 below slippage price 0.148350 for SELL"       → shorts gated DOWNWARD only
```

Only movement **in the OG's favor** is gated. A position that is *underwater for the OG* is a **cheaper entry than they got** and opens at any distance.

Live, on `0x08c14b32c8a48894e4b933090ebcc9ce33b21135`:

| | value |
|---|---|
| engine `fresh_entry_surface_pct` | 30.5% → `mirror_fit: "partial"` |
| direction-aware recompute | 71.8% → `good` |
| **platform, 5% slippage, $400** | **opens 6 of 10** |

The largest position it opens is ZRO at **−36.0% from entry**; MEGA opens at **−76.5%**. Both are scored as un-mirrorable by the engine. On my 8-row sample one candidate is mis-bucketed outright and one sits on the `good`/`partial` boundary.

This inverts the skill's central steer. `SKILL.md` tells the agent *"a trader whose winners have already run is un-mirrorable right now — the mirror opens nothing"* and to route to a fresh-entry template — for a book the platform would copy 6/10 of.

**Fix:** compute the signed move *against* the position's direction (`adverse = moved if long else -moved`) and gate on `adverse <= band`. Roughly a two-line change in `_mirrorability`.

### 2. `trades` is derived, not measured — and it's the field the skill judges on

`_candidate` computes `trades = averageTradesPerDay × (traderAgeSeconds / 86400)`. Account age is not active days, so it inflates. Live shortlist row **#1** right now:

```
0x645b...87b8    reliability: solid    trades: 546624    ← ranked #1
```

`averageTradesPerDay: 7071.47` × 77.3 days. Row #3 reads `10957`. Checked against ground truth (`realizedProfitAndLoss / averageProfitPerTrade`, which divides exactly):

| trader | engine `trades` | real closed |
|---|---|---|
| `0xe650…f1a9` | 67 | **3** |
| `0x3a23…35d6` | 44 | 20 |
| `0x810a…e29f` | 142 | 68 |

`0xe650…f1a9` is the case worth staring at: **3 closed positions, all three losses** (−$7.5k, −$9.9k, −$3.2k), `winRate: 100`, `maxDrawdown: 0`. The derived count sails past `MIN_TRADES_FOR_TRUST = 5`, `tcsLabel: ELITE` makes `_reliability` return **`solid`**, and the dd sentinel coercion means **no `blowup_risk` flag**. `SKILL.md` tells the agent to judge on "max-drawdown + closed-trade count + mirrorability" — one is fabricated, one is missing, one is inverted per #1.

**Fix:** `realizedProfitAndLoss / averageProfitPerTrade` gives the true count with no extra call, or use `discovery_get_trader_history`'s `page_info.totalCount`. Failing that, leave `trades` as `None` rather than deriving it — `None` already flows correctly through the `thin` gate.

---

## 🟠 High

**3. A failed state lookup is indistinguishable from a flat trader.** `_enrich_for_mirror` initializes `positions = []` and only warns on exception, so a timed-out `discovery_get_trader_state` yields `flags: ["no_open_positions"]`, `mirror_fit: "unknown"`. The skill then tells the user "their book is empty, the mirror starts idle" about a trader who may hold a full book — and `unknown` is `_FIT_BUCKET` 2, **worse than `poor`**, so the failure double-penalizes them in the sort. This fires on the skill's own stated most-common entry (a pasted address). **Fix:** `positions = None` on the exception path.

**4. Attribution is dropped for the highest-volume new flow.** `senpi-trade` calls `strategy_create` and `strategy_create_custom_strategy` directly and never sets `skillName`/`skillVersion` (both are real params on both tools). `senpi-trader-research/SKILL.md` simultaneously *removed* its attribution note by deferring to senpi-trade. Per CLAUDE.md every mirror created this way is orphaned. This is a new class — a skill that creates wallets without going through `deploy.py` — so the "utility skills carry no attribution" carve-out doesn't cover it.

---

## 🟡 Medium

**5. `min_mirror_budget` overstates ~1.9×.** Engine `$302.33` vs platform `minimumBudgetRequired: $158.34`; second trader `$211.16` vs `$152.01`. The formula matches the platform's math exactly — it's slippage-blind, counting positions that will be skipped. The skill quotes it as "the minimum to run properly" *before* the sim, at the moment the user decides whether to walk away.

**6. `opens_nothing_below_usd` can fall below the platform's own floor.** Live run produced `$3.12` and `$2.48`, but `strategy_create` enforces `initialBudget minimum: 10`. The agent states a minimum the fund call then rejects. Clamp both figures to `max(computed, 10)`.

**7. The blend merges time windows into one unlabeled column.** `roi_pct` / `pnl_usd` / `win_rate_pct` / `max_drawdown_pct` are window-scoped, but the merge keeps whichever view surfaced the address first and backfills from any other. So a row can carry 7d ROI beside 30d PnL — which is exactly the input to `roi_pnl_conflict`, meaning that flag can be a **cross-window artifact** reported to the user as a real contradiction. (This is what produced the flag on live row #3.) Either tag merged fields with their source view or restrict backfill to window-invariant fields.

**8. CI never runs these tests.** `.github/workflows/tests.yml:31` runs only `senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-trading-runtime/tests`. The 21 tests + ~190 new lines this PR is built around have zero regression protection. One-line fix.

**9. `momentum` measures 4h PnL direction, not activity.** All 8 live rows came back `cold`, every `delta_pnl_4h_usd` negative, `rank` ~98k–119k. `references/mirror-monitoring.md` ships the instruction *"the OG's `momentum` is `cold`/`flat` → the trader is idle, not a bug"* — so the agent would narrate row #1 (33 open positions, 7,071 trades/day) as idle. Either rename it to what it measures or derive activity from `last_trade_days_ago`.

**10. `_mirror_sort_key` counts flags without weighting them.** `concerns = len(flags)` treats `high_turnover` on a **7,071 trades/day** wallet — 884× the 8/day threshold — as equal to `concentrated_book`. Row #1 carries one flag and outranks row #2, which is `good` fit, 68.9% surface, `$18.63` min budget: the actually-copyable one. The engine's top recommendation is max fee-bleed at `$562.38` to run, and the skill's own doc says "you pay Hyperliquid trading fees on every mirrored fill."

**11. "Fresh-entry template (Shadow / Remora)" is wrong for Remora.** Catalog `sub_style`: only `shadow` is `fresh_entry_mirror`. `remora` is `named_whales`, `oxpecker` `elite_conviction`, `cuckoo` `copy_the_copiers`, `raptor` `hot_streak`; only `shadow/main/scanners/scan.py` diffs against `ctx.state` for newly-opened positions. The frontmatter blanket-claims all five "enter fresh-only", and the stale-book steer names Remora — sending the user to a template with the exact problem they're fleeing.

**12. The skill contradicts the tool description the agent also has in context.** `strategy_create`'s own text says *"do NOT also chain … `execution_estimate_position_opening` … do NOT poll `strategy_list` or call `strategy_get_clearinghouse_state`."* The skill mandates all three. Tracked as a follow-up in the PR body, but on a light model this decides at runtime whether the hero check happens at all — worth resolving before this ships rather than after.

---

## 🔵 Low

13. `_mirrorability` divides `near` (summed over the *scored* subset) by `total` (summed over **all** positions) — one position missing `entryPx` deflates the headline percentage and can demote the trader a bucket.
14. Momentum is fetched for the first `MOMENTUM_TOP = 10` in **blend** order, but `mirror_shortlist` is re-sorted by copyability — so the recommended row #1 can be `momentum: "unknown"` purely because the call was never made. `SKILL.md`'s gloss that "`unknown` just means not in the 4h-hot set" is then wrong.
15. `mirrorMultiplier` "**0.1×–4×**" in `mirror-trading-explained.md` is unsourced — `strategy_create` constrains it only to `exclusiveMinimum: 0`, no maximum, and no other file states a range. An invented bound in the file declared the single source, contradicting the same doc's advice to raise the multiplier to clear the $10 floor. (The adjacent "locked after creation" claim *is* correct.)
16. The `"Ranked the {label} leaders…"` progress line is emitted unconditionally after `_fetch_view`, including when it raised and returned `[]` — the user is told a view was ranked when nothing came back.
17. **Jackal** is missing from `mirror-trading-explained.md`'s template list while `senpi-strategy-discover` lists it and defers to senpi-trade for "the full flavor breakdown" — the doc pointed at as authoritative is the incomplete one. All nine templates exist in `catalog.json`.
18. `meta.degraded` reports `"no ranking data — check the token is USER-scoped"` for a `limit` validation error — wrong cause.
19. `ratchet_stop_add` is the protection primitive this skill leans on hardest and its payload is never given: `strategyId` + `strategy_wallet_address` + `ratchetStopConfig.tiered.tiers[{triggerRoe, lockRoe}]`. The skill cites `max_loss_pct`/`retrace_threshold` (the dropped fields) but never the shape that works.
20. Other under-specified primitives for a skill whose job is teaching them: `create_position`'s `orders:[]` nesting (the schema flags it "ARG SHAPE (critical)"), `strategy_get_asset_trading_limits` by name plus the fact that exceeding a leverage cap **silently clips** rather than erroring, and `strategyName`'s SERR055/056/058 rules.

---

## Verified clean

No `@senpi/` (missing `-ai`) anywhere. `senpi-strategy-discover/catalog.json` is byte-identical to `strategies/catalog.json`. All nine named templates exist as real `copy_trading` packages with `dsl_preset` and `marginPct`. No other skill re-explains mirror mechanics, so the single-source claim holds (Jackal aside). README `Ver` column matches every touched skill. All 21 research tests pass, and `_SeqClient` does properly exercise cross-view backfill.

---

**Reproduction** for #1 and #2 — `--limit 12 --enrich-top 8` against prod, then `execution_estimate_position_opening(ogTraderAddress=0x08c14b32…, totalBudget=400, slippagePercentage=5, mirrorMultiplier=1)` and compare `positionsToOpen` against the candidate's `mirrorability`.

Happy to push the patch for #1, #2, #3, #6, #8, #13, #16 and #18 — they're all small and local. #9, #10 and #12 are design calls that need your take.

0xsarvesh:
## Follow-up: end-to-end results against a live agent — plus a correction to my last comment

I installed this branch's skills on a Railway OpenClaw box (`samurai-pro`, all 15 skills registered, versions verified independently) and drove real agent turns via `openclaw agent --json`. Below is what the skills actually do in production, and one thing I got wrong above.

### ⚠️ Correction to my previous comment

I wrote that `minimumBudgetRequired` "equals `totalBudgetUsed` at every budget I tested, so that field tracks utilization rather than being a floor." **That was wrong** — I'd only tested traders where nothing was budget-skipped. On `0x3a23…35d6` at $10 the two diverge (`totalBudgetUsed: 7.98`, `minimumBudgetRequired: 10.38`, with VIRTUAL in `positionsSkippedInsufficientBudget`, shortfall $1.38). It **is** a genuine floor. Finding #5 stands on its own evidence; disregard that one sentence.

---

## What passed — and these are the load-bearing ones

**The pre-fund sim beats the conflicting tool description (my finding #12).** Asked to "mirror `0x3a23…` with 10 dollars, please set it up," the agent ran `execution_estimate_position_opening` **before** `strategy_create` and stopped for explicit confirmation — despite `strategy_create`'s own description saying "do NOT also chain `execution_estimate_position_opening`." The skill wins in practice. The contradiction is still worth resolving, but I'm downgrading it from a pre-merge blocker.

**The protection framing landed exactly.** Asked whether DSL can be added to a manual position, the agent produced a three-row table distinguishing profit-lock ladder / downside floor / coordinated stops, said `ratchet_stop_add` "silently drops the max-loss floor (Phase 1) without a runtime; only the profit-lock tiers stick," and called them "two *separate, uncoordinated* stops" — while avoiding the opposite failure of saying a raw position can't be protected. This is the nuance several commits went into and it survived into production output.

**Failure honesty passed under real conditions.** When the write path failed (`Plugin approval unavailable (no approval route)` — infra, not the skill), the agent surfaced the verbatim error, attempted a documented fallback, reported that failure too, and said: *"I won't pretend it went through — nothing was created, no positions were opened, and no funds moved."* No invented approval-gateway step.

**Discover's copy branch works.** "I want to copy other traders, what are my options?" returned both flavors, all six direct-mirror templates **including Jackal**, list-formatted options, and no proven-vs-hot window pick.

**Cross-skill routing works.** "My mirror hasn't traded in a week, should I close and re-set-up?" correctly identified the strategy wasn't a mirror at all, routed to ops, refused to blindly close+recreate, and warned that "a raw `strategy_close` would leave a ghost runtime registered."

---

## What the findings look like in production

Given a real "find me a trader to copy, I have $300" turn, the agent produced a genuinely excellent response — ranked shortlist, a *"Considered but skipped"* section naming five rejects with reasons, market cross-reference. But the numbers it told the user were wrong, and I checked each against the platform's own dry-run.

**Reject `0xe650…f1a9`** — agent said *"only **4% of the book is fresh** — the winners already ran and you'd be chasing"*, *"$161.09 min"*.

Platform at 5% slippage: **2 of 5 open.** AVGO (entry 416.34 → 394.10, **−5.3%**) and META (**−3.1%**) — both open precisely *because* they're underwater. Only MSFT (+19.2%), XYZ100 (+10.7%) and HOOD (+14.5%) skip. Budget used: **$19.27**, not $161.09 — an **8.4× overstatement**. AVGO is the largest position that opens. Both stated reasons are false.

**The ⭐ pick `0xfbc9…a361`** — agent said *"69% fresh… a mirror opens most of it near their entry"*, *"$18.60 min"*. Platform: **1 of 3 opens**; budget used at $300 is **$58.16**, a **3.1× understatement**.

That sharpens **#5**: `min_mirror_budget` scales off position **notional** while the platform scales off **margin**, so it errs in both directions unpredictably — 8.4× over on one trader, 3.1× under on the next.

And it refines **#1** usefully: `abs()` is **correct** when a position ran in the OG's favor, **wrong** when it's underwater for the OG. `0xfbc9`'s shorts (which ran −23.5%/−24.2% in favor) made the model look right; `0xe650`'s underwater longs exposed it.

Also confirmed verbatim in user-facing output:
- **#2** — "76 trades over 572 days", "525 trades/day" quoted as fact. *Usefully bounded*: the **vet** path uses real closed counts from `discovery_get_trader_history` (it quoted 24 for `0x3a23…`, not the engine's derived 44). Only the **find** path propagates the fabricated number.
- **#7** — the cross-window artifact reached the user: *"283% ROI… but PnL is −$1.7M. The ROI is a paper-gain illusion"* — 7d ROI beside 30d PnL, narrated as a real contradiction.
- **#11** — **three separate turns** offered Remora as a "fresh-entry template," including as the fix for a stale book.
- **#14** — momentum rendered as `—` for the re-sorted rows, exactly the predicted gap.

Two findings were **self-corrected by the model**: it demoted the high-turnover wallet in prose despite the sort ranking it #2 (#10), and used `last_trade_days_ago` rather than `momentum` to judge idleness (#9). Both remain worth fixing — that mitigation is model judgment, not a property of the engine.

---

## One new finding

**At small budgets the copy is equal-weighted, not proportional.** Simulating `0x3a23…35d6` at $10, every opened position lands at `notionalValue` ≈ **$12.00** against `cpNotionalValue` (the proportional target) of 9.00 / 4.49 / 3.22 / 2.51 / 2.50 / 2.43. The ~$12 floor-bump overrides proportional sizing entirely, so a small mirror holds six roughly equal positions rather than a scaled copy of the trader's book.

This contradicts `mirror-trading-explained.md`'s central sizing claim — *"Your capital utilization mirrors the trader's… proportional copying **preserves** their utilization %"* — for exactly the budget range where users need the explanation most. The doc's "$10 floor causes skips" framing covers the positions that drop out but not the distortion in the ones that remain.

---

## Untested

`go long HYPE 10x` (the Branch A managed-vs-one-off fork) and the funded execution path — the box has no approval channel configured, so no write op could complete. Everything up to the money movement is verified.

betashop:
## Review addressed — thanks, this was the review that mattered

The execution-diff method (engine output vs `execution_estimate_position_opening`) caught what the data-only pass couldn't. Fixes in `e8311e6` (engine) + `e324533` / `c321f32` (skills/docs/CI). **17 of 20 done, 3 deferred with reasons.**

**🔴 Blockers — both fixed**
- **#1** `_mirrorability` now gates the *adverse* move, not `abs()` — underwater-for-the-OG positions correctly count as openable (shared `_adverse_move` helper; `_min_mirror_budget` reuses it).
- **#2** `trades` = `realizedPnL / averageProfitPerTrade` (true closed count); the fills-based age×rate derivation is gone (leaves `None` when it can't divide).

**🟠 High — fixed:** #3 a failed state lookup → `positions=None` (unknown, not a false empty book); #4 `senpi-trade` passes `skillName`/`skillVersion`.

**🟡 Medium — fixed:** #5 min-budget slippage-aware + clamped + reframed as a rough pre-sim estimate; #6 clamp to \$10; #7 window-scoped fields no longer backfilled across views (kills the false `roi_pnl_conflict`); #8 CI now runs `senpi-trader-research` + `senpi-market-pulse` tests; #9 momentum documented as 4h PnL direction (idleness keys on `last_trade_days_ago`); #10 heavy flags weighted; #11 the "fresh-entry" mislabel + Remora stale-book steer fixed (full menu kept, honest labels, Jackal added).

**🔵 Low — fixed:** #13, #14 (gloss), #15, #16, #17, #18 — plus a note that small budgets are NOT truly proportional (the \$12 floor equal-weights the survivors).

**Deferred — non-blocking:**
- **#12** — you downgraded it yourself (the pre-fund sim wins in practice); product/MCP-side reconciliation.
- **#14 (deep)** — enrich momentum *after* the copyability sort; the gloss is fixed, the structural tidy is a tracked follow-up.
- **#19 / #20** — under-specified primitives (`ratchet_stop_add` payload shape, `create_position` `orders[]` nesting, etc.). Deliberately **not** guessed: that `ratchet_stop_add` schema is the open ergonomics issue, and a wrong payload in a public skill is worse than a gap. Needs the verified shape first — happy to take these once we lock it.

**+5 tests (26 in `test_research.py`); the 247-test CI set is green.**

One honest caveat: I fixed the logic to your exact prescriptions and the tests confirm it, but I couldn't re-run your execution-diff (this session's token is app-scoped, 401s on discovery). **A re-run of `execution_estimate_position_opening` vs the engine is the real confirmation** — over to you for the re-review.

0xsarvesh:
## Re-review of the three fix commits — verified against prod

Re-ran every original reproduction against live data on `c321f32`. **Most of this is genuinely fixed and I confirmed it end-to-end.** One fix doesn't work, and it's the one that matters most. One new issue surfaced as a side effect of an otherwise-correct fix.

---

## ✅ Fixed and verified

**#1 direction-aware slippage — the headline fix, and it's right.** `_adverse_move()` gates on `moved if long else -moved`, matching the platform's BUY/SELL semantics exactly. The engine's `openable` count now matches `execution_estimate_position_opening` **exactly on all three** reproduction traders:

| trader | engine `openable` | platform actually opens |
|---|---|---|
| `0xe650…f1a9` | 2 | **2 of 5** ✓ |
| `0xfbc9…a361` | 1 | **1 of 3** ✓ |
| `0x3a23…35d6` | 6 | **6 of 9** ✓ |

`0xe650` moved 4% → **16.0%** surface, which I verified is arithmetically correct (AVGO $49.2k + META $18.4k against a $420k book). The "the winners already ran, the mirror opens nothing" mis-steer is gone.

**#11 is now precisely stated** — *"Only Shadow / Jackal enter fresh-only; the rest open the cohort's current positions, so they don't cure an already-run book."* That's exactly the distinction, and it fixes the steer that appeared in three separate live turns.

**#17 Jackal** added, and I verified the new claim rather than taking it on faith: `jackal/main/scanners/scan.py` has `_DEFAULT_MAX_ENTRY_AGE = 600` — "only mirror entries < 10 min old." Accurate.

Also confirmed fixed: **#13** (`scored_total` denominator), **#7** (`_WINDOW_SCOPED` — and keeping `max_drawdown` backfillable is the right call, worst-drawdown-in-any-window is a legitimate risk signal), **#10** (`_HEAVY_FLAGS` ranked above flag count), **#16**, **#18**, **#6** (clamped to `MIN_STRATEGY_BUDGET_USD`), **#15** (multiplier corrected to "greater than 0" — matches the schema), **#3** (`positions = None` with the `st is not None` guard), **#8** (CI now runs trader-research + market-pulse; 41 tests pass locally).

The equal-weighting finding is documented in the single-source doc too.

---

## 🔴 #2 is not fixed — it's disguised, which is worse

`realizedPnL / averageProfitPerTrade` doesn't yield closed-position count on the payload `vet_trader` actually uses. Ground truth is `discovery_get_trader_history` → `page_info.totalCount`:

| trader | before | **after the fix** | real |
|---|---|---|---|
| `0x3a23…35d6` | 44 | **132** | **20** |
| `0xe650…f1a9` | 67 | **98** | **3** |
| `0xfbc9…a361` | 76 | **111** | **0** |

`0xfbc9…a361` has **zero closed positions** and the engine now reports **111**. That's the textbook `thin_track_record` case, and 111 clears `MIN_TRADES_FOR_TRUST = 5` exactly as 546,624 did — so the guard is still bypassed and the trader still reads as a real record.

This is arguably more dangerous than before: 546,624 was self-evidently absurd and a careful reader would catch it. 111 looks entirely plausible and will be narrated to users as fact.

The ratio *does* divide cleanly on the WEEKLY rows I originally sampled (49227.547 / 6153.443 = 8.000), which is presumably why it looked correct — it doesn't hold on the ALL_TIME payload the vet path uses.

**Suggested fix:** `discovery_get_trader_history(limit=1)` → `page_info.totalCount` is exact and costs one call, free in `vet_trader` which is already making several. For the 20-candidate find path, either accept the calls or leave `trades: None` and have the skill say the count isn't established — `None` already flows correctly through the `thin` gate, and "unknown" is a far better thing to show a user than a fabricated number.

---

## 🟠 New: `min_mirror_budget` now overstates ~20× on diversified books

The openable-only filter fixed the concentrated cases, but it exposed the notional-vs-margin mismatch much more severely than "treat this as an estimate" conveys:

| trader | before | after | platform | |
|---|---|---|---|---|
| `0xe650…f1a9` | $161.09 | $76.29 | ~$84 | ✓ good |
| `0xfbc9…a361` | $18.60 | $10.00 | $10 opens SOL | ✓ correct |
| `0x3a23…35d6` | — | **$212.69** | **$10.38** | ✗ **20×** |

Root cause is precisely what the updated docs now describe but the code still doesn't model: the platform **bumps** a sub-floor position *up* to ~$12 notional and only needs the **margin** for it (VIRTUAL: $2.40 at 5x). The engine assumes such positions are skipped and solves for proportional notional ≥ $12.

Concretely: a user asking "how much do I need to mirror `0x3a23…35d6`" is told **$213** when **$11** actually opens the whole slippage-passing book. That's the number the skill instructs the agent to quote as a factual floor *before* the sim runs — i.e. at the exact moment a user decides whether to walk away.

A margin-based estimate would land it: `Σ (MIN_NOTIONAL_USD / leverage)` over openable positions ≈ **$10.38** here, matching the platform's `minimumBudgetRequired` exactly.

---

## Still open

- **#14** untouched — momentum is still gated on `i < MOMENTUM_TOP` in *blend* order while `mirror_shortlist` re-sorts by copyability, so the recommended row #1 can still show `unknown` purely because the call was never made.
- **#4 partially fixed** — attribution is now required at `senpi-trade/SKILL.md:204`, but that sits in **Branch B (mirror)**. Branch A's manual-open path (line 115, `strategy_create_custom_strategy` / `create_position`) has no attribution reminder, so a model reading only the manual branch still creates an orphaned wallet.
- **#19/#20** untouched — `ratchet_stop_add`'s payload shape (`ratchetStopConfig.tiered.tiers[{triggerRoe, lockRoe}]`), `create_position`'s `orders:[]` nesting, `strategy_get_asset_trading_limits` and the silent leverage clip.
- **#12** unresolved as a doc conflict, though live E2E showed the skill wins in practice — the agent ran the sim before `strategy_create` despite the tool description forbidding it.

Happy to re-run the live agent scenarios against these updated skills — the `0xe650` "4% fresh, the winners already ran" line should now read materially differently, and that's the cleanest way to confirm the #1 fix reaches users rather than just the engine.

0xsarvesh:
## Correction — the #2 numbers above, re-sourced from prod

The "actual closed" counts in my re-review came from **`senpi-dev`** while the engine's numbers came from **prod** (`mcp.prod.senpi.ai`, per the engine's own `SENPI_MCP_URL`). Not a valid comparison. I've re-pulled **both sides from prod** through the engine's vendored `mcp_client`. The finding is unchanged and the gap is wider than I reported — but two figures were wrong, including one claim stated too strongly.

**Corrected, all prod:**

| wallet | engine `trades` | actual closed | ratio |
|---|---|---|---|
| `0xfbc9…a361` | **111** | **2** | 55× |
| `0xe650…f1a9` | **98** | **3** | 33× |
| `0x3a23…35d6` | **132** | **25** | 5.3× |

I previously wrote that `0xfbc9…a361` had **zero** closed positions — it has **2** on prod. And `0x3a23…35d6` is **25**, not 20. Please check against this table, not the one above it.

### A single wallet to reproduce against

**`0xfbc9638938262f15cc1b3a751a54d538559da361`** — two calls, both prod:

```
discovery_get_top_traders(time_frame="ALL_TIME", addresses=["0xfbc9…a361"], limit=1)
  realizedProfitAndLoss = 821,679.97
  averageProfitPerTrade =   7,402.52
  821679.97 / 7402.52   = 111          ← what _candidate() computes today

discovery_get_trader_history(trader_address="0xfbc9…a361", limit=1)
  page_info.totalCount  = 2            ← the real closed-position count
```

A wallet with **2 closed positions** is reported as **111**, clearing `MIN_TRADES_FOR_TRUST = 5` by 22×, so `thin_track_record` never fires on a record that barely exists.

### Why the ratio looked right

It divides exactly on the **WEEKLY** rows (`49227.547 / 6153.443 = 8.000`) — which is the payload I sampled when I first suggested this approach, and I should have checked it against ALL_TIME before recommending it. It does not hold on **ALL_TIME**, which is what `vet_trader` requests. My suggestion was the source of the bad fix; apologies for that.

`page_info.totalCount` from the call above is exact, and `vet_trader` is already making several calls, so it's effectively free there. For the find path the honest fallback is still to leave `trades: None` — it flows correctly through the `thin` gate, and "not established" beats a fabricated count.

Everything else in the re-review — the verified #1 fix, the `min_mirror_budget` 20× overstatement on `0x3a23…35d6` ($212.69 vs the platform's $10.38), and the still-open items — is unaffected.

0xsarvesh:
## `min_mirror_budget`: the exact derivation, and the one-line cause

Following up on the 21× overstatement — here's precisely where $212 comes from and what it should be. All numbers pulled from prod at the same moment, on `0x3a23b10845e911b6944d9dc2f1f480fca3af35d6`.

### What the engine computes

```
min_budget_usd = max(10, MIN_NOTIONAL_USD × account_value / smallest_non_dust_openable_notional)
```

`_min_mirror_budget()` filters to openable positions, drops dust (< 1% of openable notional), takes the **smallest survivor**, and solves for the budget that would make *that* position's proportional copy reach the $12 minimum notional.

| | |
|---|---|
| `account_value` | $497,199.95 |
| openable notional total | $1,177,414.95 |
| dust cutoff (1%) | $11,774.15 |
| smallest **non-dust** openable | **VIRTUAL — $28,111.50** |

```
12 × 497,199.95 / 28,111.50 = $212.24
```

The premise: your copy of a position ≈ `budget × (position_notional / account_value)`. Set that to $12, solve for budget. To make VIRTUAL's *proportional* slice reach $12 of notional you would indeed need ~$212.

### What the platform actually requires

It never requires the proportional slice to reach $12. It **bumps the order up** to the $12 floor and charges only the **margin**, `$12 / leverage`:

| position | leverage | margin for a $12 notional |
|---|---|---|
| TAO | 5x | $2.40 |
| ETH | 25x | $0.48 |
| UNI | 10x | $1.20 |
| HYPE | 10x | $1.20 |
| APE | 5x | $2.40 |
| VIRTUAL | 5x | $2.40 |
| | | **Σ = $10.08** |

`execution_estimate_position_opening(totalBudget=10, slippagePercentage=5, mirrorMultiplier=1)` on the same wallet returns `minimumBudgetRequired: 10.08`. Exact match.

(At `totalBudget=300` the same field reads `98.95` — that's just what $300 deploys at this trader's ~20% utilization, since nothing is budget-skipped. The floor only surfaces when the budget is near it, which is why $10 is the right probe.)

### The cause, in one line

**The formula has no leverage term.** The engine scales with `account_value / position_notional`; the platform scales with `1 / leverage`. Unrelated quantities — so the result is off by roughly whatever the leverage is. Here that's 21×.

Two secondary effects of the same formula:

- **Hypersensitive to the dust cutoff.** VIRTUAL ($28.1k) sits just above the 1% line ($11.8k). Had it fallen below, the calc would use APE instead: `12 × 497,200 / 122,790 = $48.60`. The headline number swings ~4× on whether a single position clears an arbitrary threshold.
- **`opens_nothing_below_usd`** ($13.21, derived from the largest position) has the same defect — also notional-based.

### Suggested fix

```
min_budget_usd ≈ Σ (MIN_NOTIONAL_USD / leverage)  over openable positions
```

Lands on $10.08, matching the platform.

One prerequisite: `_positions_from_state()` currently extracts `asset, direction, notional, upnl, roe_pct, entry_px, mark_px, moved_from_entry_pct` — **not leverage**. The state payload carries it per position as `leverage: {type, value}`, so that field needs capturing before the sum is possible.

This matters because the skill instructs the agent to quote this figure as a factual floor *before* the sim runs — i.e. exactly when a user decides whether to walk away. A user asking "how much do I need to mirror this trader" is currently told **$213** when **$11** opens the whole slippage-passing book.

betashop:
## Both caught — fixed in `b620167`

Thanks for re-running against prod; both were real and your prescriptions were exact.

**#2 trades — the disguised one.** Removed the `realizedPnL / averageProfitPerTrade` derivation entirely (you're right that it only divides cleanly on WEEKLY, not the ALL_TIME payload `vet_trader` uses — 111 for a 2-closed record). Now:
- **find path** leaves `trades` **None** — "not established" over a fabricated number, and None flows through the `thin` gate. The skill's reliability rule now says *don't quote a closed-trade count off the find shortlist; vet to confirm.*
- **vet path** pulls the exact count from `discovery_get_trader_history` → `page_info.totalCount`, then recomputes reliability — so a 2-closed record now trips `thin` despite an ELITE label (regression-tested with a `totalCount: 2` fixture).

**min_mirror_budget — the ~20× overstatement.** Rewrote it **margin-based**, exactly your model: the platform bumps a sub-floor position to the ~\$12 notional minimum and needs only `\$12 / leverage` in margin, so the estimate is `Σ (MIN_NOTIONAL_USD / leverage)` over openable positions (I now extract per-position leverage in `_positions_from_state`). Clamped to \$10. On `0x3a23…35d6` this should land ~\$10.38 against the platform's `minimumBudgetRequired`, not \$213.

**Also fixed from "still open":**
- **#14** — momentum is now fetched **after** the copyability sort, for the top of the shortlist, so the recommended row #1 carries it instead of `unknown` (it was gated on blend order).
- **#4** — Branch A's manual-open path (`strategy_create_custom_strategy` / `create_position`) now carries the `skillName`/`skillVersion` reminder too, not just Branch B.

**Still deferred:** #19/#20 (the `ratchet_stop_add` / `create_position` payload shapes — I won't paste an unverified schema into a public skill; happy to add them once we lock the shape), and #12 (the doc conflict your own E2E showed the skill wins in practice).

**Caveat:** I can't re-run your execution-diff from here (this session's token is app-scoped, 401s on discovery), so these rest on matching your formulas + the offline tests. **Your offer to re-run the live agent scenarios is the real confirmation** — the `0xe650` "4% fresh, the winners already ran" line should now read ~72% fresh / opens 6-of-10, and the min-budget prompts should read ~\$10, not \$200+.

0xsarvesh (approved):
## Round 2 verified on prod — both findings fixed. Approving.

Re-ran every reproduction against `mcp.prod.senpi.ai` on `1271aa6`.

### #2 `trades` — fixed

The find path now leaves `trades: None` rather than deriving it, and the vet path pulls `page_info.totalCount` from `discovery_get_trader_history` — then **recomputes `_reliability`**, which is the part that makes it actually bite:

| wallet | before | now | ground truth | reliability |
|---|---|---|---|---|
| `0xfbc9…a361` | 111 | **2** | 2 | **thin** ✓ |
| `0xe650…f1a9` | 98 | **3** | 3 | **thin** ✓ |
| `0x3a23…35d6` | 132 | **25** | 25 | choppy |

The 2-closed-trade wallet that previously read as a real record now trips `thin_track_record`. That was the whole point.

Leaving `trades: None` in the find path does weaken thin-detection there (only `active_days` can trigger it now), but "not established" beating a fabricated number is the right trade, and vet is where the due-diligence claim gets made.

### #5 `min_mirror_budget` — fixed, and exact

`Σ (MIN_NOTIONAL_USD / leverage)` over openable positions, with `leverage` newly captured in `_positions_from_state`. On `0x3a23…35d6`:

**engine `$10.08` vs the platform's `minimumBudgetRequired` `$10.08`** — exact, down from `$212.69`.

The other two clamp to `$10.00`, correctly: the platform's own `initialBudget` minimum is the binding constraint there (raw values `$2.40` and `$1.83`).

### #14 momentum — fixed as a bonus

Now fetched after the copyability sort. **0 rows `unknown`**, and `hot` appears for the first time (rows 2 and 4 of a live 8-row shortlist). Previously all eight read `cold` because only the blend-order top-10 got the call and those happened to be uniformly cold — so this was suppressing real signal, not merely leaving gaps.

Also confirmed working: `_HEAVY_FLAGS` demotes the 7,071-trades/day wallet `0x645b…87b8` from **#1 to #3**.

27 tests pass locally; CI now covers this suite.

---

## Three nits — non-blocking, worth a follow-up sweep

1. **`MIRROR_DUST_FRAC` (line 44) is dead code.** The dust logic went with the notional formula; nothing references the constant now.
2. **`mult` is accepted and reported as `at_multiplier`, but never used arithmetically.** `mult = mult or 1.0` is bound then ignored — the old `_b()` divided by it, `sum(margins)` doesn't. The figure is still a defensible floor (the $12 bump makes the margin leverage-bound regardless of multiplier), but reporting `at_multiplier: 4` beside a number computed without it implies it was applied. Either apply it or drop the field.
3. **`account_value` is now an unused parameter** — fine as a signature-stability choice, but worth folding into the same cleanup as (1).

One to confirm rather than change: `min_budget_usd` returns **`None`** when nothing is openable (two rows in my live run, both `mirror_fit: poor`). Semantically correct, but `SKILL.md` instructs the agent to quote this figure — worth checking it narrates "nothing opens right now" instead of silently dropping the line.

## Still open from the original review (tracked, not blocking)

- **#4 partial** — attribution is required at `senpi-trade/SKILL.md:204`, but that sits in Branch B (mirror). Branch A's manual-open path (`strategy_create_custom_strategy` / `create_position`, ~line 115) has no attribution reminder, so a model reading only the manual branch still creates an orphaned wallet.
- **#19/#20** — `ratchet_stop_add`'s payload shape, `create_position`'s `orders:[]` nesting, `strategy_get_asset_trading_limits` + the silent leverage clip.
- **#12** — the `strategy_create` tool description still forbids the chain this skill mandates. Live E2E showed the skill wins in practice on `samurai-pro`, so it's a docs-consistency issue rather than a functional one.

Good iteration on the two that mattered — both were verified against the platform's own dry-run rather than by reading the diff, and both now match it. **Approving.**

PR #538, opened 2026-08-14
fee-honest realized PnL: net-of-fees in improve-trades + portfolio (was reporting gross as booked)

betashop:
## The bug

`senpi-improve-trades/scripts/review.py` and `senpi-portfolio/scripts/portfolio.py` both reported realized PnL sourced straight from `discovery_get_trader_history` — which is **GROSS** (HL `closedPnl` is price-PnL, fees are a separate field). Both narrated it to the user as **"TOTAL PnL" / "total booked PnL / real gains booked."**

On a churny book that reads **~5× rosier than the wallet**. Found live on a funded account where the agent praised a strategy at "+$19, doing its job well" that was **+$2 net lifetime** — fees had eaten essentially all of it — while the user stared at a truthful (worse) wallet balance. It is **not a prompting slip**: the schema had nowhere to put fees, and the only fee-aware path (a separate subcommand) could time out, leaving its "fees UNDETERMINED" caveat orphaned from the numbers it invalidated.

Reported by @shnoodles; portfolio.py surfaced by a fleet sweep of the same idiom.

## The fix — both skills, identical schema

- **Fees come from the HL fills ledger** (`userFills.fee`, **already builder-fee-inclusive** — never add builder fee on top). improve-trades attributes a round-trip fee per trade (matched open + close, pro-rated); portfolio sums the window.
- Every PnL surface now carries **`gross_realized_pnl` / `fees` / `net_realized_pnl` / `fees_status`**, plus `gross_total` / `net_total` on the headlines.
- **Degradation is first-class:** a failed/empty fills read → `fees`/`net_*` = `null`, `fees_status: "undetermined"` — **never a fake $0**. The only $0-fees branch is a genuinely-no-closed-trades wallet.
- **SKILL narration now LEADS with net**; `gross_*` is labeled pre-fee; the old "authoritative fee $ is a pending upgrade" note is now "wired."
- Bonus: fees are now decoupled from the flaky telemetry event-log read (they come from the durable fills ledger, not the subcommand that timed out).

## Sweep result

The same idiom was audited across the repo. **Only these two** compute the *user's own* realized PnL as net — everything else is an external trader's copy-ranking stat, a cohort threshold, or a clearly-labeled platform passthrough (not this bug). **Out of repo:** `strategies-of-the-week` and `funnel-report` were named as suspects but live elsewhere — flagged for a **separate audit**.

## Tests & versions

- improve-trades **1.8.0 → 1.9.0**: 73 pass, incl. fee-attribution (discovery rows get the fills-ledger fee, net = gross − fees) + undetermined-degradation (no fills read → `null`, not $0).
- portfolio **1.13.3 → 1.14.0**: 72 pass (+6, incl. group-total goes undetermined if any sleeve is).
- Both skills **added to CI** (`.github/workflows/tests.yml`) — they weren't run before. README version table synced.

### DIFF
--- a/senpi-improve-trades/scripts/review.py
+++ b/senpi-improve-trades/scripts/review.py
@@ -46,7 +46,9 @@ CLOSED_HISTORY_PULL = 100    # closed positions to pull per wallet before the wi
 WINDOW_DEFAULT_DAYS = 7      # the default review window
 TOP_MOVERS_CAP = 12          # cap the book-vs-market movers surfaced
 HL_INFO_URL = "https://api.hyperliquid.xyz/info"   # public, no-auth — durable closed-trade recovery
-HL_FILLS_CAP = 2000          # userFills returns the most recent N fills; ample for any review window
+HL_FILLS_CAP = 2000          # userFills / userFillsByTime page size — the most recent (or per-page) N fills
+FILL_LOOKBACK_MS = 90 * 24 * 3600 * 1000   # page fills from since−90d so a closed trade's OPEN leg is captured
+FILL_PAGE_CAP = 25           # userFillsByTime page backstop (×~2000) — never hit in practice; stops a runaway
 
 # The runtime registers every deployed strategy in installed_runtimes.json in the state dir — the
 # UNIVERSAL source of a strategy's mandate (the runtime.yaml `description`) + its DSL ladder. Reused
@@ -778,6 +780,44 @@ def _hl_info(payload, meta, client=None, timeout=12):
         return None
 
 
+def _fetch_window_fills(client, wallet, since_ms, until_ms, meta):
+    """The authoritative fee source — HL fills covering the review window. `userFills` returns only the most
+    recent ~2000 fills with NO paging, so on a long / high-turnover window an older trade's fills fall off the
+    slice and its fee silently reads UNKNOWN (blanking net for the whole wallet). When a window is set we page
+    `userFillsByTime` from since−lookback .. until, so COVERAGE MATCHES the window being reported (the lookback
+    captures the OPEN leg of a position that closed inside the window). Ascending pages, deduped on the boundary
+    overlap. Fails OPEN → None (a failed read) — distinct from [] (genuinely no fills). No window → the recent
+    `userFills` slice (the standalone / test path)."""
+    if since_ms is None or until_ms is None:
+        return _hl_info({"type": "userFills", "user": wallet}, meta, client)
+    cursor, end = max(0, int(since_ms) - FILL_LOOKBACK_MS), int(until_ms)
+    out, seen = [], set()
+    for _ in range(FILL_PAGE_CAP):
+        page = _hl_info({"type": "userFillsByTime", "user": wallet,
+                         "startTime": cursor, "endTime": end}, meta, client)
+        if page is None:
+            return out if out else None          # earlier pages are still usable; nothing yet → propagate failure
+        if not isinstance(page, list):
+            break
+        fresh, last_t = 0, cursor
+        for f in page:
+            if not isinstance(f, dict):
+                continue
+            key = (f.get("tid"), f.get("oid"), f.get("time"), f.get("dir"), f.get("sz"), f.get("px"))
+            if key in seen:
+                continue
+            seen.add(key)
+            out.append(f)
+            fresh += 1
+            ft = _num(f.get("time"))
+            if ft is not None and int(ft) > last_t:
+                last_t = int(ft)
+        if len(page) < HL_FILLS_CAP or fresh == 0 or last_t <= cursor:
+            break                                # last (short) page, or no forward progress → done
+        cursor = last_t                          # advance; the dedup set absorbs the boundary overlap
+    return out
+
+
 def _reconstruct_closed_from_fills(fills, since_ms, until_ms, cap):
     """Rebuild round-trip CLOSED trades from raw HL fills — the durable, close-independent source. HL gives
     per-fill `dir` ('Open/Close Long/Short'), px, sz, closedPnl, fee, time, oid. Walk fills per coin
@@ -801,7 +841,7 @@ def _reconstruct_closed_from_fills(fills, since_ms, until_ms, cap):
         if d.startswith("Open"):
             # carry the OPEN-side fee + original size so a later Close can pro-rate the matched share.
             lots[coin].append({"px": px, "sz": sz, "orig_sz": sz, "time": t,
-                               "fee": abs(_num(f.get("fee")) or 0.0)})
+                               "fee": _num(f.get("fee")) or 0.0})    # SIGNED — a maker rebate is a NEGATIVE fee
         elif d.startswith("Close"):
             side = "long" if "Long" in d else ("short" if "Short" in d else None)
             remaining, entry_notional, matched, open_time = sz, 0.0, 0.0, t
@@ -831,9 +871,10 @@ def _reconstruct_closed_from_fills(fills, since_ms, until_ms, cap):
                 "open_time": open_time,
                 "close_time": t,
                 "closed_order_id": f.get("oid"),
-                # round-trip trading fee = matched OPEN-side fees + this CLOSE fill's fee. HL `fee` ALREADY
-                # includes the builder fee, so this is the whole cost — never add a builder fee on top.
-                "fee": round(open_fee + abs(_num(f.get("fee")) or 0.0), 4),
+                # round-trip trading fee = matched OPEN-side fees + this CLOSE fill's fee. HL `fee` is SIGNED
+                # (a maker rebate is NEGATIVE — the wallet was paid) and ALREADY includes the builder fee, so
+                # sum it as-is: never abs() it (that turns rebates into costs) and never add a builder fee.
+                "fee": round(open_fee + (_num(f.get("fee")) or 0.0), 4),
                 "source": "onchain_fills",
             })
     out = []
@@ -899,10 +940,10 @@ def fetch_closed_trades(client, wallet, since_ms, until_ms, cap, meta):
             "source": "discovery",
         })
     # FEES: the fills ledger is the ONLY authoritative fee source — discovery_get_trader_history carries no
-    # fee, so the realized_pnl above is GROSS. Pull userFills ONCE and use it both to attribute a round-trip
+    # fee, so the realized_pnl above is GROSS. Pull the window fills ONCE (userFillsByTime, paged) and use them to attribute a round-trip
     # `fee` to each discovery trade and (when discovery is empty) to reconstruct the closed book. `_hl_info`
     # fails OPEN → None (distinct from an empty list): None ⇒ fee UNDETERMINED for these trades, never a fake $0.
-    fills = _hl_info({"type": "userFills", "user": wallet}, meta, client)
+    fills = _fetch_window_fills(client, wallet, since_ms, until_ms, meta)
     recon = _reconstruct_closed_from_fills(fills, since_ms, until_ms, None) if isinstance(fills, list) else []
     fee_by_oid = {}
     for r in recon:
@@ -1438,6 +1479,8 @@ def _timing_summary(trades):
         "fees_total": fees_total,              # window trading fees (builder-fee-incl); None = UNDETERMINED
         "net_realized_pnl_total": net_realized_total,   # gross − fees; None when fees undetermined — LEAD with this
         "fees_status": ("ok" if fees_total is not None else "undetermined"),
+        "fees_coverage": {"resolved": sum(1 for f in _fees if isinstance(f, (int, float)) and not isinstance(f, bool)),
+                          "total": trade_count},
         "if_all_reclosed_now_total": if_all,   # counterfactual aggregate — CONTEXT, NEVER a projection/target
         "by_asset_class": by_class,
     }
@@ -1726,6 +1769,7 @@ def _strategy_reads(trades, strategies, open_book=None):
             "gross_total_pnl": gross_total,         # gross_realized + unrealized; None when unrealized UNKNOWN
             "net_total_pnl": net_total,             # net_realized + unrealized; None when either UNKNOWN
             "fees_status": fees_status,             # ok | undetermined — never render gross as net when undetermined
+            "fees_coverage": {"resolved": agg.get("fees_known", 0), "total": cnt},  # net knowable for resolved/total trades
             "open_position_count": len(open_positions),
             "open_positions": open_positions,    # asset/direction/unrealized_pnl/return_on_equity_pct/entry/lev
             "on_mandate_note": note,
@@ -1759,6 +1803,7 @@ def _closed_strategy_rollup(trades, strategies):
             "fees": _fees,                          # None = UNDETERMINED (fills read failed for some trade)
             "net_realized_pnl": (round(agg["pnl"] - _fees, 2) if _fees is not None else None),
             "fees_status": ("ok" if _fees is not None else "undetermined"),
+            "fees_coverage": {"resolved": agg.get("fees_known", 0), "total": cnt},
         })
     return out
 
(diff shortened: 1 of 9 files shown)


senpi-improve-trades/scripts/review.py
```diff
@@ -821,12 +826,14 @@ def _reconstruct_closed_from_fills(fills, since_ms, until_ms, cap):
...
-                "realized_pnl": round(_num(f.get("closedPnl")) or 0.0, 2),   # HL's own realized — authoritative
+                "realized_pnl": round(_num(f.get("closedPnl")) or 0.0, 2),   # HL's own realized — GROSS (pre-fee)
                 "margin_used": None,
                 "open_time": open_time,
                 "close_time": t,
                 "closed_order_id": f.get("oid"),
-                "fee": _num(f.get("fee")),
+                # round-trip trading fee = matched OPEN-side fees + this CLOSE fill's fee. HL `fee` ALREADY
+                # includes the builder fee, so this is the whole cost — never add a builder fee on top.
+                "fee": round(open_fee + abs(_num(f.get("fee")) or 0.0), 4),
```
shnoodles: **Sign bug: `abs()` converts maker rebates into costs.**

HL's `fee` is signed. A maker fill on a rebate tier returns a **negative** fee, meaning the wallet was paid. `abs()` flips it to a cost, so net PnL is understated by twice the rebate. Same issue on the open lot at line 804.

This is not theoretical. Across `hyperliquid_raw_fills` since 2026-08-01:

| | |
|---|---:|
| fills with `fee < 0` | 5,118,779 |
| total fills | 26,080,657 |
| share negative | **19.6%** |
| largest single rebate | **-$201.21** |

So roughly one fill in five is currently sign-flipped. It doesn't show on the account that triggered this PR (M415566 is all-taker, 271 fills, zero negative) which is likely why it survived, but a maker-heavy book gets a fabricated cost and could be reported as losing when it is profitable. That is the same class of error as the bug being fixed, just pointing the other way.

Fix is to sum signed:
```python
"fee": round(open_fee + (_num(f.get("fee")) or 0.0), 4),
```

Worth noting `senpi-portfolio/scripts/portfolio.py:977` already does this correctly (`total += _num(fl.get("fee")) or 0.0`, no `abs()`), so the two skills currently disagree on sign despite the PR describing an identical schema. A shared helper would stop them drifting again.

shnoodles (commented):
Reviewed against the live fills ledger. The core fix is right and the schema is the correct shape: fees sourced from `userFills`, `gross_*`/`fees`/`net_*`/`fees_status` on every surface, degradation as `null` rather than a fake `$0`, and fees decoupled from the flaky event-log read. The round-trip attribution with pro-rated open-side fees is correct for partial closes, and the `HL fee ALREADY includes the builder fee, never add it on top` note is the exact trap that would have double-counted (I hit it myself measuring this).

Four things before merge. The first is a sign bug that makes fees wrong in the opposite direction, and it's the one I'd block on.

1. `abs()` on the fill fee turns maker rebates into costs (review.py:804, 836)
2. `userFills` is unpaginated, so on high-fill books the fee map is incomplete (review.py:905, portfolio.py:962)
3. All-or-nothing blanking discards the coverage it already computed (review.py:1691)
4. portfolio.py's fee window has no lower bound while gross does (portfolio.py:977)

Detail inline.

senpi-improve-trades/scripts/review.py
```diff
@@ -891,18 +898,31 @@ def fetch_closed_trades(client, wallet, since_ms, until_ms, cap, meta):
             "closed_order_id": _field(p, "closedOrderId", "closed_order_id"),
             "source": "discovery",
         })
+    # FEES: the fills ledger is the ONLY authoritative fee source — discovery_get_trader_history carries no
+    # fee, so the realized_pnl above is GROSS. Pull userFills ONCE and use it both to attribute a round-trip
+    # `fee` to each discovery trade and (when discovery is empty) to reconstruct the closed book. `_hl_info`
+    # fails OPEN → None (distinct from an empty list): None ⇒ fee UNDETERMINED for these trades, never a fake $0.
+    fills = _hl_info({"type": "userFills", "user": wallet}, meta, client)
```
shnoodles: **`userFills` is capped and unpaginated, so `fee_by_oid` can be silently incomplete.**

The HL Info `userFills` endpoint returns only the most recent slice of fills (~2000), and there is no pagination here. `recon` is built from whatever came back, so any discovery trade whose `closed_order_id` predates that slice gets `fee = None`.

Combined with the all-or-nothing rule at line 1691, a single unmatched order id blanks `fees`, `net_realized_pnl` and `net_total_pnl` for that entire wallet. The narrator then falls back to a gross-only headline, which is the exact user-visible symptom this PR exists to remove.

The books most likely to trip it are high-turnover ones over a long window, which are also the books where the gross/net gap matters most. Consider `userFillsByTime` paged across `since_ms`..`until_ms` so coverage matches the window being reported.

senpi-improve-trades/scripts/review.py
```diff
@@ -1638,9 +1683,17 @@ def _strategy_reads(trades, strategies, open_book=None):
...
         open_positions = ob.get("positions") or []
-        realized = agg["pnl"]
-        total = (round(realized + unrealized, 2)
-                 if isinstance(unrealized, (int, float)) and not isinstance(unrealized, bool) else None)
+        gross_realized = agg["pnl"]                  # HL closedPnl sum — GROSS, before trading fees
+        _uok = isinstance(unrealized, (int, float)) and not isinstance(unrealized, bool)
+        # NET is available only when EVERY closed trade's fee resolved. If the fills read was UNDETERMINED for
+        # even one, subtracting a partial fee understates the cost — so fees/net stay UNKNOWN (never gross-as-net).
+        cnt = agg.get("count", 0)
+        fees_ok = (cnt == 0) or (agg.get("fees_known", 0) == cnt)
```
shnoodles, line 1734: **Coverage is computed and then thrown away.**

`fees_known` already carries how many trades resolved, but the emitted record reduces it to `fees_status: "ok" | "undetermined"`. A consumer cannot distinguish 0 of 51 resolved from 50 of 51, so both degrade to the same gross-only narration.

The conservative choice not to subtract a partial fee total is right, but the caller should be able to say "fees resolved for 50 of 51 trades, so net is at least $X" instead of dropping to gross. Suggest carrying it through:

```python
"fees_coverage": {"resolved": agg.get("fees_known", 0), "total": cnt},
```

Same at line 1751 and in `_fees_total_of`. This also makes issue #2 observable in the field rather than silent.

senpi-portfolio/scripts/portfolio.py
```diff
@@ -904,13 +921,79 @@ def hydrate(strat):
...
+        meta.setdefault("warnings", []).append(
+            f"userFills {wallet[:8]} empty/unavailable — fees undetermined (NOT $0)")
+        return None, "undetermined"
+    total = 0.0
+    for fl in fills:
+        if not isinstance(fl, dict):
+            continue
+        t = _ms(fl.get("time"))
+        if until_ms is not None and t is not None and t > until_ms:
+            continue                                # a fill after the last reported close — not this window
```
shnoodles: **Fee window is unbounded below while gross is not.**

The loop skips fills after `until_ms` but applies no lower bound (the docstring says so deliberately). `until_ms` comes from the reported rows, but the gross figure those rows produce is whatever `discovery_get_trader_history` returned, which may be windowed or capped.

When it is, this sums fees for trades whose PnL is not in the gross total, so `net = gross - fees` overstates the cost. On a wallet with a long history and a short reported window the error scales with the history, not the window.

Bounding fees by the same range as the rows (or deriving `since_ms` from `min(closes)`) would keep the two sides symmetric.

Separately, the sign handling here is correct and `review.py:836` is not. Worth reconciling in one direction.

betashop:
All four addressed in `99fbee6` — thanks, these were sharp.

1. **Sign bug** — you're right, `abs()` flipped maker rebates into costs. Now summed **signed** (open + close), so a rebate correctly *adds* to net. Added a signed-rebate test. On the shared-helper point: these are vendored, self-contained packages so they can't import a common module, but review.py and portfolio now agree on sign and both have tests pinning it, so they can't silently drift again.
2. **`userFills` cap** — switched to paged **`userFillsByTime`** over `since−90d .. until` (the lookback captures the OPEN leg of a position that closed in-window), so coverage matches the reported window instead of the recent-2000 slice. Falls back to `userFills` only when there's no window (the standalone/test path).
3. **Coverage thrown away** — now carrying **`fees_coverage {resolved, total}`** on every PnL surface (`strategies[]`, `closed_strategies[]`, `pnl_summary`, `timing_summary`), so a partial read can say "net for R of T" instead of collapsing the whole wallet to gross. Makes #2 observable in the field.
4. **Portfolio fee window** — bounded below by the earliest reported **open** (not close, so both legs stay in range). If opens are unavailable it stays *unbounded* rather than clip at `min(close)` — undercounting fees would overstate net (the exact bug), so over-attribution is the safe direction. Added a pre-window-fill-exclusion test.

148 tests pass. Ready for another look.

senpi-improve-trades/scripts/review.py
```diff
@@ -891,18 +939,31 @@ def fetch_closed_trades(client, wallet, since_ms, until_ms, cap, meta):
             "closed_order_id": _field(p, "closedOrderId", "closed_order_id"),
             "source": "discovery",
         })
+    # FEES: the fills ledger is the ONLY authoritative fee source — discovery_get_trader_history carries no
```
shnoodles, line 942: **`discovery_get_trader_history` does carry a fee.** This comment is the load-bearing assumption for everything below it, and it is not correct.

A live call returns, per closed position:

```json
{"closedOrderId": "501107553651", "coin": "ENA", "realizedPnl": "1225133.921128",
 "openTime": 1768680956, "closeTime": 1784790554, "totalFills": "13594",
 "totalFees": "1035.102496", "totalHyperliquidFees": "929.015645", "totalBuilderFees": "0"}
```

and `sort_by` accepts `TOTAL_FEES`.

I reconciled `totalFees` against the raw fills ledger before raising this. Wallet `0xbd7b48a5…` / CC, 4 fills:

- signed `fee` sum across the fills: `3.950085 + 1.658373 + 1.536760 + 0.696565` = **7.841783**
- the row's `totalFees`: **7.841783**, exact
- `totalFees − totalHyperliquidFees` = `4.127254`, which is precisely the summed `builder_fee` on those fills, so it is builder-inclusive

Across 90 days of closed positions (888,815 rows):

| property | measured |
|---|---|
| populated | 888,801 rows (14 blanks, 0.002%) |
| signed | 36,387 rows carry a negative `totalFees`, so maker rebates survive |

That is the same quantity this function derives via `userFills` → round-trip walk → open-side pro-rating → `closed_order_id` match. Reading it off the row that already supplies the gross would collapse three of the four issues from the last review by construction: there is no page cap to outrun, coverage is 1:1 with the trades being reported, and the fee is inherently aligned with the gross it nets against, so the window-symmetry question does not arise.

Suggested shape: prefer `totalFees` when present, and keep `_fetch_window_fills` + `_reconstruct_closed_from_fills` as the fallback for the discovery-empty path, where it is genuinely the only source, and as a cross-check.

One trap if you take this: **do not use `totalBuilderFees`.** It reads `0` on 93,182 of the 93,280 rows where `totalFees != totalHyperliquidFees + totalBuilderFees`, including the CC example above whose real builder fee is `4.127254`. `totalFees` is the all-in figure; the other two are not a reliable decomposition of it.

senpi-portfolio/scripts/portfolio.py
```diff
@@ -904,13 +921,95 @@ def hydrate(strat):
...
+                             the user's booked profit, the exact bug this guards).
+    FAIL-OPEN: any fee-source error degrades to `undetermined`, it never raises into the portfolio read."""
+    opens, closes = [], []
+    for p in rows:
+        if not isinstance(p, dict):
+            continue
+        for k in ("openTime", "open_time"):
+            ov = _ms(p.get(k)) if p.get(k) is not None else None
+            if ov is not None:
+                opens.append(ov)
```
shnoodles, line 965: **`openTime` is 0-as-missing in real data, and `_ms(0)` returns `0.0`, not `None`.**

So `opens` collects a `0`, `since_ms = min(opens)` becomes epoch, and the `t < since_ms` guard below never fires. The window is unbounded below again, which is the exact state this hunk exists to fix.

It is common, not a corner case. A live `discovery_get_trader_history` row:

```json
{"closedOrderId": "389873481498", "coin": "HYPE", "openTime": 0, "closeTime": 1776714045}
```

Measured over 90 days of closed positions: **35.58% of wallets** (4,591 of 12,905) have at least one epoch-0 open inside their most recent 50 closed positions, which is exactly the `CLOSED_HISTORY_PULL = 50` page this reads. A single zero anywhere in that page collapses the bound for the whole wallet.

This fails in the direction the commit message intends (unbounded below over-attributes fees, understating net), so it is not a regression. But the fix is inert for roughly a third of wallets while reading as applied, and the fallback comment below claims a deliberate choice that is in fact being taken by accident. Guarding the sentinel restores the intent:

```suggestion
            ov = _ms(p.get(k)) if p.get(k) else None      # 0 is a missing-value sentinel, not epoch
            if ov is not None and ov > 0:
                opens.append(ov)
```

Worth noting that PR #537 hit the same sentinel class in trader-research and coerces 0-as-missing there.

senpi-improve-trades/scripts/review.py
```diff
@@ -778,6 +780,44 @@ def _hl_info(payload, meta, client=None, timeout=12):
...
+    slice and its fee silently reads UNKNOWN (blanking net for the whole wallet). When a window is set we page
+    `userFillsByTime` from since−lookback .. until, so COVERAGE MATCHES the window being reported (the lookback
+    captures the OPEN leg of a position that closed inside the window). Ascending pages, deduped on the boundary
+    overlap. Fails OPEN → None (a failed read) — distinct from [] (genuinely no fills). No window → the recent
+    `userFills` slice (the standalone / test path)."""
+    if since_ms is None or until_ms is None:
+        return _hl_info({"type": "userFills", "user": wallet}, meta, client)
+    cursor, end = max(0, int(since_ms) - FILL_LOOKBACK_MS), int(until_ms)
+    out, seen = [], set()
+    for _ in range(FILL_PAGE_CAP):
```
shnoodles, line 795: Minor: **exhausting `FILL_PAGE_CAP` truncates silently.**

If the loop runs all 25 pages without breaking, it returns `out` with no `meta.warning`, so a truncated fill set is indistinguishable from a complete one and the missing fees read as UNKNOWN with no stated cause.

The ceiling is ~50k fills over a `since−90d..until` span, which is further than it sounds but not unreachable: the ENA position in the sample I checked reports `totalFills: 13594` for a **single** closed round-trip, so a handful of positions like that on one wallet clears it.

`fees_coverage` now makes the consequence visible as `resolved < total`, which is the important half. This just makes it attributable. The `for` loop's `else` clause runs only on exhaustion, so it is a clean place to hang it:

```python
    else:
        meta.setdefault("warnings", []).append(
            f"fills paging hit FILL_PAGE_CAP for {wallet[:8]} — fee coverage may be partial")
```

shnoodles (commented):
Re-reviewed at `99fbee6d`. **All four points from the previous review are fixed**, and fixed properly rather than suppressed:

1. **Sign** — `abs()` is gone from both the open lot in `_reconstruct_closed_from_fills` and the round-trip total, summed signed, with the reasoning recorded in the comment. The two skills now agree on sign.
2. **Coverage** — `userFillsByTime` paged over `since−90d..until` replaces the capped, unpaginated `userFills`, so the fill set matches the window being reported and the lookback picks up the open leg of a position that closed in-window.
3. **Observability** — `fees_coverage {resolved, total}` is carried on `pnl_summary`, `_timing_summary`, `_strategy_reads` and `_closed_strategy_rollup`, and documented in `output-shape.md`. 0-of-51 and 50-of-51 are now distinguishable.
4. **Window** — `_window_fees` is bounded below by the earliest reported open, with the fallback deliberately left unbounded so the failure direction over-attributes cost rather than understating it. Right call, and the reasoning is worth keeping in the comment.

Three things came out of verifying the result against production data. One is a design question rather than a defect, and it is the interesting one: **the per-trade fee this PR reconstructs from the fills ledger is already returned by `discovery_get_trader_history`, on the same row as the gross it has to be netted against.** The other two are a sentinel value that makes fix 4 inert for about a third of wallets, and a silent truncation path. All three inline.

betashop:

PR #541, opened 2026-08-15
PnL from MCP: exact realized PnL, no fill reconstruction

shnoodles:
## The problem

`review.py` derived closed-trade PnL by pulling raw Hyperliquid `userFills` and rebuilding round-trips itself: FIFO lot matching, pro-rated open-side fees, `closedPnl` accumulation. Every number that machinery produced is already returned by MCP, per trade, on the row the function was **already parsing**.

`discovery_get_trader_history` returns, per closed position:

```json
{"closedOrderId": "501107553651", "coin": "ENA", "realizedPnl": "1225133.921128",
 "openTime": 1768680956, "closeTime": 1784790554, "totalFills": "13594",
 "totalFees": "1035.102496", "totalHyperliquidFees": "929.015645", "totalBuilderFees": "0"}
```

`totalFees` reconciles exactly against the raw fills ledger. Wallet `0xbd7b48a5…` / CC, 4 fills: signed `fee` sum `3.950085 + 1.658373 + 1.536760 + 0.696565` = **7.841783**, and the row's `totalFees` is **7.841783**. It is signed (36,387 of 888,815 rows over 90 days are negative, so maker rebates survive) and builder-inclusive (`totalFees − totalHyperliquidFees` = the summed `builder_fee`).

## What changed

**Closed trades.** `fetch_closed_trades` reads `discovery_get_trader_history` and nothing else. `realizedPnl` is the exact realized PnL of the round-trip and is emitted unchanged. `totalFees` rides along as a separate `fees` field. Removes `_hl_info` (a `curl` to `api.hyperliquid.xyz`) and `_reconstruct_closed_from_fills`.

**The on-chain fallback is removed, not replaced — its premise does not hold.** It existed because "`strategy_close` clears Senpi's index, so a CLOSED strategy returns empty". History is keyed by wallet address, which outlives the strategy record. Verified live below.

**Unrealized PnL.** `fetch_open_book` is now ONE `discovery_get_trader_state` call per 50 wallets, replacing a threaded per-wallet `strategy_get_clearinghouse_state` fetch with retry/backoff. That tool merges main perps and the HIP-3 `xyz:` dexes into a single `openPositions` list, so an `xyz:` position can no longer be missed — the old path read `main` and `xyz` as separate sections.

**New `account` step, run first.** `account_get_historical_info`, then the full strategy list (ACTIVE + PAUSED + CLOSED). Emits per-period totals only; the raw `pnlHistory` / `accountValueHistory` arrays are deliberately dropped, since they are account-value snapshots where a drop is usually a withdrawal, which guardrail 3 already warns against narrating as trades.

**`openTime: 0` is treated as missing, not epoch 1970.** Discovery sends `0` for "no open time" on a meaningful share of rows.

**Two new HARD RULES** in `SKILL.md`:

> 6. Realized PnL is exact — never adjust it. Do not subtract fees from it, do not recompute it, do not "correct" it. Same for `unrealized_pnl`.
> 7. Do not mention fees unless the user asked about fees. If they did not ask, leave them out entirely: no "fees ate your gains", no fee-adjusted figure, no parenthetical.

## Verified against production

Not just fixtures — the engine was run against a live account with the real MCP token.

**108 strategies, every one CLOSED, and it still collected 26 closed trades from them.** That is the direct disproof of the trap the reconstruction existed to work around.

```
realized (exact) : 17.05
fees (separate)  : 17.23
unrealized       : 0.0
TOTAL            : 17.05
```

If fees had leaked into the headline that would read −0.18, so the separation is demonstrated rather than asserted.

The live run also surfaced three bugs the fixtures could not, fixed in `b7588bb`:

1. `fees` was dropped by the stdout projection (`_TRADE_STDOUT_FIELDS`), so the per-trade fee never reached the narrator even though the aggregate did.
2. A book with **no** current strategies reported `unrealized`/`total` as `null`, leaving a fully-closed book with no TOTAL headline at all (HARD RULE 1). No current strategies means no open positions — a real zero. A current wallet that exists but is unreadable still reports UNKNOWN; that distinction is what the null is for.
3. `_fresh_meta` carried the pre-change `sources` list, so the stepped path disagreed with `run()` about where its data came from.

Fail-open was also observed working for real: one run had `account_get_historical_info` return `UNAVAILABLE`, the warning was recorded, and the review completed with no fabricated zero.

## Tests

`review.py` is 66 lines shorter. The five tests covering deleted code are replaced by tests for the new contract:

- one batched call for N wallets, not N calls; `xyz:` positions counted; 120 wallets chunk to `[50, 50, 20]`
- `realized_pnl` stays `19.00` when `fees` is `16.99`
- `openTime: 0` reads as `None`
- a closed-only book gets a real `0.0` unrealized, while all-unreadable current wallets stay `None`
- `account_history` emits period totals and no raw series

75 improve-trades tests pass; 563 across the CI set, plus both `min_budget` parity scripts.

## Out of scope, flagged

CI never ran `senpi-improve-trades/tests` — the workflow gates ops, discover, trading-runtime, trader-research, market-pulse and portfolio only, so this skill's suite has only ever run locally. Since this PR rewrites its PnL path, `315cd11` adds it to the gate. `account-status`, `smart-money` and `strategy-author` still collide on test module names when collected alongside the others; that is pre-existing and left for a separate fix.

## Relationship to #538

This **supersedes** #538 rather than building on it. That PR spends ~200 lines paging `userFillsByTime` to compute a number this branch reads off the row it was already parsing, and the two rewrite the same code, so they will conflict. #538's sign fix, coverage plumbing and window bound are all correct work — they are simply unnecessary once the fee comes from `totalFees`. Closing #538 in favour of this is my suggestion, but it is your call.

One trap worth carrying forward from that review: **do not use `totalBuilderFees`.** It reads `0` on 93,182 of the 93,280 rows where `totalFees != totalHyperliquidFees + totalBuilderFees`. `totalFees` is the all-in figure; the other two are not a reliable decomposition.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,5 +27,8 @@ jobs:
 
       # pytest collects every test regardless of a __main__ runner — bare `python3 <file>` exits 0
       # without running anything on a runner-less file (test_default_catalog.py), a silent green.
+      # improve-trades was never gated here, so its suite only ran locally. Added with this PR, which
+      # rewrites its PnL path. (account-status / smart-money / strategy-author still collide on module
+      # names when collected alongside the others — a separate fix.)
       - name: full suites (ops + discover + trading-runtime)
-        run: python3 -m pytest senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-trading-runtime/tests senpi-trader-research/tests senpi-market-pulse/tests senpi-portfolio/tests -q
+        run: python3 -m pytest senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-trading-runtime/tests senpi-trader-research/tests senpi-market-pulse/tests senpi-portfolio/tests senpi-improve-trades/tests -q
--- a/senpi-improve-trades/SKILL.md
+++ b/senpi-improve-trades/SKILL.md
@@ -6,8 +6,9 @@ description: >-
   to the best whales", "how could I make more gains", "suggest improvements", "review my trades", "am I
   getting shaken out too early / how are my exits firing", "what did my own limits block / what couldn't
   I take", "where am I leaking", "walk me through / explain my [asset] trade", "what am I paying in fees /
-  maker vs taker", "why is [strategy] losing". When the user has NOTHING to review yet, `meta.book_state` routes it: nothing deployed -> read the market (senpi-market-pulse) then shortlist a fit (senpi-strategy-discover); deployed-but-idle -> diagnose THAT strategy, never pitch another. A hidden engine (scripts/review.py) reconstructs every
-  CLOSED trade from discovery, enriches each exit reason + blocked signals from the runtime telemetry
+  maker vs taker", "why is [strategy] losing". When the user has NOTHING to review yet, `meta.book_state` routes it: nothing deployed -> read the market (senpi-market-pulse) then shortlist a fit (senpi-strategy-discover); deployed-but-idle -> diagnose THAT strategy, never pitch another. A hidden engine (scripts/review.py) reads the
+  account's PnL and every CLOSED trade's exact realized PnL from MCP, enriches each exit reason + blocked
+  signals from the runtime telemetry
   event log, computes the honest "if I'd held to now" counterfactual, and crosses the book against what
   the market did — you narrate it under strict guardrails: process over outcome (lead with the aggregate,
   not the one reversal), it's the STRATEGY not the user, NO fabricated "+$X/week", no performance-chasing,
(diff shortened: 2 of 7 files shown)


betashop (commented):
The engine here is the right call — reading `realizedPnl` + `totalFees` straight off the row (no `userFills` reconstruction), the live-verified fallback removal (108 closed → 26 trades), and the batched `trader_state` with main+xyz merged are all clean wins. Agreed on superseding #538, and thanks for the `totalBuilderFees` trap.

One product change before merge: **I don't want the two absolute rules.** Give the model the full picture and let it use judgment.

**Rule 7 (don't mention fees unless asked) — drop it.** improve-trades is a *coaching* skill, and your own live example (realized `17.05`, fees `17.23` → net `−0.18`) is exactly the case where staying silent on fees misleads the user. The model should be free to surface material fee drag even unprompted.

**Report TOTAL PnL net of fees.** Keep `realized_pnl` exact and shown as its own component (Rule 6 is fine there — don't re-derive the exchange's number), but the TOTAL the user judges on should be `realized − fees + unrealized`. Fees belong *in* the headline, not hidden beside it.

**Add a fee playbook — the model should be able to answer all of these:**
- *how much am I paying* → `fees` / `fees_total`
- *maker vs taker* → `execution_quality` (already computed — keep it exposed)
- *how do I pay less* → **earn senpi points** (a higher loyalty tier lowers your fee rate) — hand to `senpi-account-status` for the tier/points detail.

So: your engine + fees back in the total + a fee-education layer, minus the hard "never." Everything else is good to go.

PR #542, opened 2026-08-17
senpi-portfolio: PR #481 review follow-ups — zero-price guard, spot bucket, live-shape fixture, loud reconcile

YosephKS:
## Summary

Follow-ups from @0xsarvesh's [review on #481](https://github.com/Senpi-ai/senpi-skills/pull/481#issuecomment-5296550846) (merged) — the blocking nit inside the merged lines plus the three pre-existing issues flagged for a separate PR. All in `senpi-portfolio`.

**1. `tokenPriceInUSD: 0` re-opened the bug #481 fixed (the review's blocking nit).** `_f` returns the first *present* numeric key, so a present-and-zero price made `raw * 0 == 0` and the balance invisible again. Guarded with the suggested `or 1.0` — zero-as-missing is a known sentinel shape on this API (#537 coerces the same `0` sentinels to `None`).

**2. `idle_total` omitted the HL spot USDC bucket.** `spot_usd` was read and never added, under-reporting embedded idle by exactly the spot balance — quietly enough on the review's live payload ($1.81) to hide under the reconciliation tolerance and flip `reconciles` to a false `True`. `idle_total` is now perps + spot + EVM, matching the funding waterfall.

**3. The canonical fixture encoded a payload shape the API does not return.** `tests/fixtures/portfolio_fixture.json` used `{symbol, chain, usdValue}`; live GetPortfolioV3 carries `{tokenSymbol, chainId, balanceInUSD, formattedBalance, tokenPriceInUSD}`. That's why the evm_usdc bug shipped green — the suite proved the parser reads the fixture, not the API. Rewritten to the live shape, with a guard test (`test_canonical_fixture_token_balances_use_the_live_shape`) that fails if the invented fields ever come back.

**4. `totals.reconciles == false` emitted no `meta.warnings` entry** — the only silent failure mode in the engine (13 other append sites), so SKILL.md's "STOP and re-run on `reconciles: false`" rule never fired for a reader scanning warnings. Both totals builders (`compute()` for `all`/`positions` and `_money_totals()` for the `money` step) now append a `TOTALS DO NOT RECONCILE` warning quoting both figures, the gap, and the tolerance.

SKILL.md updated to match: the idle-in-embedded bucket definition now names all three legs, and the `reconciles` degraded-mode entry points at the new warning.

## Tests

Five regression tests added, one per finding plus the quiet side:
- `test_embedded_evm_usdc_survives_a_present_and_zero_price` — the review's probe row that failed pre-fix (`formattedBalance` present, `tokenPriceInUSD: 0` → was `[]`, now $93.95)
- `test_embedded_idle_counts_the_hl_spot_bucket` — the review's live numbers: 135.21 + 1.81 + 93.95 = 230.97
- `test_canonical_fixture_token_balances_use_the_live_shape` — durable guard on the fixture shape
- `test_reconcile_mismatch_is_a_loud_warning_not_a_silent_flag` — both builders warn, gap quoted
- `test_reconciled_totals_stay_quiet` — no warning when totals tie out

Full suite: **67/67 passing** (plus `test_name_reader_parity.py`).

https://claude.ai/code/session_01JNoB3R6UHLT9N32sZ27SeK

---
_Generated by [Claude Code](https://claude.ai/code/session_01JNoB3R6UHLT9N32sZ27SeK)_

### DIFF
--- a/senpi-portfolio/scripts/portfolio.py
+++ b/senpi-portfolio/scripts/portfolio.py
@@ -1127,6 +1127,13 @@ def enrich_market(client, strategies, meta):
 
 
 # ──────────────────────────────────────────────────────────────── taxonomy + signals
+def _reconcile_tolerance(grand_total):
+    """THE reconciliation tolerance — the single definition behind the `reconciles` flag in BOTH totals
+    builders (compute / _money_totals) and the warning that quotes it. Inlining the formula at each site
+    is how the warning ends up quoting a stale tolerance while the flag uses the real one."""
+    return max(2.0, 0.01 * (grand_total or 0.0))
+
+
 def _warn_if_not_reconciled(totals, meta):
     """A `reconciles: false` must land in meta.warnings like every other failure mode in this file —
     a reader scanning warnings and seeing [] next to a silent mismatch never triggers SKILL.md's
@@ -1137,7 +1144,7 @@ def _warn_if_not_reconciled(totals, meta):
     pbal = totals.get("portfolio_total_balance_usd")
     grand = totals.get("grand_total_usd") or 0.0
     gap = round(abs((pbal or 0.0) - grand), 2)
-    tol = round(max(2.0, 0.01 * grand), 2)
+    tol = round(_reconcile_tolerance(grand), 2)
     meta.setdefault("warnings", []).append(
         f"TOTALS DO NOT RECONCILE — the portfolio aggregate (${pbal:,.2f}) and the per-wallet grand "
         f"total (${grand:,.2f}) disagree by ${gap:,.2f} (tolerance ${tol:,.2f}): a bucket is missing or "
@@ -1183,7 +1190,7 @@ def compute(embedded, strategies, portfolio_totals, meta=None):
     }
     # reconciliation flag — surfaces silent drift between the two sources
     pbal = portfolio_totals.get("total_balance_usd")
-    totals["reconciles"] = (pbal is None) or (abs(pbal - grand_total) <= max(2.0, 0.01 * grand_total))
+    totals["reconciles"] = (pbal is None) or (abs(pbal - grand_total) <= _reconcile_tolerance(grand_total))
     _warn_if_not_reconciled(totals, meta)
 
     net = round(gross_long - gross_short, 2)
@@ -1525,7 +1532,7 @@ def _money_totals(embedded, strategies, portfolio_totals, meta=None):
         "portfolio_total_withdrawable": portfolio_totals.get("total_withdrawable"),
     }
     pbal = portfolio_totals.get("total_balance_usd")
-    totals["reconciles"] = (pbal is None) or (abs(pbal - grand_total) <= max(2.0, 0.01 * grand_total))
+    totals["reconciles"] = (pbal is None) or (abs(pbal - grand_total) <= _reconcile_tolerance(grand_total))
     _warn_if_not_reconciled(totals, meta)
     return totals
 
--- a/senpi-portfolio/scripts/portfolio.py
+++ b/senpi-portfolio/scripts/portfolio.py
@@ -644,6 +644,8 @@ def fetch_embedded(client, meta):
             # Live GetPortfolioV3 uses `balanceInUSD` (+ `formattedBalance`/`tokenPriceInUSD`) —
             # without it every token read $0 and evm_usdc was always [].
             amt = _f(tb, "usdValue", "usd_value", "amountUsd", "balanceUsd", "balanceInUSD", "amount", default=0.0)
+            # `== 0.0` (not a missing-key test) is DELIBERATE: a present-and-zero amount is this API's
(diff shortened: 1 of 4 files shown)


senpi-portfolio/tests/fixtures/portfolio_fixture.json
```diff
@@ -9,7 +9,7 @@
     "total_withdrawable": 2461.98,
     "total_in_hyperliquid": 0,
     "total_spot_usd_in_hyperliquid": 0,
-    "token_balances": [{"symbol": "USDC", "chain": "Base", "usdValue": 1.51}]
+    "token_balances": [{"tokenSymbol": "USDC", "tokenAddress": "0x833589fcd6edb6e8f4c7c32d4f71b54bda02913", "formattedBalance": "1.510231", "balanceInUSD": 1.5099873, "tokenPriceInUSD": 0.9998, "chainId": 8453}]
```
im-vignesh: `tokenAddress` here is 39 hex chars — it drops the `0` from `E08f` in the real Base USDC address, so it's not a valid EVM address. The parser never reads this field, so nothing breaks, but since this PR's point 3 is "the fixture must mirror the live shape," let's use the real address:

```suggestion
    "token_balances": [{"tokenSymbol": "USDC", "tokenAddress": "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913", "formattedBalance": "1.510231", "balanceInUSD": 1.5099873, "tokenPriceInUSD": 0.9998, "chainId": 8453}]
```

senpi-portfolio/scripts/portfolio.py
```diff
@@ -1122,7 +1127,25 @@ def one(item):
...
+    """A `reconciles: false` must land in meta.warnings like every other failure mode in this file —
+    a reader scanning warnings and seeing [] next to a silent mismatch never triggers SKILL.md's
+    "STOP and re-run on reconciles: false" rule. Names the two figures and the gap so the narrator
+    can quote what disagreed."""
+    if meta is None or totals.get("reconciles") is not False:
+        return
+    pbal = totals.get("portfolio_total_balance_usd")
+    grand = totals.get("grand_total_usd") or 0.0
+    gap = round(abs((pbal or 0.0) - grand), 2)
+    tol = round(max(2.0, 0.01 * grand), 2)
```
im-vignesh: This tolerance formula now lives in three places: here, `compute()` (line 1186), and `_money_totals()` (line 1528). If either builder's check ever changes, this warning will quote a stale tolerance while `reconciles` uses the real one — a confusing mismatch in exactly the message we're telling readers to trust verbatim. Consider extracting a `_reconcile_tolerance(grand)` helper (or passing `tol` into this function) so they can't drift.

senpi-portfolio/scripts/portfolio.py
```diff
@@ -646,13 +646,18 @@ def fetch_embedded(client, meta):
             amt = _f(tb, "usdValue", "usd_value", "amountUsd", "balanceUsd", "balanceInUSD", "amount", default=0.0)
             if amt == 0.0:
```
im-vignesh, line 649: Non-blocking, pre-existing but now load-bearing twice: this line treats a *genuinely zero* `balanceInUSD` as "missing" and falls into the price-fallback path. Harmless today (raw would also be ~0 and `if amt:` filters it out), but with the new `or 1.0` below, "0 means missing" is now a stacked convention. Worth a one-line comment here citing the same #537 sentinel rationale so nobody "fixes" this to `is None` later.

im-vignesh (changes requested):

senpi-portfolio/tests/fixtures/portfolio_fixture.json
YosephKS: Fixed in 801ee17 — corrected to the real Base USDC address. The same truncated form had been copied into the `test_embedded_evm_usdc_reads_balanceInUSD` regression test (it originated in #481's test), so that occurrence is fixed in the same commit.

---
_Generated by [Claude Code](https://claude.ai/code)_

senpi-portfolio/scripts/portfolio.py
YosephKS: Done in cadd531 — extracted `_reconcile_tolerance(grand_total)` and pointed all three sites at it (`compute()`, `_money_totals()`, and this warning), so the flag and the quoted tolerance can no longer drift.

---
_Generated by [Claude Code](https://claude.ai/code)_

senpi-portfolio/scripts/portfolio.py
YosephKS: Added in cf7decc — a one-line comment on the `amt == 0.0` check pinning the #537 zero-as-missing sentinel rationale, so it doesn't get "corrected" to a missing-key test later.

---
_Generated by [Claude Code](https://claude.ai/code)_

im-vignesh (approved):

PR #373, opened 2026-06-23
New skill: senpi-market-pulse — "what's happening in the markets today?"

betashop:
## What

A new guide/analysis skill, **`senpi-market-pulse`**, that answers *"what's happening in the markets today?"* with thoughtful, structured, cross-asset analysis — the read a human couldn't assemble alone, not just "BTC is up."

Built on **Sarvesh's `senpi-strategy-discover` pattern** (`feat/strategy-discover-engine`): a hidden deterministic engine does the streamlined, resilient data-gathering + computes the cross-asset signals; the LLM (SKILL.md) does the analysis, narration, catalyst layer, and the closing CTAs.

## Why

Today, unguided, the agent answers this badly — it wanders across dozens of tool calls, leaves out whole asset classes (answered crypto-only until pushed, twice), leads with a single coin, stalls on infra hiccups (Hyperfeed down, XYZ candles 500ing, needs `dex="xyz"`), and stops at "BTC is down 3%." This skill makes the gold-standard answer the **first** answer.

## How it works

- **Engine (`scripts/pulse.py`)** — one health-check + parallel multi-class pull (crypto + XYZ equities + indices + commodities + macro). Daily moves come from `prevDayPx` on the instruments list, so a candle 500 / Hyperfeed outage **never drops an asset class**. Capped parallel deep-pull of the biggest movers for volume/funding conviction. Health-gated `leaderboard_*` smart-money layer (degrades to `null` cleanly). Deterministic signals: **dispersion** (index vs. components), the **gold/DXY/VIX confirmation checklist**, day classification. Fails open end-to-end.
- **`scripts/_mcp.py`** — self-contained MCP transport, reused from `senpi-strategy-discover`.
- **`SKILL.md`** — top-down output contract, the analysis framework, mandatory all-class coverage, and the two mandatory CTAs.
- **`references/analysis-framework.md`** — the cross-asset reasoning (dispersion vs capitulation, the gradient, confirmation checklist, K-shape, volume/funding).
- **`references/scope.md`** — v1 / v1.1 / v2 build plan.
- **`tests/`** — offline fixture test, **7/7 passing**, no network.

## Mandatory CTAs (per Jason)

Every market read ends with, verbatim:
> 1. Want me to check how our strategies and positions are positioned in this?
> 2. Want me to create a new strategy catered to this market?

CTA 1 → positions read. CTA 2 → hands **senpi-strategy-author** a brief pre-built from the thesis (proposes, never auto-builds).

## Review notes

[...]

PR #546, opened 2026-08-18
docs: route live-strategy edits through senpi update, not close-and-redeploy

0xsarvesh:
`senpi update` shipped in `@senpi-ai/runtime` ([senpi-trading-runtime#342](https://github.com/Senpi-ai/senpi-trading-runtime/pull/342)): it swaps a running strategy onto a new recipe **in place**, keeping DSL position files, scanner stores and action history. These skills predate it and still teach the only path that existed before — close the strategy and redeploy on a fresh wallet.

## Why this is worth doing now

`references/editing-a-live-strategy.md` opened with:

> **There is no in-place scanner reload, and no single verb for this today.** … Applying an edit means **closing the strategy and redeploying it on a fresh wallet** — which market-exits its open positions and returns the funds to main.

An agent following that today **flattens a live book** to apply a re-tune that no longer requires closing anything. That costs the user spread, fees and their open P&L, for nothing.

## What changed

**`senpi-strategy-ops`** — SKILL.md and `references/editing-a-live-strategy.md` now lead with `senpi update`:

```bash
openclaw senpi validate <instance-dir>                          # writes the proof --apply needs
openclaw senpi update <instance-dir> --id <runtime_id>          # PLAN — show this to the user
openclaw senpi update <instance-dir> --id <runtime_id> --apply
```

Close-and-redeploy is kept, scoped to the three changes `senpi update` refuses outright, each of which genuinely needs a new deployment:

| Change | Why it can't be in-place |
|---|---|
| different `strategy.wallet` | A different deployment; the old wallet's positions would be left unmanaged. |
| renamed / reordered scanner | External scanner state is keyed by name **and** position. Appending at the end is fine. |
| changed `action_type` | The new class would inherit the old one's execution history. |

All the money and consent guidance is preserved for that path — it is still a market exit and still needs explicit consent in those words.

**Both surfaces now lead the plan with forward-only exits.** A tightened `dsl_preset` governs new entries and never reaches a position already open. That is the thing a user will otherwise mishear, so the agent is told to say it out loud: never let *"I made it tighter"* be heard as *"my open trades are now tighter."*

**`senpi-strategy-author`** — the handoff sentence. Authoring still does not apply edits (it hands to ops), so the boundary is unchanged; it now names `senpi update` as what ops will do, and carries the two facts the author must say out loud when handing over — forward-only exits, and the three changes that still force a market exit.

**`senpi-trading-runtime`** — same routing added.

All three skills previously said some version of *"there is no single verb for this today."* Fixing two and leaving the third would have made the catalogue contradict itself, so all three move together.

## On the line budget

`senpi-strategy-ops/SKILL.md` sits exactly at its 300-line budget and `senpi-strategy-author/SKILL.md` at its 368. Both edits are **length-neutral** — the depth went into the reference, which is pay-per-read.

`test_skill_surface`'s docstring predicted this task almost exactly — *"every task in this workstream had a reason to add just three more lines resident"* — so the budget was held rather than raised.

## Testing

354 ops tests pass. They must be run from the repo root: from inside the skill directory `test_pkg` fails on a relative `strategies` path, with or without this change.

https://claude.ai/code/session_012bkDp51mWSPrk6jfaNKkUt

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -354,7 +354,7 @@ loop every time:
 > wallet, no market exit. **Re-running `create` will NOT apply it** — the deploy verb is idempotent, so it
 > adopts the existing wallet and leaves the deployed scanner as it is. Tell the user two things: exit
 > changes are **forward-only**, so a tightened `dsl_preset` governs new entries and never reaches a
-> position already open; and a changed `strategy.wallet`, a renamed/reordered scanner or a changed
+> position already open; and a changed `strategy.wallet`, a renamed or moved scanner or a changed
 > `action_type` still forces a close-and-redeploy — a market exit. Below is the not-yet-live path.
 
 1. **Confirm with the user** — budget + "ready to deploy?" Funding a wallet is real money and one-way, so
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -349,13 +349,13 @@ only once **`senpi-strategy-ops` deploys it AND that deploy's report says `overa
 loop every time:
 
 > **Was this an edit to a strategy that is ALREADY LIVE?** (you changed the scoring / scanner / DSL of a
-> deployed package — "make my live strategy more aggressive", re-tune, re-score) — then say so before you
-> do anything. **There is no single verb for this today, and re-running `create` will NOT apply your
-> edit**: the deploy verb is idempotent, so it adopts the wallet that already exists and leaves the
-> deployed scanner as it is. Applying an edit to a live strategy means **closing it and redeploying** —
-> `close.py <id>` (which flattens its open positions and returns the funds) and then the loop below on a
-> fresh wallet. That is real money and a market exit, so **confirm it with the user in those words
-> first**; never present it as a re-tune. The steps below are for a strategy that is not yet live.
+> deployed package — "make my live strategy more aggressive", re-tune, re-score) — then hand it to
+> **`senpi-strategy-ops`**, which applies it IN PLACE with `openclaw senpi update`: no close, no fresh
+> wallet, no market exit. **Re-running `create` will NOT apply it** — the deploy verb is idempotent, so it
+> adopts the existing wallet and leaves the deployed scanner as it is. Tell the user two things: exit
+> changes are **forward-only**, so a tightened `dsl_preset` governs new entries and never reaches a
+> position already open; and a changed `strategy.wallet`, a renamed/reordered scanner or a changed
+> `action_type` still forces a close-and-redeploy — a market exit. Below is the not-yet-live path.
 
 1. **Confirm with the user** — budget + "ready to deploy?" Funding a wallet is real money and one-way, so
    this is an explicit yes, not an assumption.
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -303,19 +303,19 @@ is strategy-driven: close also cleans up an attributed package's **orphaned** (n
 
 ## Applying an edit to a strategy that is already LIVE
 
-"Make my live strategy more aggressive." **The edit is authored in `senpi-strategy-author`**, never
-here. And **re-running `create` will NOT apply it**: the deploy verb is idempotent, so it adopts the
-existing wallet and leaves the deployed scanner as it is. Applying an edit means **closing the strategy
-and redeploying on a fresh wallet** — a market exit of every open position, funds back to main, and a
-custom ratchet/stop ladder that does **not** carry over. **Get explicit consent in those words before
-you close anything**; never present it as a re-tune.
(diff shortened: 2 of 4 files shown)


dmcr (approved):

shnoodles (approved):

betashop (commented):
Review notes from the churn analysis. The routing itself is right and matches what the report showed users need — the forward-only ladder caveat is the exact thing to say out loud. Four things worth a look before this lands:

**1. Ship order — the verb must exist first.** This text routes agents to `openclaw senpi update`, which reaches a box only with the runtime release that carries senpi-trading-runtime #342 (+ #373; #397 is that stack merged onto today's `main` with the conflict resolved). On a pre-verb box Commander's root `.action()` swallows the subcommand: the agent sees `error: too many arguments for 'senpi'`, not "unknown command", and the natural fallback is delete-and-redeploy — the failure this PR exists to end. Either merge after the runtime release, or add one line here: *if the verb is missing, STOP — the runtime needs updating; never close-and-redeploy to apply an edit.* (#573 adds `deploy.py update`, which detects the pre-verb case and refuses exactly that way.)

**2. No version bumps.** `SKILL.md` changes in senpi-strategy-author, senpi-strategy-ops and senpi-trading-runtime with the versions unchanged (3.2.1 / 3.7.0 / 4.0.1), so the fleet's auto-upgrade would not pick the new text up and the README table goes stale. CLAUDE.md rule: skill change → MINOR (runtime skill → PATCH), README in the same commit. #572 bumps author → 3.3.0; #573 (stacked on this branch) bumps ops → 3.8.0 and trading-runtime → 4.0.2 — so either bump here or let those carry it, but this PR alone leaves the fleet on the old routing.

**3. `--id`.** The ops text shows `senpi update <instance-dir> --id <runtime_id>`; the trading-runtime text shows `openclaw senpi update ./pkg` with `--id` "required when more than one is installed". Every catalog package is multi-instance and most boxes run several runtimes, so `--id` is effectively always required — worth showing it in the runtime SKILL example too.

**4. After `--apply`.** One sentence would close the loop: re-read `openclaw senpi runtime list --json` (the descriptor's `dslPreset` / `description`) to confirm what is running now, re-send the How-it-runs block, and name the open positions that keep the old ladder (the plan lists them). #573's `status.py` prints "✎ running recipe ≠ disk" for exactly this state, and senpi-trading-runtime #399 publishes a `recipeHash` in the descriptor so that check stops depending on the preset name.

Checked against the runtime branch: the refused-changes table matches `update-handler.ts` (wallet / external scanner renamed-moved-inserted-removed / `action_type`), the exit-code semantics (2 = never touched, 1 = may not be where you left it) match `src/cli/update-command.ts`, and the forward-only wording matches the guide text. Nothing else to add on the diff.

shnoodles:
**Review: yes, the routing is right — and this is the most order-sensitive PR in the set.**

Replacing "close it and redeploy on a fresh wallet" with `openclaw senpi update` is the correct answer to the incident's biggest single line item (seven delete-and-redeploy cycles in five days). Demoting close-and-redeploy to its three genuine triggers — changed `strategy.wallet`, renamed/moved external scanner, changed `action_type` — is the right scope, and keeping "explicit consent in those words" on that path is the right guard.

**The one change to make: gate it on a runtime version.**

`main` has **zero** occurrences of `openclaw senpi update` in either SKILL.md, so this PR introduces the verb to the skills — and the verb itself ships in runtime #398/#397, which is not released. Meanwhile:

```yaml
requires:
  - senpi-trading-runtime        # no version constraint
```

So nothing stops this skill loading on a box whose runtime predates the verb. On that box the agent runs `openclaw senpi update`, gets an unknown command, and the fallback it used to have has just been rewritten into three narrow triggers — so the most likely improvisation is close-and-redeploy, which is a market exit of a live book, arrived at by accident rather than by the consent path this PR carefully preserved.

The Slack fix-list sequences this by hand ("land #397 → main → release → fleet release → then skills #546"). That works exactly once, and it is the same human-memory dependency that lets runtime #398 merge alone and silently `await` a `void`. Two options, either is fine:

- pin a minimum in `requires:` (`senpi-trading-runtime >= <version that carries the verb>`); or
- have the skill probe once (`openclaw senpi guide` / `senpi update --help`) and, if the verb is absent, say so and stop rather than improvise.

I would take the pin — it is one line and it makes the ordering a property of the artifact instead of a property of the release runbook.

**Merge order note:** #549 is based on `feat/senpi-update-routing`, i.e. this branch — not on main. So #546 must land first; GitHub will retarget #549 afterwards. Worth knowing because the Slack list names them in one breath. #549 also corrects this PR's "exit changes are forward-only" to the accurate "`dsl_preset` is forward-only" — so if #546 ships alone for any length of time, it ships with that overclaim live.

Ponytail pass: `+96/-26`, and it is mostly deletion-by-replacement of a longer wrong procedure with a shorter right one. Nothing to cut. `net: 0 lines`.

PR #548, opened 2026-08-24
fix(stingray): prefix XYZ (HIP-3) assets with xyz: for market reads + emits

betashop:
## The bug

Stingray derives its universe from `leaderboard_get_markets` each tick (no hardcoded whitelist — that part is fine). But the board returns **bare** tokens for XYZ (HIP-3) assets — `NVDA`, `BRENTOIL`, `SILVER` — with the DEX in a separate `dex` field. The scanner then:

1. called `market_get_asset_data` for the 4h price-confirmation gate, and
2. emitted the trade

**both under the bare token**, because `_dex_for()` only recognizes XYZ when the `xyz:` prefix is *already* present. So every XYZ asset was addressed as a main-DEX coin that doesn't exist there — the read failed and the leg was silently skipped. The scanner degrades gracefully, but the **catalog validator caught it**, and no XYZ rotation leg could ever actually trade (defeating half of a cross-asset "long semis, short crypto" strategy).

## The fix

The board row *already* carries the correct `dex`, so it just wasn't being used to build the name:

- **`_venue_asset(token, dex)`** — re-attaches `xyz:` for HIP-3 assets (idempotent; main-DEX passes through). Used for **both** the `market_get_asset_data` read and the emitted `asset`, so the price gate and the trade agree on the venue-correct name.
- **`_bare(name)`** — strips `xyz:` for the held / recent-signal dedup keys, so a held XYZ position still de-dupes against the board's bare token (and vice-versa).

This mirrors the sibling **Starling** scanner, which already does bare-for-dedup + prefixed-for-emit.

## Verified

- New `strategies/stingray/tests/test_engine.py` (3 tests): `_venue_asset`/`_bare` units + a `scan()` smoke run proving an XYZ board asset emits **`xyz:NVDA`** (never bare) and is price-gated under **`xyz:NVDA` / dex=`xyz`**. All pass.
- `validate_strategy strategies/stingray` → **✓ stingray v1.0.1**.
- stingray `1.0.0 → 1.0.1` (patch — pure bugfix; catalog stores no per-strategy version, so no regen needed).

Context: this landed as a local fix on a live deploy; this PR fixes the **upstream catalog package** so every future Stingray deploy is correct.

### DIFF
--- a/strategies/stingray/main/scanners/scan.py
+++ b/strategies/stingray/main/scanners/scan.py
@@ -70,11 +70,14 @@ def _venue_asset(token, dex):
     assets — equities/commodities/indices — are addressed as `xyz:<TOKEN>`, but `leaderboard_get_markets`
     returns the BARE token (`NVDA`, `BRENTOIL`, `SILVER`) with the DEX in a separate `dex` field. Re-attach
     the prefix here: without it, market_get_asset_data (and the trade itself) hit an asset that isn't on the
-    main DEX and fail — the exact issue the catalog validator caught. Main-DEX tokens pass through."""
+    main DEX and fail — the exact issue the catalog validator caught. Idempotent AND case-canonical: an
+    already-prefixed token is rebuilt with a lowercase `xyz:` (the board upper-cases tokens, so a future
+    prefixed row would arrive as `XYZ:…`, which is non-canonical and would fail the same way a bare name
+    does). Main-DEX tokens pass through unchanged."""
     t = str(token)
-    if str(dex).lower() == "xyz" and not t.lower().startswith("xyz:"):
-        return f"xyz:{t}"
-    return t
+    if t.lower().startswith("xyz:"):
+        return "xyz:" + t[4:]                    # canonical lowercase prefix, whatever case it arrived in
+    return f"xyz:{t}" if str(dex).lower() == "xyz" else t
 
 
 def _bare(name):
--- a/strategies/stingray/main/scanners/scan.py
+++ b/strategies/stingray/main/scanners/scan.py
@@ -65,6 +65,25 @@ def _dex_for(asset):
     return "xyz" if str(asset).lower().startswith("xyz:") else ""
 
 
+def _venue_asset(token, dex):
+    """The venue-correct asset name for BOTH the market read and the emitted trade. XYZ (HIP-3)
+    assets — equities/commodities/indices — are addressed as `xyz:<TOKEN>`, but `leaderboard_get_markets`
+    returns the BARE token (`NVDA`, `BRENTOIL`, `SILVER`) with the DEX in a separate `dex` field. Re-attach
+    the prefix here: without it, market_get_asset_data (and the trade itself) hit an asset that isn't on the
+    main DEX and fail — the exact issue the catalog validator caught. Main-DEX tokens pass through."""
+    t = str(token)
+    if str(dex).lower() == "xyz" and not t.lower().startswith("xyz:"):
+        return f"xyz:{t}"
+    return t
+
+
+def _bare(name):
+    """Bare, upper key for held / recent-signal dedup — strips an `xyz:` prefix so a held or
+    recently-signaled XYZ position matches the leaderboard's bare token (and vice-versa)."""
+    s = str(name)
+    return (s[4:] if s.lower().startswith("xyz:") else s).upper()
+
+
 # ── ACCOUNT + HELD ASSETS (bison _get_account, verbatim shape incl. read-sanity guard) ──
 
 def _get_account(ctx):
@@ -289,26 +308,27 @@ def _emit_side(ctx, ranked, direction, cap, held_set, signaled, ttl, now,
     for c in ranked:
         if len(emits) >= cap:
             break
-        coin = c["token"]
-        cu = coin.upper()
+        cu = _bare(c["token"])                     # bare, upper — held + recent-signal dedup key
         if cu in held_set:
             continue
-        if _recently_signaled(signaled, coin, ttl, now):
+        if _recently_signaled(signaled, cu, ttl, now):
             continue
-        ok = _price_ok(ctx, coin, direction)
+        # venue-correct name (xyz:<TOKEN> for HIP-3 assets) — required by market_get_asset_data AND the emit
+        venue = _venue_asset(c["token"], c.get("dex"))
+        ok = _price_ok(ctx, venue, direction)
         if ok is None:
-            print(f"[stingray.scan] SKIP {coin} {direction}: trend read failed (don't size blind)",
+            print(f"[stingray.scan] SKIP {venue} {direction}: trend read failed (don't size blind)",
                   file=sys.stderr)
             continue
         if not ok:
-            print(f"[stingray.scan] SKIP {coin} {direction}: price gate (hard "
+            print(f"[stingray.scan] SKIP {venue} {direction}: price gate (hard "
                   f"{'down' if direction == 'LONG' else 'up'}trend, don't fight price)",
                   file=sys.stderr)
             continue
         margin_pct = scoring.margin_pct_for(abs(c["net_tilt"]), base_margin, max_margin)
         signaled[cu] = now
         emits.append({
-            "asset": coin,
+            "asset": venue,
             "direction": direction,
             "marginPct": margin_pct,          # PERCENT in (0,100] — runtime sizes the dollars
             "leverage": leverage,             # clamped [1,5] + venue max
@@ -358,7 +378,7 @@ def scan(inputs, ctx):
                 print(f"[stingray.scan] WARNING: state append failed: {exc!r}", file=sys.stderr)
         return []
     held_assets = [p["coin"] for p in positions if p.get("coin")]
-    held_set = {h.upper() for h in held_assets}
+    held_set = {_bare(h) for h in held_assets}     # strip xyz: so held HIP-3 positions dedup vs bare board tokens
 
     signaled = _prune_signaled(_load_signaled(ctx), ttl, now)
 
--- a/strategies/stingray/tests/test_engine.py
+++ b/strategies/stingray/tests/test_engine.py
@@ -1,0 +1,94 @@
+"""Stingray scanner tests. Pure helpers + a scan() smoke run against a fake ctx (no network).
+Run: python3 -m pytest strategies/stingray/tests -q  (or plain `python3 .../test_engine.py`)."""
+import os
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "main", "scanners"))
+import scan     # noqa: E402
+import scoring  # noqa: E402
+
+
+# ── venue-name + dedup helpers (the XYZ-prefix fix) ──
+
+def test_venue_asset_reattaches_xyz_prefix():
+    # the leaderboard returns BARE tokens with the DEX in a separate field; market reads + emits need xyz:
+    assert scan._venue_asset("NVDA", "xyz") == "xyz:NVDA"
+    assert scan._venue_asset("BRENTOIL", "xyz") == "xyz:BRENTOIL"
+    assert scan._venue_asset("BTC", "") == "BTC"          # main-DEX passes through
+    assert scan._venue_asset("xyz:GOLD", "xyz") == "xyz:GOLD"   # already prefixed → no double-prefix
+    assert scan._venue_asset("SILVER", None) == "SILVER"  # missing dex → treated as main (safe fallback)
+
+
+def test_bare_strips_prefix_for_dedup():
+    assert scan._bare("xyz:NVDA") == "NVDA"
+    assert scan._bare("NVDA") == "NVDA"
+    assert scan._bare("btc") == "BTC"
+
+
+# ── scan() smoke: an XYZ board asset must emit + be price-gated under its xyz: name ──
+
+class _State:
+    def __init__(self):
+        self._log = []
+
+    def last(self):
+        return self._log[-1] if self._log else None
+
+    def append(self, d):
+        self._log.append(d)
+
+    def __len__(self):
+        return len(self._log)
+
+
+class _MCP:
+    """Board carries a strong LONG on the XYZ asset NVDA (bare token, dex 'xyz') and on crypto BTC.
+    Captures every market_get_asset_data call so the test can assert the venue-correct name was used."""
+    def __init__(self):
+        self.asset_data_calls = []
+
+    def call_tool(self, tool, args):
+        args = args or {}
+        if tool == "strategy_get_clearinghouse_state":
+            return {"main": {"marginSummary": {"accountValue": "1000"}, "assetPositions": []},
+                    "xyz":  {"marginSummary": {"accountValue": "1000"}, "assetPositions": []}}
+        if tool == "leaderboard_get_markets":
+            return {"markets": [
+                {"token": "NVDA", "dex": "xyz", "direction": "long",  "pct_of_top_traders_gain": 80, "volume": 1000, "trader_count": 50},
+                {"token": "NVDA", "dex": "xyz", "direction": "short", "pct_of_top_traders_gain": 20, "volume": 1000, "trader_count": 50},
+                {"token": "BTC",  "dex": "",    "direction": "long",  "pct_of_top_traders_gain": 75, "volume": 2000, "trader_count": 60},
+                {"token": "BTC",  "dex": "",    "direction": "short", "pct_of_top_traders_gain": 25, "volume": 2000, "trader_count": 60},
+            ]}
+        if tool == "market_get_asset_data":
+            self.asset_data_calls.append({"asset": args.get("asset"), "dex": args.get("dex")})
+            return {"data": {"candles": {"4h": [{"high": 100, "low": 100} for _ in range(6)]}}}  # NEUTRAL → gate passes
+        return {}
+
+
+class _Ctx:
+    def __init__(self, mcp, state):
+        self.senpi_mcp = mcp
+        self.wallet = "0xWALLET0000000000000000000000000000wallet"
+        self.state = state
+
+
+def test_xyz_asset_emits_and_is_read_under_its_xyz_prefix():
+    mcp = _MCP()
+    inputs = {"minTiltLong": 58, "minTiltShort": 42, "marginPctBase": 12,
+              "maxLong": 2, "maxShort": 2, "leverageDefault": 4,
+              "leaderboardLimit": 100, "convictionVolFloor": 1.0}
+    out = scan.scan(inputs, _Ctx(mcp, _State()))
+    assets = {e["asset"] for e in out}
+    # THE FIX: the XYZ asset is emitted WITH its venue prefix, never the bare token
+    assert "xyz:NVDA" in assets, assets
+    assert "NVDA" not in assets, assets
+    assert "BTC" in assets, assets      # main-DEX asset is unaffected
+    # and the price-confirmation read hit the venue-correct name + dex — not bare NVDA on the main DEX
+    xyz_reads = [c for c in mcp.asset_data_calls if c["asset"] == "xyz:NVDA"]
+    assert xyz_reads and xyz_reads[0]["dex"] == "xyz", mcp.asset_data_calls
+    assert all(c["asset"] != "NVDA" for c in mcp.asset_data_calls), mcp.asset_data_calls
+
+
+if __name__ == "__main__":
+    import pytest
+    sys.exit(pytest.main([__file__, "-q"]))
--- a/strategies/stingray/tests/test_engine.py
+++ b/strategies/stingray/tests/test_engine.py
@@ -15,8 +15,13 @@ def test_venue_asset_reattaches_xyz_prefix():
     assert scan._venue_asset("NVDA", "xyz") == "xyz:NVDA"
     assert scan._venue_asset("BRENTOIL", "xyz") == "xyz:BRENTOIL"
     assert scan._venue_asset("BTC", "") == "BTC"          # main-DEX passes through
-    assert scan._venue_asset("xyz:GOLD", "xyz") == "xyz:GOLD"   # already prefixed → no double-prefix
-    assert scan._venue_asset("SILVER", None) == "SILVER"  # missing dex → treated as main (safe fallback)
+    # already-prefixed → canonicalised to lowercase xyz:. The board UPPER-cases tokens, so the REACHABLE
+    # prefixed spelling is XYZ:… — both must land on the canonical name, never the non-canonical XYZ:.
+    assert scan._venue_asset("XYZ:GOLD", "xyz") == "xyz:GOLD"
+    assert scan._venue_asset("xyz:GOLD", "xyz") == "xyz:GOLD"
+    # no dex on the row → can't infer the venue, so addressed on main. NOT "safe": an XYZ asset arriving
+    # without a dex silently no-trades (the same failure the PR fixes) — the board must carry dex.
+    assert scan._venue_asset("SILVER", None) == "SILVER"
 
 
 def test_bare_strips_prefix_for_dedup():
@@ -58,6 +63,9 @@ class _MCP:
                 {"token": "NVDA", "dex": "xyz", "direction": "short", "pct_of_top_traders_gain": 20, "volume": 1000, "trader_count": 50},
                 {"token": "BTC",  "dex": "",    "direction": "long",  "pct_of_top_traders_gain": 75, "volume": 2000, "trader_count": 60},
                 {"token": "BTC",  "dex": "",    "direction": "short", "pct_of_top_traders_gain": 25, "volume": 2000, "trader_count": 60},
+                # a strong SHORT on an XYZ commodity — the "short crypto/commodities" half of the thesis
+                {"token": "BRENTOIL", "dex": "xyz", "direction": "short", "pct_of_top_traders_gain": 80, "volume": 1500, "trader_count": 55},
+                {"token": "BRENTOIL", "dex": "xyz", "direction": "long",  "pct_of_top_traders_gain": 20, "volume": 1500, "trader_count": 55},
             ]}
         if tool == "market_get_asset_data":
             self.asset_data_calls.append({"asset": args.get("asset"), "dex": args.get("dex")})
@@ -87,6 +95,13 @@ def test_xyz_asset_emits_and_is_read_under_its_xyz_prefix():
     xyz_reads = [c for c in mcp.asset_data_calls if c["asset"] == "xyz:NVDA"]
     assert xyz_reads and xyz_reads[0]["dex"] == "xyz", mcp.asset_data_calls
     assert all(c["asset"] != "NVDA" for c in mcp.asset_data_calls), mcp.asset_data_calls
+    # the fix must hold on the SHORT side too — the "short crypto/commodities" half of the thesis
+    short_assets = {e["asset"] for e in out if e["direction"] == "SHORT"}
+    assert "xyz:BRENTOIL" in short_assets, out
+    assert "BRENTOIL" not in short_assets, out
+    brent_reads = [c for c in mcp.asset_data_calls if c["asset"] == "xyz:BRENTOIL"]
+    assert brent_reads and brent_reads[0]["dex"] == "xyz", mcp.asset_data_calls
+    assert all(c["asset"] != "BRENTOIL" for c in mcp.asset_data_calls), mcp.asset_data_calls
 
 
 if __name__ == "__main__":
--- a/senpi-strategy-discover/catalog.json
+++ b/senpi-strategy-discover/catalog.json
@@ -5447,7 +5447,7 @@
       "emoji": "🐠",
       "tagline": "Reads the entire smart-money board every tick, ranks every asset by net smart-money conviction (tilt x notional x breadth), then rides the rotation — LONG the assets the best traders are crowding into and SHORT the ones they're fleeing, gated by price so it never fights a hard trend. Captures days like 'long semis/indices, short the crypto block'.",
       "belief_plain": "Most copy-trading bots ask 'are the best traders long this one coin?'. Stingray instead looks at the whole market at once and asks 'what are the best traders rotating INTO, and what are they rotating OUT OF?'. It buys the crowd's new favorites and sells the ones they're abandoning — but only when the price chart isn't already screaming the other way — then trails a wide stop so a real rotation can run for days.",
-      "version": "1.2.0",
+      "version": "1.2.1",
       "thesis": "Smart money doesn't just pick coins, it rotates capital across the whole board — out of one asset class and into another. Per-asset copy-traders miss this because they evaluate each market in isolation; they can't see that the SAME flow leaving crypto is funding the bid in indices. Stingray ranks the entire board by net tilt (long_ratio - 50) weighted by notional and trader breadth, so a strong, deep, widely-held lean outranks a thin spike, then takes BOTH sides at once — long the inflows, short the outflows — for a market-neutral-ish rotation book. A price-confirmation gate (no longs into a hard 4h downtrend, no shorts into a hard uptrend) keeps it from fighting price, and a DSL ratchet lets the winning legs run while cutting the losers.",
       "tags": [
         "smart-money",
--- a/strategies/catalog.json
+++ b/strategies/catalog.json
@@ -5447,7 +5447,7 @@
       "emoji": "🐠",
       "tagline": "Reads the entire smart-money board every tick, ranks every asset by net smart-money conviction (tilt x notional x breadth), then rides the rotation — LONG the assets the best traders are crowding into and SHORT the ones they're fleeing, gated by price so it never fights a hard trend. Captures days like 'long semis/indices, short the crypto block'.",
       "belief_plain": "Most copy-trading bots ask 'are the best traders long this one coin?'. Stingray instead looks at the whole market at once and asks 'what are the best traders rotating INTO, and what are they rotating OUT OF?'. It buys the crowd's new favorites and sells the ones they're abandoning — but only when the price chart isn't already screaming the other way — then trails a wide stop so a real rotation can run for days.",
-      "version": "1.2.0",
+      "version": "1.2.1",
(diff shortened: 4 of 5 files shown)


strategies/stingray/main/scanners/scan.py
```diff
@@ -65,6 +65,25 @@ def _dex_for(asset):
...
+def _venue_asset(token, dex):
+    """The venue-correct asset name for BOTH the market read and the emitted trade. XYZ (HIP-3)
+    assets — equities/commodities/indices — are addressed as `xyz:<TOKEN>`, but `leaderboard_get_markets`
+    returns the BARE token (`NVDA`, `BRENTOIL`, `SILVER`) with the DEX in a separate `dex` field. Re-attach
+    the prefix here: without it, market_get_asset_data (and the trade itself) hit an asset that isn't on the
+    main DEX and fail — the exact issue the catalog validator caught. Main-DEX tokens pass through."""
+    t = str(token)
+    if str(dex).lower() == "xyz" and not t.lower().startswith("xyz:"):
+        return f"xyz:{t}"
+    return t
```
shnoodles: `_venue_asset` detects the existing prefix case-insensitively but returns the input **unchanged**, and the ingest at line 174 uppercases every token:

```python
token = str(m.get("token", ...)).upper()   # line 174
```

So if `leaderboard_get_markets` ever returns an already-prefixed token, it arrives here as `XYZ:NVDA`, the guard on line 75 matches, and line 77 emits it verbatim:

```
_venue_asset('XYZ:NVDA', 'xyz')  ->  'XYZ:NVDA'
```

Every real coin string I see in production is lowercase-prefixed (`xyz:NVDA`, `xyz:GOLD`), so `XYZ:NVDA` is non-canonical and would fail the same way the bare name does. `_dex_for` still returns `'xyz'`, so this fails *silently* at the venue rather than being caught locally.

Not reachable today — the board returns bare tokens, which is the whole premise of the PR. But the idempotency guard exists precisely to survive that contract changing, and this is the same class of drift that caused the original bug. Suggest normalising rather than passing through:

```python
t = str(token)
if t.lower().startswith("xyz:"):
    return "xyz:" + t[4:]          # canonical, regardless of input case
return f"xyz:{t}" if str(dex).lower() == "xyz" else t
```

Worth noting the test on `test_engine.py:18` asserts `_venue_asset("xyz:GOLD", "xyz") == "xyz:GOLD"` — a *lowercase* prefixed input, which line 174 can never produce. So the guard is currently tested only on the unreachable spelling and not on the reachable one.

Separately, and only as a note: Starling (cited as the pattern being mirrored) doesn't reconstruct the prefix at all — `scoring.name_map` keeps `{BARE_UPPER: raw coin string}` and line 204 emits `names.get(bare, bare)`, i.e. it preserves whatever the API returned. That sidesteps case normalisation entirely. Not asking you to refactor to it here, but it may be the better long-term shape if a third scanner needs this.

strategies/stingray/tests/test_engine.py
```diff
@@ -0,0 +1,94 @@
...
+
+# ── venue-name + dedup helpers (the XYZ-prefix fix) ──
+
+def test_venue_asset_reattaches_xyz_prefix():
+    # the leaderboard returns BARE tokens with the DEX in a separate field; market reads + emits need xyz:
+    assert scan._venue_asset("NVDA", "xyz") == "xyz:NVDA"
+    assert scan._venue_asset("BRENTOIL", "xyz") == "xyz:BRENTOIL"
+    assert scan._venue_asset("BTC", "") == "BTC"          # main-DEX passes through
+    assert scan._venue_asset("xyz:GOLD", "xyz") == "xyz:GOLD"   # already prefixed → no double-prefix
+    assert scan._venue_asset("SILVER", None) == "SILVER"  # missing dex → treated as main (safe fallback)
```
shnoodles: The trailing comment reads `missing dex -> treated as main (safe fallback)`, but this isn't safe — it's exactly the bug the PR fixes, just triggered by a missing field instead of a missing prefix:

```
_venue_asset('NVDA', None)  ->  'NVDA'    # addressed on the main DEX, read fails, leg silently skipped
```

The behaviour is a reasonable default (you can't infer the venue from nothing), but the whole failure mode here is that it degrades **silently** — the description notes the scanner "degrades gracefully" and that it took the catalog validator to notice. Given `dex` is now a single point of failure for venue correctness, I'd either reword the comment to say what it really is, or log when an XYZ-looking token arrives without a dex:

```python
if not dex and t.upper() in KNOWN_XYZ:
    print(f"[stingray.scan] WARN {t}: no dex on board row, addressing on main DEX", file=sys.stderr)
```

At minimum the comment shouldn't call it safe — a future reader will trust it.

strategies/stingray/tests/test_engine.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, mcp, state):
+        self.senpi_mcp = mcp
+        self.wallet = "0xWALLET0000000000000000000000000000wallet"
+        self.state = state
+
+
+def test_xyz_asset_emits_and_is_read_under_its_xyz_prefix():
+    mcp = _MCP()
+    inputs = {"minTiltLong": 58, "minTiltShort": 42, "marginPctBase": 12,
+              "maxLong": 2, "maxShort": 2, "leverageDefault": 4,
```
shnoodles, line 86: The smoke test only exercises the LONG side for an XYZ asset. With the fixture's `NVDA` short at `pct_of_top_traders_gain: 20` against `minTiltShort: 42`, and `BTC` short at 25, **no SHORT leg qualifies at all** — so `_emit_side(..., direction="SHORT", ...)` never runs with a HIP-3 asset in this test.

`_emit_side` is shared across both directions so the fix almost certainly holds, but "long semis, short crypto" is the strategy's stated thesis and the short half of it is the part currently unproven. Adding a fourth board row (or flipping NVDA's tilt in a second case) would close it cheaply:

```python
{"token": "BRENTOIL", "dex": "xyz", "direction": "short", "pct_of_top_traders_gain": 80, ...},
{"token": "BRENTOIL", "dex": "xyz", "direction": "long",  "pct_of_top_traders_gain": 20, ...},
```

then assert `xyz:BRENTOIL` appears in the SHORT emits and that its price-gate read used `dex='xyz'`.

shnoodles (commented):
Good fix, and the diagnosis in the description is right: `xyz:` markets do reject a bare name (Starling's `scoring.name_map` docstring says so explicitly), so addressing a HIP-3 asset as a main-DEX coin genuinely cannot trade. Using the board's own `dex` field to rebuild the name is the correct direction, and doing it for **both** the price-gate read and the emit is the important part.

**What I verified**
- Ran `strategies/stingray/tests` on the branch: **3 passed**.
- Checked the `_bare()` dedup collision risk against 30 days of production fills: **0 collisions across 535 distinct tickers** (no ticker is listed on both the main and `xyz` venues), so folding both venues into one dedup key is safe as things stand.

Three non-blocking comments below. The first is the one I'd actually like changed before merge; the other two are a test-comment correction and a coverage gap.

betashop:
All three addressed in `eecd2f5` — thanks, sharp as always.

1. **Canonicalization** — right, the guard passed `XYZ:…` through verbatim. `_venue_asset` now rebuilds an already-prefixed token as lowercase `xyz:` regardless of input case, and the test hits the REACHABLE uppercase spelling (`XYZ:GOLD`) instead of the lowercase one the board can never produce.
2. **"safe" comment** — reworded. It's a silent no-trade, not a safe fallback, and the comment now says exactly that (an XYZ asset without a dex fails the same way; the board must carry `dex`). I left out the `KNOWN_XYZ` warn-log — a hardcoded xyz set cuts against the derive-live universe, and "`dex` present on every board row" is the real invariant to lean on.
3. **SHORT side** — added a strong-short `xyz:BRENTOIL` to the smoke fixture; it now asserts the short leg emits `xyz:BRENTOIL` and price-gates under `dex=xyz`. The "short crypto/commodities" half is covered.

On the Starling note: agreed that preserving the raw coin string (bare→raw, no reconstruction) is the more robust long-term shape — especially for case-sensitive crypto like `kPEPE`. Left it out of scope here per your call; worth a shared helper if a third scanner needs it. 3 pass.

shnoodles:
Re-checked `eecd2f5` — all three are properly addressed, not just acknowledged:

1. **Canonicalization** — verified by probe, every previously-broken spelling now converges:
   ```
   _venue_asset('XYZ:NVDA','xyz') -> 'xyz:NVDA'
   _venue_asset('XYZ:NVDA','')    -> 'xyz:NVDA'   # prefix check now runs BEFORE the dex check
   ```
   Moving the prefix test ahead of the `dex` test is a nice extra — an already-prefixed token no longer depends on `dex` being present at all.
2. **Comment** — now says it's a silent no-trade. And you're right to drop the `KNOWN_XYZ` warn-log; a hardcoded set does cut against a live-derived universe, and "`dex` on every board row" is the better invariant. My suggestion was the weak part of that comment.
3. **SHORT side** — assertions at lines 99-104 do exercise it (`direction == "SHORT"`, prefixed present, bare absent, `dex=xyz` on the read). 3 pass on my side too.

**LGTM — merge it.**

---

On your `kPEPE` aside: that isn't hypothetical, it's live right now, and it's the same bug you just fixed wearing different clothes.

`scan.py:174` uppercases every board token:
```python
token = str(m.get("token", m.get("coin", m.get("asset", "")))).upper()
```
Six actively-traded Hyperliquid coins carry a **lowercase** `k`:

| coin | fills (30d) | wallets |
|---|---|---|
| `kPEPE` | 193,210 | 547 |
| `kSHIB` | 90,635 | 331 |
| `kBONK` | 45,538 | 231 |
| `kNEIRO` | 17,538 | 71 |
| `kLUNC` | 16,870 | 65 |
| `kFLOKI` | 12,794 | 73 |

and casing can't be normalised in either direction, because `KAITO` and `KAS` are legitimately uppercase.

**Proof it's biting today.** Across every stingray wallet: **0 fills on any of those six**, against 49 fills on `KAITO`/`KAS` (1,762 fills / 83 distinct coins total). And the failure is in the logs verbatim — `SKIP KPEPE SHORT: trend read failed (don't size blind)`, 175× on M408865 alone, plus `KBONK`, `KSHIB`, `KNEIRO`, `KFLOKI`, `KLUNC`. **801 such skips across 3 users in the last 7 days.**

`KPEPE` isn't a coin, so `market_get_asset_data` fails, `_price_ok` returns `None`, and the leg is skipped — structurally identical to the XYZ case, just on the main DEX. Stingray is currently blind to the entire k-prefixed meme complex.

The fix is the split this PR already introduces, extended from prefix to case: **uppercase for the dedup key, raw board spelling for the venue name.** `_bare()` already gives you the former. The latter just means not destroying the original at line 174 — carry the raw token alongside and feed *that* to `_venue_asset`.

Which is, I think, the argument for the Starling `name_map` shape (bare→raw, never reconstruct) rather than against it: it fixes the xyz prefix and the k case with one mechanism. Happy for that to be its own PR rather than growing this one.

One caveat so the number isn't overread: those 801 are only ~2% of the 37,137 `trend read failed` events fleet-wide over the same window. The other 98% is a different fault (the MCP `ClosedResourceError` family) — this explains a slice of that log line, not all of it.

shnoodles:
Concrete fix for the `kPEPE` case, since a comment alone isn't much use. I built and verified it — **4 edits across 2 files**, and it's one more edit than you'd expect because `scoring.rank_board` rebuilds the candidate dict from an explicit key allowlist, so anything extra you hang on the row gets silently dropped. That's the wall you'd have hit.

**Verified before/after** with the live board's own spelling (`kPEPE` short-tilted, `KAITO` long-tilted):

```
eecd2f5      emitted=['KAITO', 'KPEPE']   reads=['KAITO', 'KPEPE']   -> FAIL
with fix     emitted=['KAITO', 'kPEPE']   reads=['KAITO', 'kPEPE']   -> PASS
```

`KAITO` is unaffected (legitimately uppercase) and your existing 3 tests still pass.

```diff
--- a/strategies/stingray/main/scanners/scan.py	2026-08-24 12:51:27
+++ b/strategies/stingray/main/scanners/scan.py	2026-08-24 12:55:47
@@ -174,7 +174,8 @@
     for m in markets:
         if not isinstance(m, dict):
             continue
-        token = str(m.get("token", m.get("coin", m.get("asset", "")))).upper()
+        raw_token = str(m.get("token", m.get("coin", m.get("asset", ""))))
+        token = raw_token.upper()          # UPPER: aggregation + dedup key only
         if not token:
             continue
         dex = str(m.get("dex", "")).lower()
@@ -188,7 +189,7 @@
 
         key = (token, dex)
         row = by_asset.setdefault(key, {
-            "token": token, "dex": dex,
+            "token": token, "raw": raw_token, "dex": dex,   # raw preserves case (kPEPE vs KAITO)
             "long_pct": 0.0, "short_pct": 0.0,
             "volume": 0.0, "traders": 0.0,
         })
@@ -317,7 +318,7 @@
         if _recently_signaled(signaled, cu, ttl, now):
             continue
         # venue-correct name (xyz:<TOKEN> for HIP-3 assets) — required by market_get_asset_data AND the emit
-        venue = _venue_asset(c["token"], c.get("dex"))
+        venue = _venue_asset(c.get("raw") or c["token"], c.get("dex"))
         ok = _price_ok(ctx, venue, direction)
         if ok is None:
             print(f"[stingray.scan] SKIP {venue} {direction}: trend read failed (don't size blind)",
--- a/strategies/stingray/main/scanners/scoring.py	2026-08-24 12:51:28
+++ b/strategies/stingray/main/scanners/scoring.py	2026-08-24 12:57:14
@@ -83,6 +83,7 @@
         weight = conviction_weight(r.get("volume", 0), r.get("traders", 0), vol_floor)
         base = {
             "token": r.get("token", ""),
+            "raw": r.get("raw", "") or r.get("token", ""),   # board spelling, for the venue name
             "dex": r.get("dex", ""),
             "long_ratio": round(long_ratio, 2),
             "net_tilt": round(tilt, 2),
```

The shape is the same split you already introduced, applied to case instead of prefix: **uppercase for the aggregation/dedup key, the board's raw spelling for the venue name.** `key = (token, dex)` stays uppercase so long/short rows still aggregate, `_bare(c["token"])` stays uppercase so dedup against held positions is unchanged, and only the string that goes to `market_get_asset_data` and `asset` changes.

Confirmed against the live board that it is the source of truth for casing:

```
"token": "kBONK"   "token": "kPEPE"   "token": "kSHIB"    <- lowercase k
"token": "KAITO"   "token": "KAS"     "token": "KIOXIA"   <- uppercase K
```

(and 0 tokens carry an `xyz:` prefix today, which independently confirms this PR's premise).

Suggested test to go with it:

```python
def test_lowercase_k_coin_keeps_board_casing():
    # kPEPE/kSHIB/kBONK/kNEIRO/kLUNC/kFLOKI are lowercase-k; KAITO/KAS are not.
    # Casing cannot be normalised in either direction — the board spelling is the only truth.
    assert scan._venue_asset("kPEPE", "") == "kPEPE"
    assert scan._venue_asset("KAITO", "") == "KAITO"
    assert scan._venue_asset("NVDA", "xyz") == "xyz:NVDA"
```

Happy to raise this as its own PR against `main` after yours lands, or you can take the diff — whichever you prefer. It's independent of the xyz work, just the same failure mode.

strategies/stingray/main/scanners/scan.py
```diff
@@ -65,6 +65,28 @@ def _dex_for(asset):
     return "xyz" if str(asset).lower().startswith("xyz:") else ""
 
 
+def _venue_asset(token, dex):
+    """The venue-correct asset name for BOTH the market read and the emitted trade. XYZ (HIP-3)
+    assets — equities/commodities/indices — are addressed as `xyz:<TOKEN>`, but `leaderboard_get_markets`
+    returns the BARE token (`NVDA`, `BRENTOIL`, `SILVER`) with the DEX in a separate `dex` field. Re-attach
+    the prefix here: without it, market_get_asset_data (and the trade itself) hit an asset that isn't on the
+    main DEX and fail — the exact issue the catalog validator caught. Idempotent AND case-canonical: an
```
shnoodles, line 73: "the exact issue the catalog validator caught" is not accurate, and it is the kind of claim that costs the next author a manual review.

I reverted `scan.py` alone to `origin/main` (confirming the emit went out bare) and ran both validators against that tree:

```
validate_strategy.py strategies/stingray   ->  ✓ stingray v1.0.1        exit 0
validate_universe.py strategies/stingray   ->  ✓ derived universe        exit 0
```

Both pass on the broken code. `validate_strategy.py` never inspects the emitted `asset` value at all; its `scan.py` work is an `ast.parse` for syntax, a candle-key check, a `signal_data_schema` check, and a `@senpi/runtime` string check. `validate_universe.py:96-107` reads only `strategy.yaml catalog.assets` plus `runtime*.yaml` inputs, and stingray has neither.

`validate_universe.py:111-115` is the part worth reading, because it bakes in the assumption this PR is fixing:

> Scanners on the xyz DEX prefix bare names in code (`f"xyz:{token}"`), so a bare ticker is valid if `T` or `xyz:T` is live

Its correctness depends on scanners doing exactly what stingray was failing to do. A repo-wide grep for "catalog validator" returns exactly one hit: this docstring.

Suggest dropping that clause, or replacing it with what actually caught it. Same claim appears in the PR body.

Separately, "case-canonical" overstates it. The function canonicalizes the `xyz:` prefix and never the token body, which is the kPEPE bug in my other comment.

shnoodles (commented):
## Re-review of `eecd2f5`

**The core fix is correct and should land.** Pre-PR every `dex="xyz"` board row went to the venue as a bare token and was rejected. That is 95 of 281 live board rows, ~34% of the board, confirmed live: `market_get_asset_data(asset="NVDA")` returns `INVALID_ARGUMENT, Unknown coin "NVDA" on the main Hyperliquid dex` while `asset="xyz:NVDA"` returns `success:true`. The three new tests pass and `validate_strategy.py` reports `✓ stingray v1.0.1`.

Everything below is either a blocker on this PR's own artifacts, or a pre-existing bug sitting in a function this PR touches. I have left inline comments on the ones that land on diff lines. The rest are here because they are outside the diff.

### Blocker: the catalog was not regenerated

`strategy.yaml:11` is now `version: "1.0.1"`, but both generated copies still read `"version": "1.0.0"`:

- `strategies/catalog.json:5193`
- `senpi-strategy-discover/catalog.json:5193`

The PR body says the catalog stores no per-strategy version so no regen is needed. That is not the case: `gen_catalog.py:248` emits `"version": m.get("version")` and writes both paths at lines 307-313. Running it produces exactly the missing one-line diff in each file.

Stingray is the only one of 102 catalogued packages whose `strategy.yaml` version disagrees with its catalog entry. The cost is that deploys stamp `skillVersion=1.0.1` (`senpi-strategy-ops/scripts/_pkg.py:161` reads `strategy.yaml` directly) while `senpi-strategy-discover/scripts/discover.py:309` serves `1.0.0` to every agent that recommends Stingray, so fleet analytics cannot answer "is this wallet running the fixed build?". The next unrelated `gen_catalog.py` run will also sweep this bump into someone else's PR.

### The new tests never run in CI

`.github/workflows/tests.yml:31` enumerates six `senpi-*` dirs and no `strategies/`. A repo-wide find for `pytest.ini` / `pyproject.toml` / `tox.ini` / `setup.cfg` / `conftest.py` returns nothing, so no config widens collection. All 23 `strategies/*/tests` dirs are dark, and a later edit reverting `_venue_asset` or `_bare` goes red locally only while CI stays green.

Worth flagging that this is not a one-line fix today: `pytest strategies -q` currently dies with `INTERNALERROR> SystemExit: 0` from a module-scope `sys.exit()` in `strategies/chimp/tests/test_package_shape.py:64`, and 18 files share the basename `test_engine.py` with no `__init__.py`, so default import mode errors with "import file mismatch".

### Pre-existing bugs in the functions this PR touches

Not introduced here, and I would not block on them, but this is the moment someone is looking at these lines. All verified against the live API.

**1. The conviction ranking is completely inert.** `scan.py:186` reads `volume` / `avg_volume_6h` / `avgVolume`; live `leaderboard_get_markets` returns `day_notional_volume` on 283/283 rows. Zero rows carry any field the scanner looks for, so `volume` is 0 everywhere and `scoring.conviction_weight` returns `log10(1+0) * breadth` = 0.0. The `|net_tilt|` fallback that exists for exactly this case never fires, because `scoring.py:109 board_has_weight` uses `or` where `conviction_weight` uses `*`, and `traders > 0` is true on the real board. Replaying the live board: 266 normalized rows, distinct conviction `{0.0}`, LONG order before the sort identical to after. Picks are raw API arrival order. Emitted LONG `[LIT, XMR]`; with the correct volume field it is `[LIT, SOL]`.

This one interacts with the PR: pre-PR the inert ranking chose among ~174 main-DEX assets, post-PR 22 of 38 LONG and 69 of 226 SHORT candidates are xyz, so arbitrary ordering now governs a much larger pool including thin new HIP-3 listings that `conviction_weight` exists to down-rank.

**2. One-sided board rows fabricate a maximum tilt.** `scan.py:190` seeds `long_pct` / `short_pct` at 0.0 and only accumulates directions present in the payload. The live board carries an `is_dominant_direction` flag and returns only the dominant side for 231 of 253 assets, so each becomes `long_ratio` exactly 100 or 0, i.e. `net_tilt` ±50. `rank_board` then admits 74 LONG + 178 SHORT = 252 of 253 assets: the documented "minimum tilt 58/42" gate rejects exactly one. 71 of 74 long candidates sit at `|net_tilt| = 50.0`, clearing `margin_pct_for`'s `>= 20` tier, so every pick emits at base × 1.5. All four picks in the replay came out at `marginPct 18.0` against a documented 12. `pct_of_top_traders_gain` is entirely discarded for these assets: a row with `pct=26.3` and one with `pct=0.3` both normalize to `long_ratio=100`.

**3. `scan.py:177` `.upper()` still breaks the k-prefixed perps.** Covered in my earlier comment with a tested 4-edit diff. Currently live and affected: kBONK, kFLOKI, kLUNC, kPEPE, kSHIB. Note the new `_venue_asset` docstring calls itself "case-canonical", which is only true of the prefix, never the token body. `strategies/cuckoo/main/scanners/scoring.py` states the rule directly: "v2 upper-cased it, which is wrong for a derived universe ... Callers needing a case-insensitive COMPARISON upper-case at the comparison site instead."

**4. `scan.py:189` keys `by_asset` on the raw token.** Two spellings of one market, or a `"dex": null`, split it into two one-sided rows that both fabricate to ±50. Post-PR that turns a fail-closed skip into a live inverted max-size trade: given `{"token":"NVDA","dex":"xyz","direction":"long","pct":29}` plus `{"token":"xyz:NVDA","dex":"xyz","direction":"short","pct":71}`, the emit is `xyz:NVDA LONG marginPct 18.0`, opposite the real aggregate. Pre-PR the bare leg logged `SKIP NVDA trend read failed` and only the correct SHORT survived. Separately, an explicit JSON `"dex": null` yields `str(None).lower() == "none"` at line 180 (the `.get` default only fires when the key is absent), a phantom third venue that evades the `xyzBanned` filter at line 181 and routes to the main DEX.

Reachability is unproven, 0 of 281 live rows contain a colon in `token`, have a null `dex`, or appear under two dex values. But `key = (_bare(token), dex)` closes both at no cost.

**5. `leaderboardLimit` does the opposite of what its name says.** `scan.py:155` passes it as the MCP `limit`, which selects the top-N *traders to aggregate over*, not the board breadth. `scan.py:45` calls it "board breadth to scan", `scan.py:368` names it `leaderboardLimit`, and `runtime.yaml:54` repeats the framing. Measured: `limit=100` returns 275 rows / 253 assets, and the row count is unrelated to the value. A user raising it to 500 to "scan more of the board" gets the same board recomputed over a 5x wider, lower-quality trader cohort, silently re-deriving every `net_tilt`.

### The fix landed one package deep

`cheetah/main/scanners/scan.py:367` emits `"asset": best["token"]`, the bare token, with `dex` carried separately at :101 and never re-attached. `orca/main/scanners/scan.py:390` is identical. Both are disarmed only by `xyzBanned: true` defaults in a user-visible `runtime.yaml` (`cheetah/runtime.yaml:38`, `orca/runtime.yaml:50`). A user flipping either to false to "trade equities too" gets a strategy that silently never trades any XYZ name, with no validator to catch it. Stingray was the only broken package shipping `xyzBanned: false`; cheetah and orca are the loaded ones.

`orca:187-190` also adds a third spelling, bare `asset` plus explicit `dex="xyz"`, so there are now three conventions in the repo.

The correct prior art is **lemon**, already right at `strategies/lemon/main/scanners/scan.py:160`:

```python
"asset": f"xyz:{token}" if dex == "xyz" else token
```

shipped with `xyzBanned: false`. Worth pointing `_venue_asset`'s attribution at lemon rather than cheetah, since `scan.py:141-142` currently credits "cheetah `_fetch_sm_markets` field extraction", which is how this bug propagated in the first place.

### Refuted

Two first-pass candidates that did not survive verification, noting them so nobody re-raises them: the `marginPctBase` ×100 overflow (blocked at three separate validation layers) and an `xyzBanned` bypass regression (pre-PR already emitted a name the venue accepts, so zero PR delta).

strategies/stingray/main/scanners/scan.py
```diff
@@ -289,26 +311,27 @@ def _emit_side(ctx, ranked, direction, cap, held_set, signaled, ttl, now,
     for c in ranked:
```
shnoodles, line 311: This walk issues one blocking network read per surviving candidate and breaks only on `len(emits) >= cap`, and the PR removes what was accidentally acting as a circuit breaker.

Live board: 265 normalized assets, `rank_board` yields 38 LONG + 227 SHORT = 265 candidates. Measured `market_get_asset_data` latency is 0.822s/call (40-call sample, handshake amortized), so a full walk is ~218s against `runtime.yaml:42 timeout_seconds: 180`. Per `runtime-concepts.md:50-51` a timed-out tick also rolls back the dedup map, so the next tick re-walks from scratch.

To be fair about severity: the typical tick is fine. I measured 11 calls / ~9.6s, because the price gate only blocks hard trends. This is a tail risk on a broad one-directional day (when the disfavored side swells toward 227) or a degraded backend, not a live outage.

The PR's contribution is real but smaller than it first looks. Pre-fix an xyz candidate went out bare and was rejected in ~0.42s, so a full walk was roughly `173*0.822 + 92*0.42 ≈ 181s`, already at the line. Post-fix it is `265*0.822 ≈ 218s`.

Prior art caps before the read loop rather than inside it: `orca/scan.py:165` `return markets[:top_n]`, `bloodhound/scan.py:113` `out[:max_names]`. A `len(positions) >= slots` early exit like `hen/scan.py:133-136` would additionally save ~3,200 discarded reads/day while fully slotted.

strategies/stingray/main/scanners/scan.py
```diff
@@ -289,26 +311,27 @@ def _emit_side(ctx, ranked, direction, cap, held_set, signaled, ttl, now,
...
                   file=sys.stderr)
             continue
         margin_pct = scoring.margin_pct_for(abs(c["net_tilt"]), base_margin, max_margin)
         signaled[cu] = now
         emits.append({
-            "asset": coin,
+            "asset": venue,
             "direction": direction,
             "marginPct": margin_pct,          # PERCENT in (0,100] — runtime sizes the dollars
             "leverage": leverage,             # clamped [1,5] + venue max
```
shnoodles, line 337: `# clamped [1,5] + venue max` describes a clamp that no code in this package performs.

`_board` parses the per-row `max_leverage` at lines 190-201 and drops it on the floor. `scan.py:364` clamps only to the static `[1,5]`. Nothing under `senpi-trading-runtime/references/` documents a runtime-side venue clamp either.

86 of 275 live board rows carry `max_leverage: 3` (PURR, GRASS, HEMI, AERO, ACE, and others). Combined with the ±50 tilt fabrication I described in the summary, any of those that is one-sided becomes a max-tier candidate emitted at 4x against a venue cap of 3x.

Both outcomes are bad and hard to see:

- if the runtime clamps, the leg opens at 3x while `data.leverage: 4` records the wrong number and notional lands 25% under intent
- if it does not, the exchange rejects the order and the rotation leg silently never opens, which is the funded + ACTIVE + 0 fills + no errors signature

`_board` already reads the field. Threading it through and clamping to `min(leverage, row_max)` is small. Failing that, the comment should be deleted so it stops asserting a guarantee that is not there.

strategies/stingray/main/scanners/scan.py
```diff
@@ -358,7 +381,7 @@ def scan(inputs, ctx):
                 print(f"[stingray.scan] WARNING: state append failed: {exc!r}", file=sys.stderr)
         return []
     held_assets = [p["coin"] for p in positions if p.get("coin")]
```
shnoodles, line 383: This projection drops `direction` and `margin`, which makes the held gate direction-blind.

`_get_account` builds `{"coin", "direction", "margin"}` per position at line 121. Only `coin` survives here, so `direction` and `margin` are dead fields.

Concrete consequence: the strategy is LONG SOL from an earlier rotation, the board flips and SOL becomes short-dominant, `rank_board` places it on the SHORT side at `net_tilt -50`. In `_emit_side`, `cu = "SOL"` is found in `held_set` and hits `continue` at line 315. The rotate-out signal is never emitted.

Two things make it hard to notice. Unlike the price gate, which prints at lines 323 and 327, the held skip prints nothing at all. And because the DSL is the only exit, the losing long then rides to the ratchet or the 48h `hard_timeout` while the scanner had the reversal in hand the whole time.

Carrying `direction` through and allowing a reversal to pass the gate would use the fields `_get_account` is already computing.

strategies/stingray/main/scanners/scan.py
```diff
@@ -358,7 +381,7 @@ def scan(inputs, ctx):
...
-    held_set = {h.upper() for h in held_assets}
+    held_set = {_bare(h) for h in held_assets}     # strip xyz: so held HIP-3 positions dedup vs bare board tokens
```
shnoodles, line 384: `_bare` discards the DEX that `_board` deliberately keeps in its key, so a main-DEX ticker and its xyz twin now block each other.

Held `xyz:GOLD` plus a board carrying main-DEX GOLD net-long: post-PR emits `[('BTC','LONG')]`, GOLD dropped before any price read. Pre-PR (`{h.upper()}`) emitted `[('BTC','LONG'), ('GOLD','LONG')]`. So the held collapse is new in this PR.

The clearinghouse really does return prefixed coins, so this is not hypothetical shape-wise: `senpi-strategy-ops/tests/test_dex_awareness.py:51` uses `{"coin": "xyz:GOLD"}` under a comment noting "the REAL upstream shapes, verified against prod MCP 2026-07-23", and lines 71-76 assert the DEX-distinctness invariant fleet-wide.

The `signaled` half is worse for this strategy's thesis. One dict is shared across the LONG call at line 406 and the SHORT call at line 408, so emitting main GOLD LONG sets `signaled['GOLD']` and the `xyz:GOLD` SHORT is dropped at line 317 before its price read. That is exactly the cross-asset long/short pair the strategy exists to build. (That half is pre-existing, line 177 already uppercased before this PR.)

**Latent today, not firing.** Three independent checks all show a bare-ticker intersection of zero between the venues: `market_list_instruments` 232 main / 116 xyz, 90d raw fills 434 main / 101 xyz, live board 186 main / 95 xyz rows. So nothing is broken right now, and I would not block on it. But it is one HIP-3 listing of an existing crypto ticker away from firing, and keying on `(bare, dex)` in both gates matches what `_board` already does.

strategies/stingray/tests/test_engine.py
```diff
@@ -0,0 +1,109 @@
...
+    assert scan._bare("btc") == "BTC"
+
+
+# ── scan() smoke: an XYZ board asset must emit + be price-gated under its xyz: name ──
+
+class _State:
+    def __init__(self):
+        self._log = []
+
+    def last(self):
```
shnoodles, line 39: `_State.last()` never executes, so the one behavior `_bare` exists for has zero coverage.

`sys.settrace` instrumentation over the suite:

```
test_engine.py:40  (_State.last body)     hits = 0
scan.py:286        (ctx.state.last())     hits = 0
```

`_load_signaled` short-circuits on the empty log at `scan.py:284`, so `signaled` is written and never read back. The round-trip through `ctx.state` is the whole point: it is what stops a held XYZ position being re-entered on the next tick.

A real mutant escapes green. Changing `signaled[cu] = now` to `signaled[venue] = now` (dedup keyed by the prefixed name instead of `_bare`) leaves the suite at 3 passed, while a two-tick run re-emits `xyz:NVDA` and `xyz:BRENTOIL` 240s later. That is the scanner re-opening positions it already holds.

The adjacent held path is equally dark: the fixture returns `assetPositions: []`, so `held_set = {_bare(h) for h in held_assets}` at `scan.py:384` never runs. That is precisely the case `_bare` was written for, where the clearinghouse returns `xyz:NVDA` and the board returns bare `NVDA`.

A second tick against a seeded state, and a fixture with one held `xyz:` position, would close both.

strategies/stingray/tests/test_engine.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self):
+        self.asset_data_calls = []
+
+    def call_tool(self, tool, args):
+        args = args or {}
+        if tool == "strategy_get_clearinghouse_state":
+            return {"main": {"marginSummary": {"accountValue": "1000"}, "assetPositions": []},
+                    "xyz":  {"marginSummary": {"accountValue": "1000"}, "assetPositions": []}}
+        if tool == "leaderboard_get_markets":
+            return {"markets": [
```
shnoodles, line 61: The fake's response envelope is shallower than the real API, so the unwrap branch production actually takes is untested.

Live `leaderboard_get_markets` returns triple-nested:

```json
{"success": true, "data": {"markets": {"markets": [...]}}}
```

This fake returns `{"markets": [...]}`, which exercises `_board`'s shallow `data.get("markets") -> list` path and never the `isinstance(markets, dict) -> markets.get("markets")` unwrap at `scan.py:164-165` that production depends on.

The failure mode is the bad one: **delete `scan.py:164-165` and this suite still passes green**, while `_board` returns `[]` on every production tick and the strategy never trades again. The test cannot distinguish a working scanner from a permanently silent one.

Same divergence twice more in this fixture:

- line 58, the clearinghouse fake omits the `success` / `data` envelope, so `_get_account`'s `ch.get("data", ch)` unwrap is untested
- line 70, the candle fake has no `success` key, so `_hard_trend`'s `md.get("success") is False` guard is untested. The real payload also keys highs/lows as strings `h` / `l`, not `high` / `low`

Since the tests are new in this PR, matching the real envelope now is cheap and makes them worth having.

PR #549, opened 2026-08-27
fix forward-only and move overclaims

im-vignesh:
Two overclaims, both routed to a wrong action. "Exit changes are forward-only" let an agent tell a user their open positions were untouched after an order_type change — the exit: siblings (order_type, interval_seconds, fee options) are read live and DO reach open positions once an apply lands; only the ladder is frozen. And "a renamed or moved scanner" sent a harmless internal-scanner move to close-and-redeploy — a market exit — when only EXTERNAL scanner state is keyed by name and position. Length-neutral: ops stays at its 300-line body budget.

### DIFF
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -352,10 +352,10 @@ loop every time:
 > deployed package — "make my live strategy more aggressive", re-tune, re-score) — then hand it to
 > **`senpi-strategy-ops`**, which applies it IN PLACE with `openclaw senpi update`: no close, no fresh
 > wallet, no market exit. **Re-running `create` will NOT apply it** — the deploy verb is idempotent, so it
-> adopts the existing wallet and leaves the deployed scanner as it is. Tell the user two things: exit
-> changes are **forward-only**, so a tightened `dsl_preset` governs new entries and never reaches a
-> position already open; and a changed `strategy.wallet`, a renamed or moved scanner or a changed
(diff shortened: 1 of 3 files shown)


betashop:
Review notes for this stack are on #546 (ship order vs the runtime release, the missing version bumps, `--id`, and the after-apply re-read). Nothing specific to the three wording fixes here.

shnoodles:
**Review: yes.** I grepped every `forward-only` in the repo after this change — four remain, all four now scoped to `dsl_preset`, none still claiming "exit changes" wholesale. No stragglers. The `scanner` → `external scanner` tightening is a real correction too: state is keyed by name and position, which is only true of external scanners.

**One gap worth closing in this PR.** `senpi-strategy-ops/references/editing-a-live-strategy.md:36-40` is already scoped correctly to `dsl_preset`, so it is not an overclaim and the grep passes — but it never states the **positive** half this PR adds everywhere else:

```
**forward-only**: a new `dsl_preset` governs NEW entries, while every position already open keeps a
snapshot of the preset it was opened under. A tightened stop does **not** reach a position that is
already running…
```

No mention that `order_type` / `interval_seconds` / fee options are read live and **do** reach open positions. That file is the deep reference `senpi-strategy-ops/SKILL.md` defers to ("Everything else: `references/editing-a-live-strategy.md`"), so an agent that follows the pointer for detail gets only the frozen half — the exact asymmetry this PR exists to fix. One sentence after line 40.

Ponytail pass: 10 lines changed, 10 removed, three files, no new structure. Exactly the shape a wording fix should have. `net: 0 lines`.

PR #556, opened 2026-09-03
ops: an empty runtime inventory is an answer, not a failed read

shnoodles:
## The bug

`_parse_runtime_list` tested for the `"No runtimes installed."` line **only after** it had seen the table header. A box with no runtimes prints that line and **no header at all**, so the `continue` in the not-yet-seen-header branch skipped straight past it. Neither `empty_marker` nor `seen_header` was ever set, and the parser returned `(rows=[], valid=False)`.

`list_runtimes_or_none` turns `valid=False` into `None`, whose documented meaning is *"the inventory is unreadable"*. Its own docstring is explicit about why that distinction matters:

> `or_none` is the whole point: `[]` means "no runtimes", `None` means "the inventory is unreadable", and a check that reads the second as the first reports every strategy on the box as runtime-less.

The bug is that same confusion, inverted. On a box with zero runtimes the read came back as a **failure** rather than as an empty list.

## Blast radius: two call sites, both wrong on an empty box

**1. `deploy.py verify` could not answer at all.** The check whose entire job is *"is this strategy live?"* raised `ReadFailed`:

```
? stingray: COULD NOT CHECK — a read this check needs failed, so NOTHING about
  stingray is verified here (this says nothing about whether it is live):
    - `openclaw senpi runtime list` could not be read — the runtime inventory
      is not visible from here
```

while `openclaw senpi runtime list`, run directly on the same box seconds later, answered fine with `No runtimes installed.` So `verify` was structurally unable to answer **precisely on the boxes whose honest answer is "nothing is running"** — which is the broken-deploy case it exists for.

Found on a live box (M416163) where a funded `ACTIVE` strategy had no runtime installed. `verify` could not say so, and the deploy was reported to the user as successful.

**2. `close.py`'s teardown money path reported failure on a successful delete.** `_runtime_gone` reads the same helper. Measured with `run_cli` stubbed to the real empty-box output:

| | `list_runtimes_or_none()` | `_runtime_gone("pkg-main")` |
|---|---|---|
| `origin/main` | `None` | `False` |
| this branch | `[]` | `True` |

So deleting the **last** runtime on a box left an inventory that read back as unreadable: `close_one` retried the delete, then recorded `status: "failed"` — *"still in (or unreadable from) `runtime list` after delete"* — on a teardown that had actually succeeded. The new `True` does not weaken the fail-closed rule, because the inventory genuinely **was** read and `[]` genuinely means the runtime is absent.

## The fix

Hoist the empty-inventory test above the header gate, **anchored** with `low.startswith("no runtimes")`.

Anchoring matters because running before the header gate means the test now sees every stdout line, and an unanchored substring would let any future line merely *containing* the phrase short-circuit to a confident "empty" — the fail-**open** direction this parser exists to prevent. The only stdout emission today is `console.log("No runtimes installed.")`; the near-miss `"…No runtimes were read."` is an unreadable state on stderr with `rc != 0`, which `list_runtimes_or_none` already rejects before parsing. Anchoring excludes it a third time and errs toward `None` if the wording ever drifts.

The guard the parser was built for is untouched: a banner-only or garbled read still has no header, no rows and no marker, so it stays `valid=False` → `None`.

## Tests

**`test_cli_reads.py`** — new `ParseRuntimeListEmptyVsUnreadable`, alongside the sibling *"unreadable != empty"* contract that already lives there (module docstring extended to name this direction too). Empty inventory parses and reaches the `_or_none` caller as `[]`; a populated table still parses every row; banner-only, `rc != 0`, and a line that merely mentions the phrase all stay `None`.

**`test_close.py`** — new `CloseLastRuntimeOnTheBox`, driven through the **real** parser via `run_cli` rather than a stubbed `list_runtimes_or_none`, because the existing `CloseRuntimeDeleteConfirm` stubs that helper and so cannot see this. Covers: the empty inventory reads as gone, the teardown reports `closing` with exactly one delete (no redundant retry), and a garbled read on the same path still fails closed.

**Verified both directions.** 4 tests across the two files fail on `origin/main`'s `_cli.py` and pass with the fix. The two fail-closed guards pass on both, which is the point of them.

```
python3 -m pytest senpi-strategy-ops/tests -q
362 passed, 1 skipped, 45 subtests passed
```

Also green: `test_min_budget_vendor_parity.py` (VENDOR PARITY OK) and `test_min_budget_golden.py` (GOLDEN OK — 107 templates).

One unrelated pre-existing failure across the wider CI selection, `senpi-portfolio/tests/test_portfolio.py::test_all_step_is_byte_identical_to_run`, confirmed failing identically on clean `origin/main`.

### DIFF
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -22,7 +22,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.7.0"
+  version: "3.6.24"
   platform: senpi
   exchange: hyperliquid
   requires:
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -22,7 +22,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.6.23"
+  version: "3.7.0"
   platform: senpi
   exchange: hyperliquid
   requires:
--- a/senpi-strategy-ops/scripts/_cli.py
+++ b/senpi-strategy-ops/scripts/_cli.py
@@ -207,13 +207,19 @@ def _parse_runtime_list(out):
         if not line:
             continue
         low = line.lower()
+        # The empty-inventory line is tested BEFORE the header gate. A box with no runtimes prints
+        # "No runtimes installed." and NO header row, so looking for it only after a header is seen
+        # means it is never seen at all: the `continue` below skips it. The (rows=[], valid=False)
+        # that falls out is read as "inventory unreadable" by `list_runtimes_or_none`, inverting the
+        # very contract this parser exists to hold — and it does so precisely on the boxes whose
+        # honest answer is "nothing is running", which is the broken-deploy case `verify` is for.
+        if "no runtimes" in low:
+            empty_marker = True
+            break
         if not seen_header:
             if low.startswith("id") and "status" in low and "wallet" in low:
                 seen_header = True
             continue
-        if "no runtimes" in low:
-            empty_marker = True
-            break
         parts = [p for p in re.split(r"\s{2,}|\t+", line) if p]
         if len(parts) >= 2:
             rows.append({"name": parts[0], "wallet": parts[1],
--- a/senpi-strategy-ops/scripts/_cli.py
+++ b/senpi-strategy-ops/scripts/_cli.py
@@ -213,7 +213,16 @@ def _parse_runtime_list(out):
         # that falls out is read as "inventory unreadable" by `list_runtimes_or_none`, inverting the
         # very contract this parser exists to hold — and it does so precisely on the boxes whose
         # honest answer is "nothing is running", which is the broken-deploy case `verify` is for.
-        if "no runtimes" in low:
+        #
+        # ANCHORED, because this now runs before the header gate and so sees every stdout line: an
+        # unanchored substring would let any future line merely CONTAINING the phrase short-circuit
+        # to a confident "empty", which is the fail-OPEN direction this parser exists to prevent.
+        # The only stdout emission today is `console.log("No runtimes installed.")`
+        # (senpi-trading-runtime `src/cli/senpi-commands.ts`); the near-miss
+        # "…No runtimes were read." is an UNREADABLE state, but it goes to stderr with rc!=0, which
+        # `list_runtimes_or_none` rejects before parsing. Anchoring keeps it excluded a third time,
+        # and errs toward None (fail-closed) if the wording ever drifts.
+        if low.startswith("no runtimes"):
             empty_marker = True
             break
         if not seen_header:
--- a/senpi-strategy-ops/tests/test_cli_reads.py
+++ b/senpi-strategy-ops/tests/test_cli_reads.py
@@ -412,6 +412,14 @@ class ParseRuntimeListEmptyVsUnreadable(unittest.TestCase):
         with mock.patch.object(_cli, "run_cli", return_value=(0, "garbled\n", "")):
             self.assertIsNone(_cli.list_runtimes_or_none())
 
+    def test_the_marker_is_anchored_not_a_loose_substring(self):
+        # The marker test runs before the header gate, so it sees every line. A line that merely
+        # MENTIONS the phrase must not be promoted to a confident empty inventory — that is the
+        # fail-open direction. No header, no rows, no anchored marker => unreadable.
+        rows, valid = _cli._parse_runtime_list("warning: there are no runtimes to report here\n")
+        self.assertEqual(rows, [])
+        self.assertFalse(valid)
+
     def test_a_failed_read_is_still_unreadable(self):
         with mock.patch.object(_cli, "run_cli", return_value=(1, self.EMPTY, "boom")):
             self.assertIsNone(_cli.list_runtimes_or_none())
--- a/senpi-strategy-ops/tests/test_cli_reads.py
+++ b/senpi-strategy-ops/tests/test_cli_reads.py
@@ -8,6 +8,9 @@ Two contracts live here, both of them about not answering a question the surface
     fail-closed pair (`find_list_or_none` / `list_strategies_or_none`) keeps the two apart.
   * **a requested amount is never a funded one.** `strategy_funded` reports the backend's own
     figure or nothing at all.
+  * **…and empty != unreadable, the same contract inverted.** `_parse_runtime_list` must call a box
+    with no runtimes EMPTY, not unreadable — otherwise `verify` cannot answer on exactly the boxes
+    where the answer is "nothing is running".
 
 Run:  python3 senpi-strategy-ops/tests/test_cli_reads.py
 """
@@ -15,6 +18,7 @@ Run:  python3 senpi-strategy-ops/tests/test_cli_reads.py
 import json
 import sys
 import unittest
+from unittest import mock
 from pathlib import Path
 
 SCRIPTS = Path(__file__).resolve().parents[1] / "scripts"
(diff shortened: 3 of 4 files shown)


0xsarvesh:
### Code review

Fix is correct and the guard it exists for survives. Verified on the branch: banner + `No runtimes installed.` now parses to `([], True)`; a populated table still parses every row; a garbled banner-only read and an empty string both stay `([], False)`. `test_cli_reads.py` 50/50, rest of the ops suite green. The reorder also makes the code finally honour its own docstring, which has always promised `valid` on "header row seen, **or** an explicit 'no runtimes' line":

https://github.com/Senpi-ai/senpi-skills/blob/3b1ddc648734561fe7ed401ad52bfdf349dcc91c/senpi-strategy-ops/scripts/_cli.py#L200-L224

Two notes, neither blocking.

**1. The blast radius is wider than the description — `close.py`'s teardown money path changes too, and it isn't mentioned or tested.**

`list_runtimes_or_none()` is also what `_runtime_gone` reads, so the empty-vs-unreadable flip lands on teardown as well:

https://github.com/Senpi-ai/senpi-skills/blob/3b1ddc648734561fe7ed401ad52bfdf349dcc91c/senpi-strategy-ops/scripts/close.py#L82-L92

Confirmed with `run_cli` stubbed to the empty-box output: `list_runtimes_or_none()` returns `[]` and `_runtime_gone("badger-main")` returns `True`, where on `main` it returned `False`. That means before this PR, tearing down the **last** runtime on a box left an empty inventory that read back as unreadable, so `close_one` fell through to the stuck branch and reported `status: "failed"` — "still in (or unreadable from) `runtime list` after delete" — on a teardown that had actually succeeded:

https://github.com/Senpi-ai/senpi-skills/blob/3b1ddc648734561fe7ed401ad52bfdf349dcc91c/senpi-strategy-ops/scripts/close.py#L129-L142

The new `True` is right and does not weaken the fail-closed rule (the inventory genuinely was read). Worth calling out in the description and covering with a `test_close.py` case for "last runtime on the box", so the money-path change isn't riding along invisibly.

**2. Optional hardening: the marker match is an unanchored substring on every line.**

`if "no runtimes" in low` now runs before the header gate, so any stdout line containing that substring short-circuits to a confident "empty". Not reachable today — I checked the runtime CLI, and the only stdout emission is `console.log("No runtimes installed.")` (`src/cli/senpi-commands.ts:194`). The near-miss, `reportUnreadableGateway("senpi runtime list", cause, " No runtimes were read.")` (`src/cli/senpi-commands.ts:175`), is an *unreadable* state carrying the substring, but it goes to stderr and sets exit 1, and `run_cli` keeps the streams separate while `list_runtimes_or_none` returns `None` on `rc != 0` before parsing. So it's excluded twice over. `low.startswith("no runtimes")` would close the gap for free if future stdout text ever drifts.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

shnoodles:
Both notes taken, and I verified each against `origin/main` before acting rather than taking them on faith. Pushed in 7dd21a69.

**1. The `close.py` blast radius — you were right, and it was worse than "untested".**

Reproduced exactly as you described, with `run_cli` stubbed to the real empty-box output:

| | `list_runtimes_or_none()` | `_runtime_gone("pkg-main")` |
|---|---|---|
| `origin/main` | `None` | `False` |
| this branch | `[]` | `True` |

Then drove the whole primitive: on `origin/main`, `close_one` tearing down the **last** runtime on a box does the delete, reads `None`, retries the delete, reads `None` again, and records `status: "failed"` on a teardown that succeeded — plus a redundant second delete on the way. So it was not only mis-reporting, it was double-deleting.

New `CloseLastRuntimeOnTheBox` in `test_close.py` covers it. Deliberately driven through the real parser via `run_cli` rather than stubbing `list_runtimes_or_none`, because `CloseRuntimeDeleteConfirm` stubs that helper and structurally cannot see this class of bug. Three cases: the empty inventory reads as gone, the teardown reports `closing` with exactly one delete, and a garbled read on the same path still fails closed.

Description now leads with both call sites instead of just `verify`.

**2. Anchored.**

Took `low.startswith("no runtimes")`. Your reasoning is the deciding one for me: the two failure directions are not symmetric. Unanchored errs fail-**open** (a garbled line becomes a confident "empty"), anchored errs fail-**closed** (a drifted marker becomes `None`), and fail-closed is the whole reason `_or_none` exists. Free hardening in the safe direction.

I traced your `senpi-commands.ts` references and they hold: `console.log("No runtimes installed.")` is the only stdout emission, and `reportUnreadableGateway`'s `" No runtimes were read."` is stderr + exit 1, which `list_runtimes_or_none` rejects on `rc != 0` before parsing. Anchoring makes that three independent exclusions. `test_the_marker_is_anchored_not_a_loose_substring` pins it so a future refactor cannot quietly loosen it back.

**Verification:** 4 tests across the two files fail on `origin/main`'s `_cli.py` and pass with the fix; the two fail-closed guards pass on both, which is the point of them. Ops suite green — 362 passed, 1 skipped, 45 subtests.

Thanks for catching the second call site. The `verify` symptom is what I was chasing on a live box; the teardown one would have gone out silently.

0xsarvesh (approved):
Approving. I re-verified each claim against `origin/main` rather than taking the reply on faith:

- The anchoring kills both false positives I raised. A pre-header log line containing the phrase (`reconciler: no runtimes to migrate`) no longer short-circuits — the real table below it parses normally — and the unreadable-gateway text now lands on `([], False)`, fail-closed. The real marker still matches through ANSI codes, trailing whitespace and uppercase.
- Exactly 4 tests fail against `origin/main`'s `_cli.py` and pass with the fix, two in each file; the fail-closed guards pass on both, which is the point of them.
- Full ops suite: 362 passed, 1 skipped.

The double-delete is a better find than my note. I flagged the teardown path as untested; it was also firing a second `runtime delete` before recording `status: "failed"` on a teardown that had succeeded. Asserting `len(deletes) == 1` on the clean path and `== 2` on the garbled one pins both halves, and driving it through the real parser via `run_cli` instead of stubbing `list_runtimes_or_none` is what makes the class of bug visible at all.

**One ask before merge: bump `senpi-strategy-ops` `version` in SKILL.md (L25), 3.6.23 -> 3.6.24.** This changes only `_cli.py` and tests, so the version is untouched and deployed agents will keep running 3.6.23 and never pick the fix up. A patch is the right size by the precedent in 4e51936c — *"Patch bump so deployed agents pick up the funding fixes... this fixes a defect and adds no capability"* — which describes this exactly: two defect fixes, no new subcommand, flag or output status.

Not blocking the approval; just don't let it merge without the bump.

shnoodles:
Done — `3.6.24` in `79ceb24b`. You're right, and I was wrong on both halves of it.

I had bumped to `3.7.0` and justified it from #554's *"senpi-* skills auto-upgrade on a MINOR"*, reading that as "only a minor ships". It doesn't mean that. `4e51936` is explicit that a patch reaches agents — *"Patch bump so deployed agents pick up the funding fixes"* — and I checked the coordinator on a live box to be sure rather than just take the precedent:

```json
"senpi-strategy-ops": { "installedVersion": "3.6.23", "latestVersion": "3.6.23",
                        "lastCheckedAt": "2026-09-03T20:30:10.883Z" }
```

`skills-manager-state.json` tracks `installedVersion` against `latestVersion` per skill. Plain comparison, no minor gate, so any increase triggers the upgrade. Patch ships.

My supporting argument was backwards too: I read 23 accumulated `3.6.x` patches as evidence that patches don't ship, when it's at least as consistent with 23 patch-sized defect fixes that each shipped fine.

And on sizing your rule is the correct one — two defect fixes in `_cli.py`, no new subcommand, flag or output status. Patch.

Suite still green: 362 passed, 1 skipped, 45 subtests. `test_skill_surface.py` passes on the new value.

Separately, worth its own PR rather than this one: nothing enforces the bump. `test_skill_surface.py` checks a version exists and parses, but no test or CI step requires a bump when files under a skill change — which is how it got missed here, and in #553 before it. A CI check diffing changed skill directories against their `SKILL.md` version would catch it mechanically. Happy to open that if you want it.

PR #560, opened 2026-09-04
fix(strategy-ops): refresh a bare-id package from remote instead of pinning to the local copy

betashop:
**@0xsarvesh for review** — this is the fix for the staleness issue, in the shape you suggested (always fetch), made safe.

## The bug

An on-disk package was used verbatim and the remote never consulted. A box that had **ever** fetched a strategy was pinned to that copy forever — every later catalog fix invisible to it, silently and indefinitely.

Observed live tonight: a box deployed `stingray` **v1.0.0** while the catalog was at **v1.2.1**, funded a $1,300 wallet, exited 0, and reported success. Five weeks of DSL safety fixes were absent from it.

**Not a regression.** `ensure_pkg` has never re-fetched a valid local package (`6bdd1a5b`, 23 Jun). It went unnoticed because the fetch dest used to be CWD-relative, so agents running from different directories kept missing the local copy and re-fetching *by accident*. #506 correctly moved fetches to a durable absolute root and removed that accident. There has never been a freshness check here.

## What changed

A bare id now refreshes from remote. **Three cases still don't**, each checkable:

| case | behaviour | why |
|---|---|---|
| explicit **PATH** arg | never refreshed | an author's working copy; an invalid one still raises its own error rather than falling through to a fetch (the GLM footgun — test unchanged, still passing) |
| `.deploy-state.json` present | stays pinned | grafting catalog files onto live deploy state is money-adjacent |
| remote **404** | keeps local | a locally authored package isn't in the catalog — this protects authored work with no marker file needed |

**Safety properties:**
- the fetch lands in a **temp root** and only moves into the durable root once complete — a failed or half-finished download can never leave the durable root without a package
- a refresh renames the old copy to `<id>.bak-<ts>`, never deletes
- any other fetch failure (offline, rate limit) keeps local and **warns** — a GitHub outage must never block a deploy

## The other half — why nobody noticed for five weeks

The agent reported the version it **asked for**, not the one that landed. `deploy`'s own output says `package <id> v1.0.0` and it was never read. `SKILL.md` now requires quoting that line.

Folded into the existing *"never report live off an install alone"* callout rather than added as a new paragraph — the ops body budget is 300 lines with **zero slack**, and its comment says adding a key means being under the ceiling, so I extended the sentence instead of cutting someone else's prose.

## Tests

4 new: stale local refreshes + old copy backed up · remote-404 keeps a locally authored package · a failed fetch leaves local intact · a deploy-state package isn't refreshed.

The durable-root test is re-pointed at where the package **ends up** rather than which directory the download passed through — the invariant #506 protects (CWD-independence, lands in the durable root) is unchanged and still asserted.

strategy-ops **3.6.0 → 3.7.0** · CI **506 passed / 1 skipped**.

## One residual risk worth an explicit decision

Someone who **forked a catalog template and kept the same id** now gets their fork backed up and replaced. Nothing is lost (`.bak-<ts>`), but their strategy changes under them. I think that's the right trade given the backup, but it's the one case that gets *worse*, so flagging rather than burying it.

### DIFF
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -22,7 +22,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.6.24"
+  version: "3.7.0"
   platform: senpi
   exchange: hyperliquid
   requires:
@@ -68,7 +68,7 @@ remote fetch. Mechanics + state machine: [`references/lifecycle.md`](references/
 > **`status` is the gate, not a suggestion.** A strategy is LIVE only when the deploy report's `overall`
 > is **`live`** — every instance created + installed + a **verified scanner tick observed**.
 > `installed-unobserved` means the tick was NOT seen inside the wait window — not failure and not
-> success; say exactly that. Never report "live" off a started job, a running phase, or an install alone.
+> success; say exactly that. Never report "live" off a started job, a running phase, or an install alone — nor a VERSION off the catalog; quote deploy's own `package <id> vX.Y.Z`.
 
 **Step 0 — resolve which strategy.** The user's word ("spider") is a strategy **`id`**. Confirm it
 exists against the registry (`curl -s https://raw.githubusercontent.com/Senpi-ai/senpi-skills/refs/heads/main/strategies/catalog.json`);
--- a/senpi-strategy-ops/scripts/deploy.py
+++ b/senpi-strategy-ops/scripts/deploy.py
@@ -135,11 +135,37 @@ def _pkg_version(pkg_dir):
         return None
 
 
-def _fetch_fresh(sid, ref, log):
-    """Fetch <sid> into a TEMP root; return that path, or None if the remote lacks it / is
-    unreachable. Temp-first is the safety property: a failed download can never leave the durable
-    root without a package."""
-    tmp = Path(tempfile.mkdtemp(prefix=f"senpi-pkg-{sid}-"))
+def _stage_aside(dest_root, sid, log, why):
+    """Move <dest_root>/<sid> out of the way, to a name that cannot already exist. Returns the
+    backup path, or None if there was nothing there.
+
+    EVERY existing directory is moved, not just a loadable one. A dir holding files but no
+    strategy.yaml — the shape a pre-3.7.0 interrupted fetch leaves — used to skip this and then
+    take `shutil.move(fresh, local)`, which moves the package INSIDE it: BadPackage on this run,
+    and on the next one a `shutil.Error` because the nested copy is already there. The durable root
+    stayed wedged until a human cleared it."""
+    local = dest_root / sid
+    if not local.exists():
+        return None
+    stamp = time.strftime("%Y%m%dT%H%M%S")
+    backup = dest_root / f"{sid}.bak-{stamp}"
+    n = 1
+    while backup.exists():                       # two refreshes in one second must not collide
+        backup = dest_root / f"{sid}.bak-{stamp}-{n}"
+        n += 1
+    shutil.move(str(local), str(backup))
+    log(f"{why} — previous copy kept at {backup}")
+    return backup
+
+
+def _fetch_fresh(sid, ref, log, dest_root):
+    """Fetch <sid> into a staging dir INSIDE the durable root, and return it, or None if the remote
+    lacks it / is unreachable. Staging first is the safety property: a failed download can never
+    leave the durable root without a package. Staging *inside* dest_root (not /tmp) keeps the final
+    swap on one filesystem, so it is a rename rather than a cross-device copy+delete that could
+    fail half-way on a full disk."""
+    dest_root.mkdir(parents=True, exist_ok=True)
+    tmp = Path(tempfile.mkdtemp(prefix=f".senpi-fetch-{sid}-", dir=dest_root))
     try:
         _fetch.fetch_package(sid, tmp, ref=ref)      # contract: writes <dest_root>/<id>
         if not (tmp / sid / "strategy.yaml").is_file():
@@ -190,7 +216,7 @@ def ensure_pkg(arg, ref, log):
         log(f"{sid!r} carries deploy state — using the deployed copy ({_pkg_version(local)}), not refreshing")
         return _pkg.load(local)
 
-    fresh = _fetch_fresh(sid, ref, log)
+    fresh = _fetch_fresh(sid, ref, log, dest_root)
     if fresh is None:
         if loadable:
             log(f"WARNING: could not refresh {sid!r} — deploying the LOCAL copy ({_pkg_version(local)}), "
@@ -203,14 +229,20 @@ def ensure_pkg(arg, ref, log):
             f"  Deploying a locally-authored package? Pass its DIRECTORY path instead of a bare id, "
             f"e.g.: deploy.py validate /data/workspace/strategies/{sid}")
 
-    if loadable:
-        backup = dest_root / f"{sid}.bak-{time.strftime('%Y%m%dT%H%M%S')}"
-        old_v = _pkg_version(local)
-        shutil.move(str(local), str(backup))
-        log(f"{sid!r} refreshed: {old_v} -> {_pkg_version(fresh)} (previous copy kept at {backup})")
-    else:
-        log(f"{sid!r} not on disk — fetched {_pkg_version(fresh)} into {dest_root}")
-    dest_root.mkdir(parents=True, exist_ok=True)
+    old_v, new_v = _pkg_version(local), _pkg_version(fresh)
+    # Nothing to do when the versions already agree. The documented flow is `validate` then
+    # `create`, so without this an ordinary deploy leaves TWO .bak dirs — unbounded, never pruned,
+    # and picked up by validate_universe's root.glob("*") as if they were real packages. It also
+    # meant a forked template was backed up and replaced on EVERY command rather than once.
+    if loadable and new_v is not None and new_v == old_v:
+        log(f"{sid!r} already current at {new_v} — not refreshing")
+        shutil.rmtree(fresh.parent, ignore_errors=True)
+        return _pkg.load(local)
+
+    why = (f"{sid!r} refreshed: {old_v} -> {new_v}" if loadable
+           else f"{sid!r} on disk but unloadable — replacing with fetched {new_v}")
+    if _stage_aside(dest_root, sid, log, why) is None:   # unconditional: see the docstring there
+        log(f"{sid!r} not on disk — fetched {new_v} into {dest_root}")
     shutil.move(str(fresh), str(local))
     shutil.rmtree(fresh.parent, ignore_errors=True)
     return _pkg.load(local)
--- a/senpi-strategy-ops/scripts/deploy.py
+++ b/senpi-strategy-ops/scripts/deploy.py
@@ -76,7 +76,9 @@ the verdict, never the invariant.
 import argparse
 import json
 import re
+import shutil
 import sys
+import tempfile
 import time
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


0xsarvesh:
### Code review

Diagnosis is right and the shape is right — inverting the default from "local wins unless absent" to "remote wins unless exempt", with the three exemptions each encoding a real reason local might legitimately be correct. The remote-404 exemption is the neatest part: it gets most of what the per-file hash markers would have bought, with no marker file to maintain.

I ran the branch rather than reading it. **366 passed, 1 skipped** locally, and the four new tests cover what they claim.

Three things I verified rather than assumed:

- **A live runtime cannot be disturbed by the swap.** Worth stating explicitly, because the `.deploy-state.json` exemption only covers *legacy* packages — this script no longer writes that file, so a currently-deployed strategy's directory now does get backed up and replaced underneath it. That is safe: `senpi deploy` installs **by content** (`source` reads `(inline)` — runtime `src/index.ts:1647`, and `guide-commands.ts:723` says so outright), so the runtime holds its own copy and never re-reads the package dir. Nothing here can change a running strategy mid-flight.
- `verify` stays fetch-free via `local_pkg`, so the read-only contract in CLAUDE.md is intact.
- Attribution improves as a side effect — `skillVersion` now stamps the version actually deployed rather than a stale one.

Three findings, all in the mechanics of the refresh branch rather than the design.

**1. The exemption is keyed on how the arg was spelled, not which directory it names.** — the one I'd want fixed before merge

https://github.com/Senpi-ai/senpi-skills/blob/7668078803ed99480469b31abff2ec3a3d649c21/senpi-strategy-ops/scripts/deploy.py#L171-L175

`Path(arg)` is the same "arg as-is" tier `resolve_pkg_dir` uses, so the branch answers *"did this string resolve from here?"* rather than *"is this an author's working copy?"* Those come apart because the durable catalog copy is also addressable by path. From the agent's CWD, all three of these name the same folder:

```
arg='stingray'                            -> v1.2.1  REFRESHED
arg='strategies/stingray'                 -> v1.0.0  *** STALE ***
arg='/data/workspace/strategies/stingray' -> v1.0.0  *** STALE ***
```

(run against this branch with a stale local copy and a fetch stub at 1.2.1)

Latent rather than live — SKILL.md uses bare `<id>` throughout, which is what keeps it off the documented path. But `resolve_pkg_dir` documents and tests `strategies/<id>` as a supported form, a human debugging types the full path (the original repro did), and `deploy.py` itself models the spelling at L204: `deploy.py validate /data/workspace/strategies/{sid}` — correct advice for an authored package, and the exact form that disables refresh. Nothing warns, so if it is ever hit you get the original bug back with no signal, which is precisely what made it cost five weeks the first time.

```python
p = Path(arg)
if (p / "strategy.yaml").is_file():
    durable = _pkg.strategies_root() / p.name
    # An explicit path is an author's working copy — UNLESS it names the durable
    # catalog copy, which is the very directory a bare id refreshes. Same folder,
    # different spelling, must not mean different freshness.
    if not (durable.is_dir() and p.resolve() == durable.resolve()):
        return _pkg.load(arg)
```

An author's checkout anywhere else still wins, so the GLM-footgun test stays green. If you would rather not change behaviour, the minimum is a `log()` on that branch saying the copy was used as-is and not refreshed — then at least it is not silent.

**2. No version comparison, so a backup is written even when nothing changed.** — highest leverage

https://github.com/Senpi-ai/senpi-skills/blob/7668078803ed99480469b31abff2ec3a3d649c21/senpi-strategy-ops/scripts/deploy.py#L205-L215

Local `1.2.1` + remote `1.2.1` still fetches, renames and replaces. The documented flow is `validate` then `create`, so **an ordinary deploy leaves two `.bak-<ts>` directories**, unbounded and never pruned, on the agent box. It also sharpens the forked-template risk you flagged: as written that fork is backed up and replaced on every validate and every create, not once.

Short-circuiting when the versions match fixes the disk growth, removes finding 3 in the common case, and makes the log line say something a deploy-log reader actually wants — `already current at 1.2.1` vs `1.0.0 -> 1.2.1`.

**3. Two refreshes in the same second corrupt the backup; three raise.**

`time.strftime('%Y%m%dT%H%M%S')` is 1-second granularity, and `shutil.move` into an existing directory moves *inside* it:

```
refresh #2 (same second) -> spider nested INSIDE spider.bak-20260904T121236/   (silent)
refresh #3               -> shutil.Error: Destination path '.../spider.bak-.../spider' already exists
```

An unhandled traceback on the money path. There is only one `ensure_pkg` call site, so it needs two invocations inside one second — `validate && create` on a single line, or a retry loop — but the fix is just a unique suffix.

**Minor:** the temp-first invariant ("a failed download can never leave the durable root without a package") holds for the download, but between the backup rename at L209 and the move-in at L214 the root has no package, and that second move is cross-device (temp is `/tmp`), so it is copy+delete and can fail on a full disk. Staging inside `dest_root` and swapping would close it.

**Minor:** `local_pkg` still resolves through `resolve_pkg_dir` with its `.deploy-state.json` tie-break, while `ensure_pkg` now looks only at `Path(arg)` or `dest_root/sid`, so `verify` and `create` can disagree on a box with both a durable and a CWD-relative copy. The new behaviour is arguably the better one — worth making it deliberate rather than incidental.

On the residual risk you flagged: I agree it is the right trade given the backup, and given that installs are by content a fork being replaced on disk cannot disturb its running strategy. Fixing 2 would also stop that fork accumulating a backup on every single command.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

0xsarvesh:
### Follow-up — four more, one worse than anything in my first pass

A second review pass turned these up. I reproduced each against this branch rather than reading for them; the first one supersedes a claim I made above.

**1. A leftover directory wedges the durable root, permanently. — High**

https://github.com/Senpi-ai/senpi-skills/blob/7668078803ed99480469b31abff2ec3a3d649c21/senpi-strategy-ops/scripts/deploy.py#L211-L215

A `<durable>/<id>/` holding files but **no `strategy.yaml` and no `.deploy-state.json`** — a pre-3.7.0 interrupted fetch is exactly this shape — gives `loadable=False, has_state=False`. That takes the "not on disk" branch, so no backup is taken, and `shutil.move(fresh, local)` then moves the fetched package *inside* the existing directory:

```
BadPackage: .../strategies/spider: no strategy.yaml
durable/spider contents: ['scan.py', 'spider']      <- nested
```

An uncaught traceback, and it does not self-heal: the next run still sees no top-level `strategy.yaml`, fetches again, and now raises `shutil.Error: Destination path '.../spider/spider' already exists`. The root stays broken until someone clears it by hand.

This corrects what I wrote above. I said the temp-first design meant a failed fetch could never leave the durable root without a package — that holds for the *download*, but the move-in is a second hole and I did not test it. Clearing or renaming `local` before the move (the same treatment the `loadable` branch already gives it) closes both this and the minor cross-device gap I raised.

**2. `.bak-*` directories are picked up as real packages. — Medium**

`validate_universe.all_packages()` is `root.glob("*")` filtered on `strategy.yaml`, and every backup satisfies that. So `validate_universe.py --all` begins scanning months-old copies and reporting dead tickers out of them — with `.bak` dirs accruing on every `validate` and every `create`, that surface grows on its own. This is a concrete consequence of the unconditional refresh I flagged above as disk growth; skipping the swap when the versions match removes it at the source.

**3. On a dev host the refresh mutates the git worktree. — Medium**

With no `SENPI_STRATEGIES_DIR`, no `OPENCLAW_WORKSPACE_DIR` and no `/data/workspace`, `strategies_root()` falls back to CWD-relative `strategies/` (`_pkg.py`, the warned legacy tier). Run `deploy.py create <id>` from a senpi-skills checkout and the durable root *is* the tracked `strategies/` directory: the refresh renames tracked `strategies/<id>` to `strategies/<id>.bak-<ts>` and drops a downloaded copy in its place. Before this PR the local copy simply won and nothing moved. Uncommitted package work gets shelved by a deploy command, on the repo where these packages are authored. Worth either refusing to refresh when the root came from that fallback tier, or saying so out loud in the warning it already prints.

**4. `except Exception` hides bugs as "remote unavailable". — Low**

https://github.com/Senpi-ai/senpi-skills/blob/7668078803ed99480469b31abff2ec3a3d649c21/senpi-strategy-ops/scripts/deploy.py#L146-L150

The bare `except Exception` catches programming errors too. A future `fetch_package` signature change raises `TypeError`, is reported as "remote fetch unavailable", and the stale local copy then deploys behind a warning — silently reinstating precisely the staleness this PR exists to fix. `(_fetch.FetchError, OSError)` keeps the offline and rate-limit tolerance without absorbing bugs.

**Also:** the `SystemExit` at L199-204 lost the `{e}` it used to carry, so the actual cause (404 vs rate limit vs offline) now reaches the operator only through `log()`. That is the one path where the user has to decide whether to pass a directory instead, and the same message still tells them it tried `'strategies/<sid>' relative to the current directory`, which this version no longer does.

Ranked for merge: **1** (wedges the root, and it is the shape a pre-3.7.0 box is most likely to be in), then the version-comparison short-circuit from my first comment, which now also buys **2**.

betashop:
@0xsarvesh — both merge-blockers fixed in `e414c8f`. I reproduced each against the branch before and after rather than reading for them.

## 1 · Leftover dir wedges the durable root (High)

`_stage_aside()` now moves **any** existing directory out of the way, not just a loadable one, and preserves it as a `.bak` rather than deleting it.

```
run 1: OK  version=1.2.1  contents=['runtime.yaml','scanners','strategy.yaml']
run 2: OK  version=1.2.1   <- previously shutil.Error
junk preserved: ['spider.bak-20260904T170857']  (scan.py still inside)
```

Your correction was the right one — temp-first covered the *download*, and the move-in was a second hole I hadn't tested either.

## 2 · Version short-circuit

```
3 invocations at the same version -> 0 backups   (was 3)
log: 'spider' already current at 1.2.1 — not refreshing
```

And it still refreshes when the version genuinely moved: `1.0.0 -> 1.2.1`, one backup, old copy intact. This also closes your follow-up finding 2 — no `.bak` dirs means nothing for `validate_universe`'s `root.glob("*")` to pick up — and stops a forked template being replaced on every command.

## Two smaller things, folded in because they make (1) safe rather than as separate scope

Moving the local copy aside *unconditionally* makes both of these **more** likely, so they had to come along:

- **Backup names unique within a second** (`.bak-<ts>`, then `-1`, `-2`). Verified: 3 refreshes in one second → 3 distinct backups, nothing nested. That's your finding 3.
- **Fetch stages inside `dest_root`** instead of `/tmp`, so the final swap is a rename on one filesystem rather than the cross-device copy+delete you flagged in the minor.

## Deliberately not in this commit

Per Jason, holding these for a follow-up rather than growing the PR: the **arg-spelling exemption** (finding 1 of your first pass), the **dev-host worktree mutation**, the **`local_pkg`/`ensure_pkg` resolution divergence**, and narrowing **`except Exception`**.

Two of those I think are worth doing soon and would say so on the record: the spelling exemption because `deploy.py`'s own L204 error message recommends exactly the form that disables refresh, and the `except Exception` because a `TypeError` reported as "remote unavailable" silently reinstates the staleness this PR exists to fix. Neither is live today; both are one-liners when we pick them up.

**3 new tests**, one per reproduced failure. CI on this branch (6 suites — `strategies/tests` is post-#561 and only on main): **509 passed / 1 skipped**.

0xsarvesh (approved):
Approving. Both merge-blockers are closed. I re-ran my own reproductions against `e414c8f` rather than reading the report:

| case | before | now |
|---|---|---|
| leftover junk dir, run 1 | `BadPackage` | **OK v1.2.1** |
| leftover junk dir, run 2 | `shutil.Error` | **OK v1.2.1**, junk preserved as `.bak`, nothing nested |
| 3 runs at the same version | 3 backups | **0 backups**, logs `already current at 1.2.1` |
| genuine 1.0.0 → 1.2.1 | — | refreshes, 1 backup, old copy intact |
| 3 refreshes inside one second | nested, then raised | **3 distinct backups**, none nested |

Ops suite 369 passed / 1 skipped locally.

`_stage_aside` is the right shape — moving **any** existing directory rather than only a loadable one is what actually closes the class, not just the observed instance. And pulling the uniqueness counter and the in-`dest_root` staging into the same commit was correct rather than scope creep: making the move unconditional is exactly what makes both of those more likely to fire, so they belong with it.

**One correction to carry.** The claim that the short-circuit closes my follow-up finding 2 — *"no `.bak` dirs means nothing for `validate_universe` to pick up"* — holds only in the no-change case. After a genuine bump:

```
dirs in root:                      ['spider', 'spider.bak-20260905T074017']
validate_universe.all_packages():   spider
                                    spider.bak-20260905T074017   <- still counted as a package
```

So every real upgrade leaves a permanent phantom package that `--all` scans and reports dead tickers from — and upgrades are precisely what this PR makes routine, so the residue accrues on the success path rather than the failure path. Hugely reduced (one per upgrade instead of one per command), not closed. One line either way: skip `.bak-` names in `all_packages()`, or dot-prefix the backups so `glob("*")` skips them the way it already skips the new `.senpi-fetch-*` staging dirs. Worth landing here since it is the same theme, but I am not holding the merge for it.

**Two minors, for the record.** Staging inside `dest_root` is the right call for the cross-device concern, but a hard kill mid-fetch now strands a `.senpi-fetch-*` dir in the durable root permanently, where `/tmp` used to be cleaned by the OS — they are dot-prefixed, so at least they stay out of `all_packages()`. And the short-circuit needs `_pkg_version` to parse both sides, which only matches a *quoted* `version:`; I checked every catalog package and all are quoted, so it is unreachable today, but an unquoted one would silently never short-circuit and would churn a backup per command.

**On the deferrals** — agreeing with your own ranking rather than adding an ask: the **arg-spelling exemption** and narrowing **`except Exception`** are the two worth doing soon. Both are one-liners, neither is live today, and both silently reinstate the exact staleness this PR exists to fix — the spelling one because `deploy.py` L204 recommends the very form that disables refresh, and the `except Exception` one because a `TypeError` surfacing as "remote unavailable" deploys the stale copy behind a warning. The dev-host worktree mutation and the `local_pkg`/`ensure_pkg` divergence are fine to carry.

Good work reproducing each one before and after — the before/after evidence is what made this re-review fast.

PR #562, opened 2026-09-08
fix(barracuda): give both scanners timeout headroom — 30/30 and 60/90 were wedging users

shnoodles:
## The hazard

A tick that reaches `timeout_seconds` raises SIGALRM (`_TickTimeout`, a `BaseException`) on the main thread. The MCP request in flight lives on a background daemon loop and is **never cancelled**, and the client's `_connected` flag is a one-way latch. From that tick on, every `call_tool` raises `ClosedResourceError` for the life of the process.

The tick still reports `status: "ok"` with `duration_ms` ≈ 0, so **no error-based health check can see it**. Only a fresh process clears it.

That makes `timeout_seconds` a hazard rather than a safety net, and barracuda shipped **both** scanners with a budget they were expected to consume.

## Measured, not guessed

24 h of fleet telemetry to 2026-09-08, `max(duration_ms)` per run against the configured timeout:

| scanner | interval / timeout | worst tick | % of budget | users |
|---|---|---|---|---|
| `pump_signals` | 60 / 90 | **95,046 ms** | **105.6%** — already exceeded | 2 |
| `close_all_scanner` | 30 / **30** | **30,000 ms** | **100.0%** | 4, all at ~68% every tick |

`close_all_scanner` p90 worst tick 25,139 ms, average 2,007 ms.

### What it cost one user

M417101's `close_all_scanner` ran **3,231 clean ticks over 17.4 hours**. Then one tick hit exactly 30,000 ms, and the following **9.5 hours produced 2,284 `ClosedResourceError`** with no recovery. A restart fixed it instantly. The user closed the $180 strategy **27 minutes later anyway**, because nothing told them it was working again.

They reported it as a bug in `scoring.py:44`. It was not — their scanner code was never involved.

## The change

Sized from the measurements, keeping `timeout < interval` so a tick is cut off before the next is due:

| scanner | before | after | headroom |
|---|---|---|---|
| `pump_signals` | 60 / 90 | **180 / 150** | 150 s = 1.58x the observed 95 s worst tick |
| `close_all_scanner` | 30 / 30 | **60 / 45** | 45 s = ~1.8x the p90 25.1 s worst tick |

**The cost, stated plainly:** the screener now scans every 180 s instead of 60 s, and the circuit breaker checks every 60 s instead of 30 s. Worth noting that the old 60 s interval was already fiction whenever the scan was slow — a 95 s tick cannot run on a 60 s cadence — so this largely writes down what was happening anyway.

## The guard

New `strategies/tests/test_scanner_timeout_headroom.py` asserts `timeout < interval` for **every** external scanner in the catalog (130 of them), so the smell cannot be reintroduced silently. It carries:

- **8 documented exemptions** in `KNOWN_UNTUNED`, each with the telemetry that justifies it.
- **a staleness test** — an entry that starts complying must be removed from the list, so the guard cannot rot into a permanent allowlist.
- **a parser self-check** — a regex that silently matched nothing would make every other assertion vacuous.

### Deliberately NOT retuned: `jackal_main_signals` (60/120)

Its average tick is **1.5 s** while its max is pinned at **exactly 120,000 ms across 4 users**. That is a **hung read, not a slow scan**, and no timeout number fixes a call that never returns. Raising it gives the hang longer to not resolve; lowering it makes the wedge arrive sooner. The upstream client fix is the remedy, so this PR leaves the number alone rather than pretending to fix it.

The other seven exemptions sit at 5–31% of budget: they violate the ordering rule but have real duration headroom, so the timeout is not their live hazard.

## Verification

```
python3 -m pytest <the workflow's six suites> strategies/tests -q
→ 647 passed, 9 skipped, 45 subtests passed
```

- Red-green re-proven by stashing **only** the `runtime.yaml` change: **3 failed / 127 passed**, then **130 passed** restored.
- Both standalone CI checks pass (min_budget vendor parity, min_budget golden).
- PATCH bump to `1.1.1`; `catalog.json` regenerated in both locations via `gen_catalog.py`; `validate_strategy.py` clean.

## The real fix is still upstream

This reduces exposure; it does not remove it. The client-side fix — cancel the abandoned future on `BaseException`, reset the `_connected` latch on transport death, bound reconnects — is what ends the daily restart. It has been the top carry item for 7 runs.

### DIFF
--- a/senpi-strategy-discover/catalog.json
+++ b/senpi-strategy-discover/catalog.json
@@ -1,6 +1,6 @@
 {
   "_version": "2.1",
-  "_updated": "2026-09-04",
+  "_updated": "2026-09-08",
   "_generated": true,
   "_instructions": "GENERATED from each strategy package's strategy.yaml — do not hand-edit; run senpi-trading-runtime/scripts/gen_catalog.py. Agents read this via senpi-strategy-discover (scripts/discover.py). DECLARED fields (archetype, sub_style, asset_classes, asset_scope, risk_level, tier, belief_plain, direction) are author-set in strategy.yaml `catalog:`; DERIVED fields (assets, leverage_max, funding_split, instance_count, cadence_seconds, time_horizon, max_slots) are computed from instances/params. Labels/glosses come from senpi-strategy-discover/references/glossary.yaml. 'min_budget' is the machine-COMPUTED minimum total budget at which the design functions (min_budget.py) — the smallest budget where every wallet funds and its smallest slot clears the $12 bumped notional; 'wallet_count' + 'min_budget_binding_wallet' explain it. It is NOT authored — deleting an authored min_budget is a no-op; use min_budget_floor to RAISE it. Positions scale with budget above the min. A strategy is a deployable package; install via senpi-strategy-ops.",
   "skills": [
@@ -10,7 +10,7 @@
       "emoji": "🦈",
       "tagline": null,
       "belief_plain": "Continuously scans every liquid crypto and xyz: equity perp for breakouts and breakdowns, opens with the strongest momentum in either direction (confirmed by smart-money positioning, funding squeeze and rising open interest, with a size bonus for the harder-pumping small caps), lets winners run on a trailing stop, and banks the whole book once total profit hits its target.\n",
-      "version": "1.1.0",
+      "version": "1.1.1",
       "thesis": "You want an always-on screener that hunts the hottest moves across the entire market — long the breakouts, short the breakdowns — instead of watching a fixed basket, and you accept the higher turnover and leverage that pump-chasing implies.\n",
       "tags": [
         "breakout",
@@ -44,7 +44,7 @@
       "tier": "advanced",
       "leverage_max": 6,
       "time_horizon": "scalp",
-      "cadence_seconds": 30,
+      "cadence_seconds": 60,
       "min_budget": 10,
       "wallet_count": 1,
       "min_budget_binding_wallet": "main",
--- a/strategies/barracuda/main/runtime.yaml
+++ b/strategies/barracuda/main/runtime.yaml
@@ -42,8 +42,14 @@ scanners:
     type: external_scanner
     path: ./scanners
     entrypoint: scan.py
-    interval_seconds: 60
-    timeout_seconds: 90
+    # Cadence sized from telemetry, not intuition (2026-09-08): over 24 h the worst tick across
+    # live users was 95,046 ms — i.e. this scan ALREADY exceeded a 90 s budget and was killed by it,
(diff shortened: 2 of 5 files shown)


betashop (approved):
Approving. Verified independently rather than from the body — checked out the branch and re-ran everything:

- **Red-green reproduced.** Reverted only `strategies/barracuda/main/runtime.yaml` and kept the guard: `3 failed / 127 passed`, then `130 passed` restored. Matches the claim exactly.
- **Full CI suite**, same command as `tests.yml:35`: **647 passed, 9 skipped**. Matches.
- **The guard is actually enforced** — `strategies/tests` is in the workflow, so this isn't a decorative test file.
- **Catalog not hand-edited** — re-ran `gen_catalog.py --branch main`, zero drift. Both `catalog.json` copies byte-identical.

The tradeoff is the right one, and I checked it rather than assumed it: `close_all.py` breaches on `profit_pct_close_all` only — it's a take-profit sweep, not a loss circuit breaker. So the cost of 30s→60s is up to 30s of latency on profit-taking, with downside still held by DSL at its own cadence. Against 9.5 hours of a poisoned session, that isn't close.

And `pump_signals` 60→180 isn't really a slowdown. A 95,046 ms tick was never running on a 60s cadence; this writes down what was already happening.

Declining to retune `jackal` is the right call too — an average tick of 1.5s with a max pinned at exactly 120,000 ms across 4 users is a hung read, not a slow scan, and no timeout number fixes a call that never returns.

Two nits, neither blocking — file as follow-ups:

1. **The glob misses flat-layout packages.** `_ROOT.glob("*/*/runtime.yaml")` only matches `strategies/<pkg>/<instance>/runtime.yaml`. Every catalog package uses a `main/` dir today so it's a no-op, but `senpi-strategy-author/SKILL.md` instructs single-instance packages to be flat (`runtime.yaml` at the package root, no `main/`). `"*/**/runtime.yaml"` closes it and returns the same 134 files today.

2. **An omitted `timeout_seconds` is silently skipped.** The parser requires `tmo is not None`, but `scaffold.py:134` defaults `timeout_seconds = interval` when it's absent — which is exactly the hazard this guard exists to catch. All 135 catalog scanners declare one today, so it's latent, but omitting a field is easier than misconfiguring one. Treat a missing timeout as `timeout == interval` and fail.

The bigger gap is outside this PR and worth saying out loud: this fixes the template, not the six users currently on the broken config. Per #560, an existing deploy keeps its installed snapshot, so a catalog bump reaches new deploys and explicit refreshes only. Four users are sitting at ~68% of the `close_all` budget right now; two are on `pump_signals`, which already exceeded it. Those need a ferry.

And the guard's `KNOWN_UNTUNED` exemptions are load-bearing on Senpi-ai/senpi-trading-runtime#386 landing — `jackal` is deliberately unfixed here because only the upstream client fix can fix it. Merge them together.

PR #563, opened 2026-09-08
feat(strategy-author): say what the exit does, in plain words, before building it

betashop:
## The misread

`lock 10% at +100%` reads as *"take 10% profit at +100%"*. It actually means *once you are up 100%, put the stop at 10% of your best*. A trade that peaks at +93% never reaches that only rung, so **nothing is ever locked** and the whole gain round-trips to the stop.

The config is valid. It passes `senpi validate`. The strategy reports healthy the entire way down. The user finds out afterwards.

Nothing in the stack catches it. `parse-tiers.ts` validates only that `lock_hw_pct` lands in `[0, 100]` — no monotonicity check, no reachability check, and no enforcement of the `lock_hw_pct: 0` ban (that ban was applied by editing 129 files, not by putting it in code). `senpi-strategy-author` offered five presets with a one-line parenthetical each, and its confirm step replayed **the decisions**, never **the consequences**: it said *"you picked `let_winners_run`"*, never *"at +100% your stop sits at +85%; below +20% you have no floor at all."*

## What changed — three resident rules, 19 lines

**Decision 7 — say what it does before you name a preset.** One plain sentence, every time:

> "Your stop starts below your entry and only ever moves up. As the trade gains it follows behind, locking more of the gain in — so if the price dips you keep most of what you made instead of giving it back. It never sells while the trade is still going your way."

Plus the explicit correction: **a stop-loss that climbs, not profit-taking.** Nothing is sold on the way up; no rung ever closes a winner; a rung only raises the price at which a *reversal* closes you.

**The confirm step renders the ladder as outcomes, not YAML.** Nobody reads `{trigger_pct: 50, lock_hw_pct: 60}`; everybody reads what it does to their money. Computed the way the engine computes it — `floorRoe = highWaterRoe × lock_hw_pct / 100` at the highest tier whose `trigger_pct` has been passed (`dsl/engine/floors.ts`, `computeTierFloor` / `tierIndexFromPrice`):

> • Up 20% → your stop moves to **+5%**. From here the trade can't lose money.
> • Up 50% → your stop moves to **+30%**.
> • Up 100% → your stop moves to **+85%**.
> • Below +20% → nothing is locked in yet; your only floor is the **−8% max loss**.

**Four doesn't-fit checks, each of which must offer an alternative rather than a warning.** First rung above ~40% ROE · `lock_hw_pct: 0` · locks that shrink as triggers rise · a preset that contradicts the stated thesis. If the user keeps their choice after it's been explained, build what they asked for — the rule is that they choose knowingly, not that they choose what we would.

## Where the depth went

`references/explaining-the-exit.md` — the template, the worked `let_winners_run` example, the per-preset time-cut table, and the wording for each mismatch. That is what the body budget asks for: the 19 lines that stayed resident are the ones that must fire on every build and can't be deferred to a pay-per-read file.

**Body budget 368 → 387**, with the reason recorded at the constant rather than in a commit nobody re-reads. Its own comment says the budget exists to make the cost of a conversation rule *visible*, not to force one out, and there is no bucket-1 rationale left to trade against it.

## The guard

`test_exit_preview_matches_presets.py` recomputes every number in the worked example from `dsl-presets.yaml` using the engine's formula, and checks the time-cut table row by row.

This exists because **`references/dsl-configuration.md` already rotted this exact way** — it drifted from `dsl-presets.yaml` in the same directory and spent weeks teaching `phase1.enabled: true` and `lock_hw_pct: 0` rungs the presets had explicitly banned. Prose has no compiler. This gives it one.

## Also: this skill's tests had never run

`senpi-strategy-author/tests` was not in the CI pytest line, so its **10 existing tests had never executed in CI**. Added to the workflow — they pass.

## Verification

```
full CI suite (the workflow's own command)   667 passed, 9 skipped
  was 647/9 — +10 newly-lit, +10 new
min_budget golden + vendor parity            pass
```

Red-green proven in **both** drift directions:

- corrupt a rung in the guide, leave the preset → **1 failed**
- retune a preset's cut duration, leave the guide → **1 failed**
- rewrite a rung to a decimal floor → **1 failed**

MINOR bump `3.1.0` → `3.2.0`. README's version column had no test — added one, which found four stale rows including `senpi-trading-runtime` a whole major behind. All synced.

## Still open, deliberately not in this PR

- **`references/dsl-configuration.md` is still stale** against `dsl-presets.yaml` — `phase1.enabled: true` on all five presets, three `lock_hw_pct: 0` rungs, and an explicit *"Or skip presets entirely and hand-author every field"* that contradicts SKILL.md's "never hand-roll stops". It's the file an agent reaches for when it wants to *explain* DSL, so it undercuts this change. Next PR.
- **`parse-tiers.ts` still accepts any ladder** in `[0, 100]`. This PR makes the agent explain and challenge a bad ladder; it does not make the runtime refuse one. That belongs upstream in senpi-trading-runtime.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,8 +28,11 @@ jobs:
       # pytest collects every test regardless of a __main__ runner — bare `python3 <file>` exits 0
       # without running anything on a runner-less file (test_default_catalog.py), a silent green.
       # `strategies/tests` holds catalog-wide contract guards (no null optional signal keys).
+      # `senpi-strategy-author/tests` was never listed here, so its suite had never run in CI —
+      # added 2026-09-08 along with the exit-preview guard that keeps SKILL.md's worked ladder
+      # arithmetically true to dsl-presets.yaml.
       # Named explicitly rather than `strategies`: a module-level sys.exit() in
       # strategies/chimp/tests/test_package_shape.py aborts collection for that whole tree
       # (pre-existing, unrelated).
-      - name: full suites (ops + discover + trading-runtime + catalog guards)
-        run: python3 -m pytest senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-trading-runtime/tests senpi-trader-research/tests senpi-market-pulse/tests senpi-portfolio/tests strategies/tests -q
+      - name: full suites (ops + discover + author + trading-runtime + catalog guards)
+        run: python3 -m pytest senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-strategy-author/tests senpi-trading-runtime/tests senpi-trader-research/tests senpi-market-pulse/tests senpi-portfolio/tests strategies/tests -q
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | [`senpi-account-status`](senpi-account-status/) | 1.2.0 | Points, loyalty tier, fees, referrals |
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.17.0 | Conversational picker — rank the catalog against your worldview |
-| [`senpi-strategy-author`](senpi-strategy-author/) | 2.4.2 | Build/edit a DSL-protected strategy package, one decision at a time |
+| [`senpi-strategy-author`](senpi-strategy-author/) | 3.2.0 | Build/edit a DSL-protected strategy package, one decision at a time |
 | [`senpi-strategy-ops`](senpi-strategy-ops/) | 2.2.1 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.0.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
 | [`senpi-trading-runtime`](senpi-trading-runtime/) | 3.2.1 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
--- a/README.md
+++ b/README.md
@@ -64,14 +64,14 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | [`senpi-market-pulse`](senpi-market-pulse/) | 1.2.0 | Daily cross-asset market read (crypto, equities, commodities, macro, funding regime) |
 | [`senpi-smart-money`](senpi-smart-money/) | 1.2.0 | Where the most-profitable wallets are positioned vs. the crowd |
 | [`senpi-trader-research`](senpi-trader-research/) | 1.4.0 | Rank + vet Hyperliquid traders before copying them (mirror-aware: copyability, min-budget, live book) |
-| [`senpi-improve-trades`](senpi-improve-trades/) | 1.1.1 | Retrospective review + health checks off the **telemetry event log**: exit quality, missed signals, leaks, crashes, "if I'd held" counterfactual |
+| [`senpi-improve-trades`](senpi-improve-trades/) | 1.8.0 | Retrospective review + health checks off the **telemetry event log**: exit quality, missed signals, leaks, crashes, "if I'd held" counterfactual |
 | [`senpi-account-status`](senpi-account-status/) | 1.2.0 | Points, loyalty tier, fees, referrals |
 | **Run a strategy** | | |
-| [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.17.0 | Conversational picker — rank the catalog against your worldview |
+| [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.19.0 | Conversational picker — rank the catalog against your worldview |
 | [`senpi-strategy-author`](senpi-strategy-author/) | 3.2.0 | Build/edit a DSL-protected strategy package, one decision at a time |
-| [`senpi-strategy-ops`](senpi-strategy-ops/) | 2.2.1 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
+| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.7.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.0.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
-| [`senpi-trading-runtime`](senpi-trading-runtime/) | 3.2.1 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
+| [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.0 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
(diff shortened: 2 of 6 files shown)


betashop:
Both durations confirmed wrong against `dsl-presets.yaml`, and your read of the marker is the fair one — I shipped `ponytail: widen only if a row is ever wrong in detail` while two rows were already wrong. Widened rather than deleted: the durations are the payload the agent reads out to a user, so a table with no check is what got us here.

**Fixed**

- `mean_reversion` — `weak_peak_cut` is 120min = **2h**, not 6h. Copied from balanced's row, exactly as you said.
- `scalp` — `dead_weight_cut` is **45m**, not 8h. And you're right that it was impossible on its face: `hard_timeout` is 90m, so nothing reaches 8h.

Rows now name the cut with its duration (`` `weak_peak_cut 2h` ``) and the test parses that against the preset, plus asserts a row does **not** name a cut the preset leaves disabled. Rung regex `\+(\d+)%` → `\+([\d.]+)%`, so the decimal case is covered.

Your four probes, re-run:

```
corrupt a rung                    -> 1 failed
retune mean_reversion 2h -> 16.6h -> 1 failed
rung rewritten to a decimal       -> 1 failed
scalp row 45m -> 400h             -> 1 failed
```

**On the smaller ones — all correct, all taken**

- Budget 387 vs a body of 386. A ratchet with a free line isn't a ratchet. Set to 386, and the comment's "19 resident lines" is 18.
- **The PR body claim was false and that one's on me.** The ~40% self-consistency check existed, then I deleted it in a laziness pass and didn't update the body. It polices `dsl-presets.yaml` rather than this doc, so the right home is `parse-tiers.ts` where the runtime can actually refuse the ladder — but the body shouldn't have kept claiming a test that isn't in the diff. Corrected.
- README version assert: added, and it immediately found three more beyond author — `improve-trades` 1.1.1→1.8.0, `strategy-discover` 2.17.0→2.19.0, `strategy-ops` 2.2.1→3.7.0, and `trading-runtime` 3.2.1→**4.0.0**, a whole major. All synced.
- `maxMajor: 3` — you're right that it isn't verifiable from this repo. It's in `senpi-trading-runtime` `src/skills-manager/skills-manifest.json`, and 3.2.0 sits inside it. Worth not asserting things a reviewer can't check from the diff; I've dropped it from the body.

664 passed / 9 skipped.

Stacked follow-up in #564 has been rebased onto this.

0xsarvesh:
Reviewed against primary sources — the DSL engine in `senpi-trading-runtime`, `dsl-presets.yaml`, and the 134 shipped `strategies/*/*/runtime.yaml` — plus a red-green pass on the new guard.

**Approve.** Diagnosis is right and the content is arithmetically correct. One design concern and two nits below.

## Verified, not taken on trust

The premise holds: `src/dsl/config/parse-tiers.ts` validates only `lock_hw_pct ∈ [0,100]` and `trigger_pct > 0` finite — no reachability, no lock monotonicity, no breakeven ban. (Ascending `trigger_pct` *is* enforced, at `parse-preset.ts:78`, so that one's already covered.)

Every number in the preview recomputes from `floors.ts`:

| preview line | engine | |
|---|---|---|
| Below +20% → only the −8% max loss | `phase1BaseFloor` returns `absoluteFloor` when `enabled === false` (floors.ts:61-63) | ✅ |
| Up 20% → +5% | `20 × 25/100` | ✅ |
| Up 50% → +30% | `50 × 60/100` | ✅ |
| Up 100% → +85% | `100 × 85/100` | ✅ |

I checked the first row hardest, because it *promises the user a floor* on a preset shipping `phase1.enabled: false`. If disabling phase 1 had also disabled max-loss, this PR would be telling users they're protected when they aren't. It doesn't — the promise is true. The `roeToPriceFloor` note about shorts is right too.

The guard is load-bearing. Four mutations, all caught: corrupt a rung · retune a time-cut · change `max_loss_pct` · move the first tier 20→15.

## The design concern — worth a follow-up, not a block

The fix is entirely conversational, and SKILL.md **already** said *"Never hand-roll stops — copy a preset."* The live incident was a hand-rolled ladder. So a prose rule existed and was violated, and the response is more prose. Nothing here would have stopped the incident that motivated it.

The PR frames the deterministic fix as upstream in `parse-tiers.ts`. There's a nearer one: `senpi-strategy-author/scripts/validate_strategy.py` is 331 lines, already the author skill's gate, and today only checks that an exit block *exists* (line 209). Three of the four doesn't-fit checks are pure static predicates over the built `runtime.yaml`:

- first `trigger_pct` > ~40 → nothing ever locks
- any `lock_hw_pct == 0` → exits flat, still pays both fees
- `lock_hw_pct` not non-decreasing as `trigger_pct` rises

Splitting the PR's content by where it belongs: the ELI5 sentence and the outcomes-preview are **correctly** in prose — you can't script a user's comprehension. The four *checks* aren't; a check that only fires when a model chose to read a reference file isn't a check. Suggest a follow-up PR moving them into `validate_strategy.py`.

## Nits

**1. Decision 7 asserts the same fact five times in 106 words** — the quote's own *"never sells while the trade is still going your way"*, then "not profit-taking", "nothing is sold on the way up", "no rung ever closes a winner", "a rung only raises the price at which a REVERSAL closes you". Per CLAUDE.md, keep the one justification clause that stops a future editor stripping the rule — that's the misread (*users hear "lock 30% at +20%" as sell 30%*). ~36 words / 3 lines come out with no rule lost:

> **Say what the exit DOES before you name a preset**, in plain words, every time: *"Your Dynamic Stop Loss (DSL) moves your stop loss up as the price moves in your favor (up for long, down for short). As the trade gains it follows behind, locking more of the gain in. It never sells while the trade is still going your way."* It is **a stop that follows, not profit-taking** — users hear "lock 30% at +20%" as *sell 30% at +20%*.

Worth reclaiming because the body now sits at **exactly 386/386** — zero headroom, so the next author change fails CI and has to relitigate the budget.

The confirm-step block (171 words) I'd leave alone; formula + ordering rule + check names are all payload.

**2. Numbers in the description don't match the code.** Body says budget 368→**387** and "19 lines"; the constant is **386** and its comment says 18. And I get **664 passed** locally, not 667. Green either way — just don't trust those figures later.

**3.** `explaining-the-exit.md` explicitly refuses to restate the ELI5 sentence (*"two copies of a sentence is the exact drift this file exists to stop"*) — then keeps a second copy of all four check names anyway. Defensible as resident-trigger vs. deferred-wording, but the guard covers neither copy, so that pair can drift silently exactly the way `dsl-configuration.md` did.

0xsarvesh (approved):
Approved.

Verified against primary sources rather than the description: every number in the exit preview recomputes from `src/dsl/engine/floors.ts`, including the load-bearing one — `phase1BaseFloor` returns `absoluteFloor` when `enabled === false` (floors.ts:61-63), so the promised −8% floor on a `phase1.enabled: false` preset is real. The premise checks out too: `parse-tiers.ts` really does validate only `lock_hw_pct ∈ [0,100]`. Guard is load-bearing — four mutations, all caught. Suite green.

Nits in my comment above (the 5× restatement in decision 7, the 386/386 headroom, the description's number drift) are non-blocking — take or leave them.

The one I'd genuinely like followed up: move the four doesn't-fit checks into `validate_strategy.py` as static predicates. SKILL.md already said "never hand-roll stops" and the incident hand-rolled anyway — the explaining half belongs in prose, the checking half doesn't. Happy to open that PR.

PR #564, opened 2026-09-08
fix(strategy-author): one home for preset values, and a guard that keeps it that way

betashop:
**Stacked on #563** — base is `feat/author-exit-preview`, because it needs that PR's CI wiring for `senpi-strategy-author/tests`. Review after #563; merge order #563 → this.

## Problem

`references/dsl-configuration.md` kept its own copy of all five presets, beside `references/dsl-presets.yaml` in the same directory. Every one had drifted:

| | doc said | actually |
|---|---|---|
| `phase1.enabled` | `true` on all 5 | `false` — trailing was dropped fleet-wide because a floor starting below entry ratchets a *profitable* trade into a loss |
| `max_loss_pct` | 20 / 15 / 15 / 8 / 25 | 8 / 8 / 6 / 5 / 18 |
| banned rung | `lock_hw_pct: 0` on 3 presets | dropped across 129 instances |

It also told the agent *"Or skip presets entirely and hand-author every field"* — the exact opposite of SKILL.md's **"never hand-roll stops"**.

And `creating-a-strategy.md` — the archetype file the build step tells the agent to copy from — carried pre-fix `let_winners_run` verbatim, banned rung included.

## Solution

**Deleted the duplicated half (−155 lines) instead of re-syncing it.** `dsl-presets.yaml` is the only copy of a preset value now. This file keeps what lives nowhere else: the close-reason table, DSL events, and how the exchange stop relates to the DSL floor — none of which are in `runtime-yaml.md`.

Fixed the same rot in the surviving field examples and in `creating-a-strategy.md`. The file was also **orphaned** — nothing linked to it — so `explaining-the-exit.md` now does.

**Guard:** one parametrized test over every `references/*.md` asserting no example teaches `lock_hw_pct: 0` or `phase1.enabled: true`. `dsl-presets.yaml` already stated both were dropped fleet-wide; nothing checked, so three separate docs kept teaching them.

## Verification

```
669 passed, 9 skipped
```
Red-green: reintroduce a breakeven rung into any reference → 1 failed. All intra-file anchors re-verified after the cut (the `#dsl-presets` section it linked to no longer exists).

Doc-only; no shipped behaviour changes, so no version bump beyond #563's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | Skill | Ver | Role |
 |---|---|---|
 | **Analyze** | | |
-| [`senpi-portfolio`](senpi-portfolio/) | 1.16.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
+| [`senpi-portfolio`](senpi-portfolio/) | 1.15.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
 | [`senpi-market-pulse`](senpi-market-pulse/) | 1.2.0 | Daily cross-asset market read (crypto, equities, commodities, macro, funding regime) |
 | [`senpi-smart-money`](senpi-smart-money/) | 1.2.0 | Where the most-profitable wallets are positioned vs. the crowd |
 | [`senpi-trader-research`](senpi-trader-research/) | 1.4.0 | Rank + vet Hyperliquid traders before copying them (mirror-aware: copyability, min-budget, live book) |
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | Skill | Ver | Role |
 |---|---|---|
 | **Analyze** | | |
-| [`senpi-portfolio`](senpi-portfolio/) | 1.15.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
+| [`senpi-portfolio`](senpi-portfolio/) | 1.16.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
 | [`senpi-market-pulse`](senpi-market-pulse/) | 1.2.0 | Daily cross-asset market read (crypto, equities, commodities, macro, funding regime) |
 | [`senpi-smart-money`](senpi-smart-money/) | 1.2.0 | Where the most-profitable wallets are positioned vs. the crowd |
 | [`senpi-trader-research`](senpi-trader-research/) | 1.4.0 | Rank + vet Hyperliquid traders before copying them (mirror-aware: copyability, min-budget, live book) |
--- a/README.md
+++ b/README.md
@@ -68,10 +68,10 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | [`senpi-account-status`](senpi-account-status/) | 1.2.0 | Points, loyalty tier, fees, referrals |
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.19.0 | Conversational picker — rank the catalog against your worldview |
-| [`senpi-strategy-author`](senpi-strategy-author/) | 3.2.0 | Build/edit a DSL-protected strategy package, one decision at a time |
+| [`senpi-strategy-author`](senpi-strategy-author/) | 3.2.1 | Build/edit a DSL-protected strategy package, one decision at a time |
 | [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.7.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.0.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
-| [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.0 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
+| [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.1 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
 | **Move money / positioning** | | |
 | [`senpi-deposit-withdraw-transfer`](senpi-deposit-withdraw-transfer/) | 1.1.0 | The money-movement rails (funds in via embedded wallet or in-app USDC purchase; out via the app) |
 | [`senpi-why`](senpi-why/) | 1.0.3 | "Why Senpi / vs. other tools" — the positioning answer |
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -16,7 +16,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.15.0"
+  version: "1.16.0"
   platform: senpi
   exchange: hyperliquid
   requires:
@@ -153,6 +153,38 @@ capital to redeploy elsewhere, and never "dead money."** That capital is *commit
 it's the dry powder the other half of the design needs to do its job. Only truly-free
 `idle_in_embedded` (and, with care, a *whole* strategy's idle) is redeployable — a flat sleeve of a
 live multi-wallet strategy is not.
+
+### "Why hasn't it traded?" / "why didn't it open that position?" — answer from the outcome codes
+
+A running, healthy strategy that has not traded is the single most common "is it broken?" question,
+and the answer is almost never "it's broken". Every evaluated signal carries a
+`senpi.outcome.reason_code`, so **say which one, how many, and whether the user needs to do
+anything** — never "I'm not sure why."
+
+**Two rules before you answer.** Count the signals rather than describing one, and name the assets
+that were skipped, because "3 eligible entries were skipped" is the part the user actually feels.
+And lead with whether they need to act: for most of these the honest answer is *nothing*, and saying
+so is the useful reply.
+
+| reason code | what to tell the user | do they act? |
+|---|---|---|
+| `withdrawable_unavailable` | "Your available balance was not available during **N** recent scans, resulting in these eligible positions being skipped. If this problem persists, I'd recommend closing the strategy and relaunching it. If that doesn't fix it, contact Senpi Support." | only if it persists |
+| `insufficient_margin` | "Your available balance is fully deployed, resulting in these eligible positions being skipped." It opens on its own as soon as a position closes. | no |
+| `no_margin_configured` | The strategy has no way to size a position, so it can never open one. Redeploy it, or contact Senpi Support if redeploying doesn't fix it. | **yes** |
+| `no_slots` | Every slot is already holding a position — working as designed. | no |
+| `below_min_notional` | The position would be smaller than the $10 exchange minimum. More budget, or higher leverage, would clear it. | their call |
+| `risk_gate_COOLDOWN` | A guard rail is holding it back — the per-asset or global cooldown from its own config. | no |
+| `risk_gate_CLOSED` / `risk_gate_OPEN` | A daily loss limit, drawdown halt or consecutive-loss brake tripped. Name **which**, from the strategy's `risk` block. | no, until it resets |
+| `position_already_exists` | It already holds that asset; it will not double up. | no |
+
+**`no_margin_configured` is the only one that is a defect**, and it is the only one where the user
+should be sent to redeploy. Do not send them to redeploy for the others — a fully deployed balance
+and a missed balance read both resolve on their own, and telling someone to tear down a working
+strategy over either is worse than saying nothing.
+
+**Never say "it isn't trading" without a reason code to hand.** If the engine surfaces no evaluated
+signals at all, that is a different answer — the scanner is not producing candidates, which is the
+"ACTIVE ≠ running" case below, not a sizing one.
 
 ### "ACTIVE" ≠ running — a strategy with no runtime registered is NOT alive, and NOT protected
 
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -16,7 +16,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.16.0"
+  version: "1.15.0"
   platform: senpi
   exchange: hyperliquid
   requires:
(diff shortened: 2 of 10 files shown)


betashop:
Every hole reproduced, every one closed. Your two-regex design is the right one — I had four substring asserts where the shape of the thing being checked is YAML.

**The seven probes, now:**

```
flow-style  { ..., lock_hw_pct: 0 }          -> 1 failed  ✅
block-style   lock_hw_pct: 0 (own line)      -> 1 failed  ✅
              lock_hw_pct: 0.0               -> 1 failed  ✅
phase1/enabled: true at 6-space indent       -> 1 failed  ✅
phase1/enabled: true at 2-space indent       -> 1 failed  ✅
prose "Never set lock_hw_pct: 0 in a ladder" -> passes    ✅  no longer a false positive
breakeven rung into dsl-presets.yaml         -> 2 failed  ✅  the home is guarded now
```

Scoped to fenced yaml, and `.yaml` files with comments stripped — same principle, since the preset file's own header states the ban. A checker that can't tell a rule from a violation makes writing the rule down a CI failure.

**Your survivor was the worst one, and the widened glob caught it on the first run.** `senpi-trading-runtime/references/runtime-yaml.md` shipped both banned patterns in a single copy-paste block — L82 `phase1: { enabled: true, max_loss_pct: 12.0 }` and L86 `{ trigger_pct: 5, lock_hw_pct: 0 }` — in the file an agent copies a whole `runtime.yaml` out of. Fixed, with the reason stated inline once. `senpi-portfolio/SKILL.md:441` is prose explaining what the field means, and correctly passes now.

**The three stale `character` strings are the sharpest catch here** — I deleted a duplicate and routed agents to a home that was itself wrong in the prose:

```
let_winners_run  "No profit lock until +10% ROE"  -> first rung is +20%
balanced         "15% margin floor"               -> max_loss_pct is 8.0
parabolic_runner "Late first lock (+15%...)"      -> first rung is +30%
```

All three named the deleted breakeven rung's trigger. Same 129-instance change, prose left behind — exactly the failure this PR is about, one level up.

**Cuts taken.** Standalone ladder gone (the full example already shows one), ToC gone (19 lines; GitHub renders a heading menu). You were right that both remaining ladders were preset copies that would drift again — I cut them to two rungs and labelled them shape-only, so there is nothing left to drift. `dsl-configuration.md` is 465 → 295.

**Version:** correct, and I'd had it wrong. `senpi-strategy-author` 3.2.0 → 3.2.1, `senpi-trading-runtime` 4.0.0 → 4.0.1 (PATCH per the runtime exception). README synced by the assert from #563.

**693 passed / 9 skipped.**

Two of yours I did not take, and why:
- **`prose "Never set lock_hw_pct: 0"` in `dsl-configuration.md`** — I didn't add the sentence back. The ban is stated once, in `dsl-presets.yaml`'s header, which is now the only home. A second copy is the thing we just deleted 155 lines to stop.
- Nothing else. The rest is in.

0xsarvesh:
**Approve on the approach — one thing in the diff I'd fix before merging.**

Deleting the duplicated half instead of re-syncing it is the right call, and the guard makes it stick. Red-green verified: reintroducing a `lock_hw_pct: 0` rung fails, flipping `phase1.enabled: true` fails. Anchors are clean after the cut — no dangling `#dsl-presets` / `#presets--tuning` anywhere in the repo. Both suites green (I get 693 passed locally, not the 669 in the description — parametrization count, not a problem).

Also confirms the "129 instances" claim from the other side: I grepped all **134** `strategies/*/*/runtime.yaml` — zero banned rungs, zero `phase1.enabled: true`. The rot really did survive only in the docs.

## 1. `creating-a-strategy.md` still ships the failure mode — please fix before merge

`references/creating-a-strategy.md:409-417`. The header comment says *"let_winners_run, with the only thesis-specific tweak: hard_timeout"*, but the ladder underneath is a 2-rung subset — `20/25` and `100/85`, with `30/40`, `50/60`, `75/75` dropped. The inner comment then says *"shape only — copy a real ladder"*, contradicting the header.

Copied verbatim — and this is the archetype file the build step tells the agent to copy from — anywhere between +20% and +100% high-water the active rung stays `20/25`. At a peak of +90% ROE the floor sits at **+22.5%**, where real `let_winners_run` (tier `75/75`) gives **+67.5%**. 45 points of ROE handed back: a trade that peaks in the nineties and round-trips the gain. That's #563's opening paragraph, verbatim.

The new guard can't see it — it bans `lock_hw_pct: 0` and `phase1.enabled: true`, not an inadequate ladder.

Simplest fix is to make the header comment true:

```yaml
      tiers:
        - { trigger_pct: 20,  lock_hw_pct: 25 }
        - { trigger_pct: 30,  lock_hw_pct: 40 }
        - { trigger_pct: 50,  lock_hw_pct: 60 }
        - { trigger_pct: 75,  lock_hw_pct: 75 }
        - { trigger_pct: 100, lock_hw_pct: 85 }
```

## 2. Two shapes slip through the guard

Confirmed by mutation against `test_no_copyable_config_teaches_a_dropped_default`:

- **block style with keys reordered** — `phase1:` / `max_loss_pct: 20.0` / `enabled: true` passes, because `_TRAILING_ON`'s `[^}\n]*\n?` only spans one line. Inline flow style *is* caught, so this is narrower than it looks, but YAML key order is arbitrary.
- **indented (non-fenced) code blocks** — `_YAML_BLOCK` only matches ` ```yaml ` fences, so a 4-space-indented example with `lock_hw_pct: 0` passes.

Both are a couple of characters of regex.

## 3. Consider extending `_COPYABLE` to the packages

The glob covers `references/` dirs plus `dsl-presets.yaml`. The fleet is clean today, but the guard protects the docs — and the docs weren't where the money was. `strategies/*/*/runtime.yaml` is the surface where a breakeven rung actually costs someone.

Nice catch on `let_winners_run`'s `character:` string, by the way — "+10%" vs. a first tier at 20 was its own small lie.

0xsarvesh:
#563 is approved. Here is exactly what I need on this one to approve — **two blocking items, both inside this diff, both small.** Everything else in my earlier comment is take-or-leave.

---

### ✅ Blocking 1 — restore the three dropped rungs in `creating-a-strategy.md`

`references/creating-a-strategy.md:409-417`.

To be fair about the baseline: main's version of this block was worse overall — it carried `phase1.enabled: true`, `max_loss_pct: 20.0`, **and** the banned `10/0` breakeven rung. This PR kills all three, which is the right direction.

But it also dropped `30/40`, `50/60` and `75/75` alongside the bad rung, and nothing about removing a breakeven rung required removing three good ones. On the +20%→+100% range main is now the better-protected version: at a peak of +90% ROE the surviving `20/25` rung floors at **+22.5%**, where `75/75` floors at **+67.5%**.

That is a 45-point giveback sitting in the file the build step tells the agent to copy from, under a header comment that claims the block *is* `let_winners_run`. Make the comment true:

```yaml
      tiers:
        - { trigger_pct: 20,  lock_hw_pct: 25 }
        - { trigger_pct: 30,  lock_hw_pct: 40 }
        - { trigger_pct: 50,  lock_hw_pct: 60 }
        - { trigger_pct: 75,  lock_hw_pct: 75 }
        - { trigger_pct: 100, lock_hw_pct: 85 }
```

…and drop the now-contradictory `# shape only — copy a real ladder` comment.

### ✅ Blocking 2 — close the `_TRAILING_ON` hole

The PR's headline is "a guard that keeps it that way", and this shape walks straight through it (I confirmed by mutation — it passes today):

```yaml
phase1:
  max_loss_pct: 20.0
  enabled: true          # ← not caught; `[^}\n]*\n?` only spans one line
```

YAML key order is arbitrary, so this isn't a contrived case. Inline flow style *is* caught, so the hole is block-style-only.

Tested replacement — 9/9 on my cases (enabled-first, reordered, inline both orders, trailing comment after `phase1:`, nested under `dsl_preset:`), and no false positives against the real `dsl-presets.yaml` or `dsl-configuration.md`:

```python
# Indent-anchored: capture only the lines nested UNDER this phase1, so a sibling
# `phase2: {enabled: true}` at the same indent isn't swallowed into the match.
_PHASE1 = re.compile(
    r"^([ \t]*)phase1:[ \t]*(?:#[^\n]*)?(\{[^}]*\}|(?:\n\1[ \t]+\S[^\n]*)*)", re.M)
_ENABLED_TRUE = re.compile(r"\benabled:\s*true\b", re.I)
```

```python
assert not any(_ENABLED_TRUE.search(block) for _, block in _PHASE1.findall(text)), \
    f"{doc.name} has `phase1.enabled: true` — trailing is off fleet-wide (ratchets into a loss)"
```

Worth noting the `\1` backreference is doing real work: without it the naive version matches sibling `phase2: enabled: true` in the real presets file and fails CI on a clean tree. I hit that on the first attempt.

---

### Non-blocking, fix whenever

- **Indented (non-fenced) code blocks skip the guard** — `_YAML_BLOCK` is fenced-only, so a 4-space-indented example with `lock_hw_pct: 0` passes. Lower risk than the above since the repo convention is fenced, and widening it risks prose false positives.
- **Extend `_COPYABLE` to `strategies/*/*/runtime.yaml`.** The fleet is clean today — I grepped all 134, zero violations — but the guard currently protects the docs, and the docs weren't where the money was.

---

Rest of the PR is good and I don't want to hold it long: deleting the duplicate instead of re-syncing is the right instinct, anchors are clean after the cut, red-green checks out, and the `let_winners_run` `character:` fix ("+10%" against a first tier at 20) was its own small lie worth catching.

Push those two and I'll approve straight away — or if you'd rather keep the stack moving, say so and I'll push them to the branch myself.

betashop:
Both blocking items pushed, plus your third.

**Blocking 1 — the rungs.** Your arithmetic is right and the cut was mine. I removed three good rungs alongside the banned one, under a header comment still claiming the block *is* `let_winners_run`. +22.5% against +67.5% at a +90% peak, in the archetype file the build step points the agent at — that's #563's opening paragraph, reintroduced by the PR cleaning it up. Full ladder restored, comment made true, contradictory "shape only" line gone.

The mistake underneath it: I over-applied "don't duplicate preset values". That's right for a schema reference; this file's example is meant to *be* a working recipe.

**Blocking 2 — the hole.** Used your indent-anchored version as written. Seven shapes now caught — enabled-first, reordered, both inline orders, trailing comment after `phase1:`, nested at 6-space indent, `True` casing — with no false positive on `phase1: false` beside a live `phase2` or `hard_timeout`. You were right that the `\1` is load-bearing; I checked the naive form against the real presets file and it fires on the sibling exactly as you said.

**Non-blocking 3 — taken.** `_COPYABLE` now covers `strategies/*/*/runtime.yaml`. "The guard protects the docs, and the docs weren't where the money was" is the correct read. **29 → 163 files guarded.** Fleet clean, matching your grep.

Two things worth flagging from wiring that up, both my errors:

- The first version used `p.parts[0] == "strategies"` against an **absolute** path, so it matched nothing. The suite stayed at 45 tests and passed — a guard that looked extended and checked nothing. `relative_to(_ROOT)` now; 45 → **179** tests.
- 134 packages ship a `runtime.yaml`, most under `main/`, so a failure read `main/runtime.yaml81`. Ids are the full package path now, so CI names the strategy.

Red-green on a shipped package: breakeven rung into stingray → `1 failed [strategies/stingray/main/runtime.yaml]`. **827 passed / 9 skipped.**

Leaving the indented-code-block case alone, as you suggested — repo convention is fenced, and widening it reintroduces the prose false positive we just removed.

#563 is merged, so this is now based on it.

betashop:
Heads-up on the diff, unprompted: a `git add -A` while applying your fixes swept an in-progress `senpi-portfolio` section into this branch — not something you reviewed, and not related to the DSL work. Reverted; net diff for that file is zero and the PR is back to 9 files. It ships on its own in #565.

Nothing else changed since your review: rungs restored, indent-anchored regex in as written, `_COPYABLE` extended to the 134 packages.

0xsarvesh:
Re-reviewed at `a0c895dc`. Blocker 1 is done. Blocker 2 is done *as I specified it* — **and my spec was wrong.** Details below; the fix is one line and I've verified it.

### ✅ Blocker 1 — resolved

`creating-a-strategy.md` now carries all five rungs and the comment says what the block actually is. Verified byte-for-byte against `dsl-presets.yaml`.

### ❌ Blocker 2 — my regex is blind on 129 of 134 packages

This one's on me. The regex I handed you was tested against synthetic cases and the two reference docs — **not** against a real `runtime.yaml`. Shipped packages put explanatory comments between `phase1:` and `enabled:`:

```yaml
    phase1:
      # Trailing DISABLED: the phase-1 floor trails HIGH-WATER, so after a favourable move it
      # could sit BELOW entry and close a winner at a loss. ...
      enabled: false
```

`_config_text` strips comments to **whitespace-only lines**, and my `(?:\n\1[ \t]+\S[^\n]*)*` requires `\S` on every line — so the capture terminates at the first stripped comment and `phase1` comes back as `''`.

Measured across the tree:

```
shipped packages: 134
  phase1 capture EMPTY with current regex : 129   <- guard blind
  phase1 capture EMPTY with fixed   regex : 0
mutating albatross phase1.enabled -> true:
  caught by current regex: False
  caught by fixed   regex: True
```

So extending `_COPYABLE` to `strategies/*/*/runtime.yaml` reads as fleet coverage but the phase1 half doesn't fire on 129 of them. (The `lock_hw_pct: 0` half is fine — that's a plain search over the whole text, not the capture. And 129/134 is a coincidence with the "129 instances" line, but a pointed one.)

**Fix — allow whitespace-only lines inside the block:**

```python
_PHASE1 = re.compile(
    r"^([ \t]*)phase1:[ \t]*(?:#[^\n]*)?"
    r"(\{[^}]*\}|(?:\n(?:\1[ \t]+\S[^\n]*|[ \t]*))*)", re.M)
```

The `\1[ \t]+\S` alternative still requires strictly deeper indent than `phase1:`, so a sibling `phase2: {enabled: true}` at equal indent is still not swallowed — I re-checked that specifically, since it's the failure the backreference exists to prevent.

Verified in situ with the patch applied:

```
full suite                                        827 passed, 9 skipped
mutate phase1.enabled -> true in albatross        1 failed  ✅ (was passing)
mutate phase1 block-style reordered in a doc      1 failed  ✅
clean tree                                        169 passed, 0 false positives
```

### Unrelated: `senpi-portfolio` is in this diff

`senpi-portfolio/SKILL.md` +34 (the `senpi.outcome.reason_code` table) and a 1.15.0 → 1.16.0 bump are in here, not on main. It's good content — `no_margin_configured` being the only redeploy-worthy code is a genuinely useful distinction — but it's a user-facing behaviour change to a different skill riding in a doc-only "one home for preset values" PR, with no test and nothing in the description about it. Easier to review, and to revert, on its own. Your call whether to split it or just call it out in the description.

---

Push the regex line and I'll approve. Everything else here is good — the guard sweeping all 134 packages is the right instinct and worth getting actually working.

0xsarvesh:
Pushed the fix myself so this isn't blocked on another round trip: 9ae3d7be

`packages: 134 · phase1 capture EMPTY: 129 → 0`

Verified before pushing — full suite 827 passed/9 skipped; `phase1.enabled: true` in albatross now fails (was passing), block-style reordered phase1 in a doc fails, `lock_hw_pct: 0` in albatross fails, clean tree 169 passed with no false positives. Only the test file changed.

Also confirmed the `\1` backreference still does its job — a sibling `phase2: {enabled: true}` at equal indent is not swallowed, which was the reason it was there.

That clears both blockers. **Approving now.** Thanks for the quick turnarounds, and for pulling the portfolio work out into its own PR.

One thing worth carrying forward rather than re-litigating here: this guard has now had three regex defects — the original one-line span, my sibling-swallow, and this whitespace blindness. Each looked right and each failed silently green. A regex over YAML can't see nesting, so every new file shape is a new hole. I'll fold a parsed-YAML replacement into the `validate_strategy.py` PR, where the same predicates (`lock_hw_pct: 0`, first rung > ~40%, non-monotonic locks, `phase1.enabled`) are needed anyway — one module, used by both the author-time gate and CI, and the regex goes away.

0xsarvesh (approved):
Both blockers cleared — the five rungs are restored verbatim from `dsl-presets.yaml`, and the phase1 guard now actually sees all 134 packages (9ae3d7be). Portfolio scope split out. Green at 827 passed, red-green verified on doc and package mutations in both directions.

Deleting the duplicate instead of re-syncing it was the right instinct, and sweeping the shipped packages was a better idea than my original ask.

PR #565, opened 2026-09-08
feat(portfolio): answer "why hasn't it traded?" from the outcome codes

betashop:
**Stacked on #564.** Review after that one; merge order #564 → this.

## Problem

A running, healthy strategy that hasn't opened a position is the most common "is it broken?" question — and `senpi-portfolio` has no answer for it today. The agent guesses or says nothing, and the user can't see the logs that hold the reason.

## Solution

Every evaluated signal already carries `senpi.outcome.reason_code`. This maps **all 11** codes `outcome()` emits to **what to tell the user** and **whether they need to act**:

| code | act? |
|---|---|
| `withdrawable_unavailable` | only if it persists — it usually clears within a few scans, then support |
| `insufficient_margin` | no — balance fully deployed, opens when a position closes |
| `no_margin_configured` · `invalid_direction` · `strategy_backend_paused` | **yes** |
| `position_open_failed` | depends on the exchange's own message, which rides the outcome |
| `no_slots` · `risk_gate_COOLDOWN` · `risk_gate_CLOSED` · `position_already_exists` | no |
| `below_min_notional` | their call — more budget or leverage |

For **most of them the honest answer is "nothing"**, and saying that plainly is the useful reply.

**Two cases with no reason code at all**, because neither reaches the per-signal path:

- **A scanner that has gone blind** — `runs > 0`, `signals = 0`, and `health=healthy` with `errors=0`, which is why users find it by noticing the quiet rather than from anything we show them. Distinguished from a merely selective strategy by whether it *used to* produce signals.
- **`clearinghouse_unknown`** — the runtime refuses to open against an exchange view it could not read and skips the batch wholesale, never calling `outcome()`. Scanner fine, exchange read not.

So "no codes" splits three ways: no codes **and** no signals → the scanner · no codes **with** signals → look for `clearinghouse_unknown` · no runtime at all → the "ACTIVE ≠ running" case.

## Depends on Senpi-ai/senpi-trading-runtime#387

That PR splits `withdrawable_unavailable` out of `no_margin_configured`, and makes a `$0` balance report as `insufficient_margin`. Until it ships, both states arrive as `no_margin_configured` and the agent will wrongly tell the user to redeploy — which is exactly the bug #387 fixes. The section is still a net improvement before then (six codes are unaffected), but the two margin rows only become correct with it.

## Verification

All 11 codes checked against the runtime rather than recalled. **`risk_gate_OPEN` is not reachable** — `open-position.action.ts` emits `risk_gate_${gate.gate}` only when the gate is *not* `OPEN` — so it's deliberately absent. I had it in a first draft; the check caught it.

**Review round 1 (@IgnasPeciura):** the table covered 8 of 11 — added `strategy_backend_paused`, `position_open_failed`, `invalid_direction`. The "no codes" fallback sent `clearinghouse_unknown` to "the scanner isn't producing candidates", when the scanner is fine and the exchange read is not; rewritten. And the `withdrawable_unavailable` remedy said "close and relaunch", which contradicted #387's own log and could not work — relaunching rebuilds the same percent-sized recipe against the same read. Now: wait, then support, with a line saying why relaunch is *not* the fix.

`827 passed / 9 skipped`. senpi-portfolio 1.15.0 → 1.16.0, README synced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | Skill | Ver | Role |
 |---|---|---|
 | **Analyze** | | |
-| [`senpi-portfolio`](senpi-portfolio/) | 1.15.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
+| [`senpi-portfolio`](senpi-portfolio/) | 1.16.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
 | [`senpi-market-pulse`](senpi-market-pulse/) | 1.2.0 | Daily cross-asset market read (crypto, equities, commodities, macro, funding regime) |
 | [`senpi-smart-money`](senpi-smart-money/) | 1.2.0 | Where the most-profitable wallets are positioned vs. the crowd |
 | [`senpi-trader-research`](senpi-trader-research/) | 1.4.0 | Rank + vet Hyperliquid traders before copying them (mirror-aware: copyability, min-budget, live book) |
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -168,7 +168,7 @@ so is the useful reply.
 
 | reason code | what to tell the user | do they act? |
 |---|---|---|
-| `withdrawable_unavailable` | "Your available balance was not available during **N** recent scans, resulting in these eligible positions being skipped. If this problem persists, I'd recommend closing the strategy and relaunching it. If that doesn't fix it, contact Senpi Support." | only if it persists |
+| `withdrawable_unavailable` | "Your available balance couldn't be read during **N** recent scans, so these eligible positions were skipped. Your funds aren't affected and it usually clears on its own within a few scans. If it's still happening after that, contact Senpi Support." | only if it persists |
 | `insufficient_margin` | "Your available balance is fully deployed, resulting in these eligible positions being skipped." It opens on its own as soon as a position closes. | no |
 | `no_margin_configured` | The strategy has no way to size a position, so it can never open one. Redeploy it, or contact Senpi Support if redeploying doesn't fix it. | **yes** |
 | `no_slots` | Every slot is already holding a position — working as designed. | no |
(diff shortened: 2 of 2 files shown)


betashop:
The three missing codes are in, and you were right that the fallback was pointing at the wrong thing.

**Added:** `strategy_backend_paused`, `position_open_failed`, `invalid_direction`. The table covered 8 of the 11 `outcome()` emits — `strategy_backend_paused` in particular is actionable and had no answer at all.

**The fallback.** `clearinghouse_unknown` never calls `outcome()`, so it carries no `reason_code`, and I was routing it to "the scanner isn't producing candidates" — when the scanner is fine and the exchange view is unreadable. That's the MCP-wedge class, i.e. exactly the case a user is most likely to be asking about. "No codes" now splits three ways: **no codes and no signals** → the scanner · **no codes with signals present** → look for `clearinghouse_unknown` · **no runtime** → the "ACTIVE ≠ running" case.

**The cross-PR mismatch is the one I'd have shipped.** The row told users to close and relaunch while #387's log points at adding a withdrawable-independent floor — and relaunching rebuilds the same percent-sized recipe against the same read, so it could not have worked. Now: it usually clears within a few scans, then support. I've left an explicit line saying *why* relaunch isn't the answer, so it doesn't get helpfully added back later.

That was copy Jason wrote, so I've flagged the change to him rather than quietly swapping it.

Both your cuts taken. All 11 codes re-verified against the runtime. **827 passed / 9 skipped.**

On merge order — agreed, 563 → 564 → 565, and 565 wants #387 shipped first. 563 is merged; 564 is waiting on Sarvesh's approve after his two blockers.

betashop:
All three verified before applying, and all three were right.

**The wrong command is the worst of them** — an agent following my text runs `senpi runtime list`, sees `running`, and is stuck exactly where the user already is. It's `senpi scanner` (`senpi-commands.ts:1985`, and its own description is literally "barren detection"). Both are named now, including that `runtime list` carries the sibling `running — NO ENTRY SCANNERS`.

On the abridged sample line: rather than reprint a line that drifts, the section no longer quotes it. Which folds neatly into your cut.

**`timeout` earns its row on the code's own words** — `open-position.action.ts:765`: *"A timeout is not a rejection: the server may have filled it."* The user can be holding a position the runtime doesn't know about, so the row says check live positions before answering and never suggest a retry first. That's a materially more dangerous answer than `position_open_failed`'s clean rejection, and I'd have shipped without it. `exception` added as the 13th. **13 of 13**, each re-verified.

**Cut taken.** You were right that the mechanism half contradicted the `diagnose.py` handoff this skill already makes — and both factual errors lived in that half, which is its own argument. Kept the interpretation: nothing-qualified vs blind, and never trust `health=healthy` on this failure. Net, the section is *shorter* than before this round despite gaining two rows.

**One I didn't take: the `BODY_BUDGET` entry for portfolio.** You're right that nothing pushes back on its size. But setting the ratchet at a number I just grew is the wrong precedent — portfolio is 795 lines against ops' 300 and author's 386, and what the right bar should be deserves a decision of its own rather than being set as a side effect of this PR. Worth doing; worth doing deliberately.

**827 passed / 9 skipped.** And thank you for testing the divergence on #387 rather than taking my word — that's the second time this stack a probe has been the thing that settled it.

PR #569, opened 2026-09-10
fix(close): the documented `closed` poll signal was unreachable on the package path

### LINKED ISSUE #568: close.py: the documented `closed` poll signal is unreachable on the package path, so a close/redeploy loop mints a wallet per pass
## The contract that isn't kept

`close.py`'s module docstring defines the polling contract:

> → reports `closing`. strategy_close is async, so POLL by re-running `close.py <id>`: it's idempotent (runtime already gone → skip; status already closing/closed → skip re-submit) and **reports `closed`** once the strategy leaves the active set.

And `references/editing-a-live-strategy.md` step 4 depends on it:

> **Then `close.py <id>`**, wait for `closed`, and `create` the edited package…

**On the package path, `closed` is unreachable.** The poll's only terminal output is the generic `"<pkg>: no OPEN strategies to close."`

## Mechanism

1. The `--all` / `<package>` path reads strategies filtered **server-side** by `LIVE_STATUSES` (`_cli.py:673`) = `ACTIVE, PAUSED, CREATE_WALLET, FUND_WALLET, INITIALIZE_POSITIONS, SUBSCRIBE_TRADER, CLOSING_POSITIONS`.
2. Once `strategy_close` settles, the status becomes `CLOSED` — which is in `DEAD_STATUSES`, not `LIVE_STATUSES`.
3. The next poll therefore returns **zero rows** → `targets` is empty → `close.py:250-252` prints `no OPEN strategies to close.` and `sys.exit(0)`.

So a successful close and "there was never anything there" produce **byte-identical output and the same exit code**. There is no state in which the package path can print `closed`.

## This exact hazard is already fixed one path over

The file's own comments call it out twice, and the `--strategy-id` / `--address` path was deliberately built to avoid it (`close.py:204-211`):

> Read UNFILTERED by status, unlike `--all`/`<package>` above: a status-filtered read cannot tell "no such id" from "already closed" (both return zero rows), and either would otherwise fall through to the "no OPEN strategies to close" + exit 0 below — a false all-clear on a target named explicitly.

and `_read_or_refuse` (`close.py:49-53`):

> teardown then finds no targets, prints "no OPEN strategies to close" and exits **0**. That is a positive all-clear the surface never earned.

The reasoning is correct and already written down. It just wasn't applied to the path the edit loop actually uses.

## Measured cost

With the poll signal ambiguous, an agent reads the all-clear as "close finished" and runs `create`,
which mints a fresh wallet rather than adopting. Observed on one account over four days:

- **21 strategy wallets**, all now CLOSED, lifetimes of **5, 8, 11 and 14 minutes**
- Funding bled across the cycle as each attempt stranded a little more
- **71 separate agent conversations in 3.5 days**, which exhausted the plan's AI credits
- The box went `suspended` / `out_of_credits` and stayed dark, while the subscription kept billing

The budget decrements are the tell — the agent lowering its own ask as each attempt strands funds:

```
22:31:29  deploy.py create <pkg> --budget 131
22:31:51  deploy.py create <pkg> --budget 130      # 22 seconds later
...
22:36:12  deploy.py create <pkg> --budget 130
22:36:38  deploy.py create <pkg> --budget 129
```

And the poll that triggered each one:

```
22:22:51  close.py <pkg>  → "Command still running"          (async trigger)
22:23:28  close.py <pkg>  → "<pkg> v1.0.0: no OPEN strategies to close."
22:23:34  deploy.py create <pkg> --budget 131                 → new wallet
```

Nothing was stranded on-chain in the end (every wallet verified $0 across perp + spot + `dex:xyz`,
no open positions, no resting orders), so this cost money and credits rather than a lost position.
On a package whose exit engine was configured not to fire, a mid-close position would have been
worse.

**Note on the trigger path.** This was originally written when every tune iteration was
`close → validate → create` by design. Since #546/#573, a live edit applies in place via
`deploy.py update`, so the edit loop is no longer the usual way in. The defect stands regardless:
the changes `update` refuses — a different wallet, a renamed/moved external scanner, a changed
`action_type` — still close and redeploy, and every plain `close.py <id>` still polls this path.

## Suggested fix

Make the package path able to report the terminal state, the same way the direct path already does:

1. On the `--all` / `<package>` path, when the `LIVE_STATUSES` read returns zero rows, do a second **unfiltered** read for that package before concluding anything.
2. Distinguish three outcomes that are one outcome today:
   - rows exist and all are in `DEAD_STATUSES` → print **`closed`** (the documented terminal signal), exit 0
   - no rows at all for the package → `no strategies found for <pkg>` (never was there)
   - rows still in `LIVE_STATUSES` → `closing`, keep polling
3. Consider a distinct exit code for `closed` vs `nothing-to-close`, so the caller can branch without string-matching.

That makes `editing-a-live-strategy.md` step 4 ("wait for `closed`") actually satisfiable, and it removes the ambiguity that turns one edit into one wallet.

Found while auditing this week's new paid subscribers against telemetry (2026-09-10). Read-only investigation; nothing was mutated on the user's box.

shnoodles:
Fixes #568.

## The contract that couldn't be kept

`close.py`'s module docstring defines the polling contract:

> → reports `closing`. strategy_close is async, so POLL by re-running `close.py <id>`: it's idempotent … and **reports `closed`** once the strategy leaves the active set.

`references/editing-a-live-strategy.md` step 4 depends on it ("wait for `closed`"), and the poll hint the command prints says the same. **It could never print it.**

The `<package>` read is filtered server-side by `_cli.LIVE_STATUSES`, which does not include `CLOSED`. The moment `strategy_close` settles, the row leaves the read → zero rows → `main` falls through to `no OPEN strategies to close.` + exit 0. Byte-identical output *and* exit code to "this package was never deployed here."

A caller polling for `closed` cannot tell a completed close from an absence.

## Why that costs money

With the poll ambiguous, an agent reads the all-clear as completion and runs `create`, which mints a fresh wallet rather than adopting — one per pass of any loop that closes and redeploys.

Observed on one account over four days:

- **21 strategy wallets**, lifetimes of 5, 8, 11 and 14 minutes
- funding bled a little further with each stranded attempt
- **71 agent conversations in 3.5 days**, exhausting the plan's AI credits
- the box went `suspended` / `out_of_credits` and stayed dark, while the subscription kept billing

**On the trigger path:** this was found when applying a live edit still meant `close → validate → create`. Since #546/#573 that is no longer so — `deploy.py update` applies in place. The defect stands regardless: the changes `update` refuses (a different wallet, a renamed/moved external scanner, a changed `action_type`) still close and redeploy, and every plain `close.py <id>` still polls this path.

The transcript tell is the agent decrementing its own budget as each attempt stranded funds:

```
22:22:51  close.py <pkg>  → "Command still running"                       (async trigger)
22:23:28  close.py <pkg>  → "<pkg> v1.0.0: no OPEN strategies to close."  ← read as "done"
22:23:34  deploy.py create <pkg> --budget 131                             → new wallet
...
22:31:29  deploy.py create <pkg> --budget 131
22:31:51  deploy.py create <pkg> --budget 130    # 22 seconds later
22:36:12  deploy.py create <pkg> --budget 130
22:36:38  deploy.py create <pkg> --budget 129
```

## The fix is the one this file already applies one path over

`--strategy-id` / `--address` reads **unfiltered** by status, and says why:

> a status-filtered read cannot tell "no such id" from "already closed" (both return zero rows), and either would otherwise fall through to the "no OPEN strategies to close" + exit 0 below — a false all-clear on a target named explicitly.

The reasoning was already written down. It just wasn't applied to the path the edit loop uses. So on the package path, when the LIVE-filtered read is empty, re-read unfiltered and split one outcome into two:

| state | before | after |
|---|---|---|
| rows exist, none LIVE | `no OPEN strategies to close.` | **`closed.`** |
| no rows at all | `no OPEN strategies to close.` | `no strategies for this package on this account — nothing to close.` |

**Scope.** Guarded by `pkg is not None`. `--all` answers a set question where zero rows is a legitimate "nothing open", and the direct path is already correct — both keep their exact existing output. `pkg` is initialized to `None` at line 189, so neither path can reach the new branch.

**Money-path safety.** The second read goes through `_read_or_refuse`, so an unreadable list still refuses rather than rendering as a finished close — the same rule guarding the first read. `test_an_unreadable_second_read_refuses_rather_than_claiming_closed` covers exactly that.

## Tests

Six new tests in `MainPackagePollReportsClosed`. The one that states the defect plainly just asserts the two outcomes differ:

```
AssertionError: 'omega v1.0.0: no OPEN strategies to close.'
             == 'omega v1.0.0: no OPEN strategies to close.'
```

The test double **honours the `statuses=` argument** the way the real server-side filter does. The pre-existing fake in `MainDirectAddressing` ignores it — which is why no existing test could catch this.

Full CI suite: **841 passed, 8 skipped** (`test_close.py` alone: 30, up from 24). `test_min_budget_vendor_parity.py` green.

Found while auditing this week's new paid subscribers against telemetry. The investigation itself was read-only — nothing was mutated on the user's box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.20.0 | Conversational picker — rank the catalog against your worldview |
 | [`senpi-strategy-author`](senpi-strategy-author/) | 3.4.0 | Build/edit a DSL-protected strategy package, one decision at a time |
-| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.9.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
+| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.10.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.2.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
 | [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.3 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
 | **Move money / positioning** | | |
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -24,7 +24,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.9.0"
+  version: "3.10.0"
   platform: senpi
   exchange: hyperliquid
   requires:
--- a/senpi-strategy-ops/scripts/close.py
+++ b/senpi-strategy-ops/scripts/close.py
@@ -248,6 +248,32 @@ def main(argv):
         strategy_label = pkg.id
 
     if not targets and not a.dry_run:
+        # A LIVE-filtered read that came back empty has TWO causes and they are not the same answer:
+        # the close we are being polled for has completed, or this package was never deployed here.
+        # `close.py`'s contract is that this poll "reports `closed` once the strategy leaves the
+        # active set" (module docstring), and the poll hint printed above tells the agent to re-run
+        # for exactly that. But `LIVE_STATUSES` excludes `CLOSED`, so a settled close reads as zero
+        # rows and used to print the same all-clear as "nothing here". An agent that reads absence
+        # as completion goes on to `deploy.py create`, which mints a fresh wallet — one per pass of
+        # any loop that closes and redeploys.
+        # Editing a live strategy no longer takes that path (`deploy.py update` applies in place),
+        # but the changes `update` refuses — a different wallet, a renamed/moved external scanner,
+        # a changed `action_type` — still close and redeploy, and every plain `close.py <id>` still
+        # polls right here.
+        # `--strategy-id`/`--address` already re-reads UNFILTERED for this reason; so does this now.
+        if pkg is not None:
+            ever = _read_or_refuse(
+                _cli.strategies_for_or_none(mcp, skill_name=pkg.id, why=why), why, pkg.id)
+            if ever:
+                # Rows exist but none is LIVE, so every one is terminal. Name the statuses rather
+                # than printing a bare `closed.`: this branch also fires for a package whose only
+                # records are old FAILED/TERMINATED strategies, where no close ever ran.
+                seen = sorted({str(_cli.strategy_status(r) or "?").upper() for r in ever})
+                noun = "strategy" if len(ever) == 1 else "strategies"
+                print(f"{hdr}: closed — {len(ever)} {noun}, all {'/'.join(seen)}.")
+            else:
+                print(f"{hdr}: no strategies for this package on this account — nothing to close.")
+            sys.exit(0)
         print(f"{hdr}: no OPEN strategies to close.")
         sys.exit(0)
 
--- a/senpi-strategy-ops/tests/test_close.py
+++ b/senpi-strategy-ops/tests/test_close.py
@@ -305,5 +305,143 @@ class ReadOrRefuse(unittest.TestCase):
         self.assertNotEqual(ctx.exception.code, 0)
 
 
+class MainPackagePollReportsClosed(unittest.TestCase):
+    """`close.py <pkg>` is the POLL the module docstring documents: "reports `closed` once the strategy
+    leaves the active set".
+
+    It could never print it. The package read is filtered server-side by `LIVE_STATUSES`, so the moment
+    `strategy_close` settles the row becomes `CLOSED` (a `DEAD_STATUSES` member), the read returns zero
+    rows, and `main` falls through to the generic `"no OPEN strategies to close."` + exit 0 — byte
+    identical to "this package was never deployed here". An agent polling for `closed` reads that
+    all-clear as completion and runs `deploy.py create`, which mints a fresh wallet. A close/redeploy
+    loop therefore leaks one wallet per pass, each carrying whatever funding it was given.
+
+    `--strategy-id`/`--address` was already fixed for exactly this (it reads UNFILTERED by status,
+    because "a status-filtered read cannot tell 'no such id' from 'already closed'"). This is that same
+    fix on the package path."""
+
+    def setUp(self):
+        self._orig = (close.MCPClient, _cli.strategies_for_or_none, _cli.list_runtimes,
+                      _cli.list_runtimes_or_none, _cli.run_cli, close._pkg.load)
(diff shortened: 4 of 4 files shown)


betashop (commented):
**Review: the fix is right and minimal — one hygiene change before merge, two wording notes.**

Ran it on the branch (rebased on today's main, 0 behind): `test_close.py` 30 passed, full CI suite 887 passed / 9 skipped. README row and frontmatter agree.

### Verified
- **The edge I checked first: an in-flight close cannot print `closed` early.** `CLOSING_POSITIONS` is in `LIVE_STATUSES` and `strategy_open()` is "not in `DEAD_STATUSES`", so a closing row stays in `targets`, `close_one` skips the re-submit and reports `closing`, and the new branch is only reached once the LIVE read has zero open rows. Correct.
- The second read goes through `_read_or_refuse`, so an unreadable list still refuses instead of rendering as a finished close. `--all` and the direct path are untouched (`pkg` stays `None` there).
- The test double honouring `statuses=` is the real improvement in the test file; the pre-existing fake could never have caught this.

### Change before merge — this repo is public
The user id and the incident figures (wallet count, funding numbers, budget-decrement transcript) are in the `close.py` comment, the test-class docstring, the PR body and issue #568. House rule for senpi-skills: code and tests carry the **rule**, identifiers and incidents stay in the internal repos. Please scrub the id and the dollar/wallet figures from the comment and the docstring (the mechanism explanation is what a future reader needs and it stands on its own), and edit the PR body / #568 to drop the id — link the internal write-up instead if you want the evidence preserved.

### Wording
1. **The premise moved this morning.** The comment says the edit loop is "close → validate → create by design". Since #546/#573 landed today, a live edit is `deploy.py update` in place; close-and-redeploy is only for the three refused changes (wallet, renamed/moved external scanner, `action_type`). The fix is still needed — that procedure and every plain "close X" still poll for `closed` — but say that, so nobody reads this file as endorsing close-and-redeploy as the edit path.
2. **"rows exist, none LIVE → `closed.`"** also fires for a package whose only records are old `FAILED`/`TERMINATED` strategies, where no close ever ran. Worth stating the terminal statuses in the line (`closed — 2 strategies, all CLOSED`) so a first-time `close.py <id>` on a dead package does not read as a close that just completed. Optional; the important property (distinguishable from absence) holds either way.

### Minor
- The unfiltered re-read pulls the package's whole history, but only on the empty-LIVE branch — fine.
- Version: convention here has been a MINOR bump for a behaviour change in a `senpi-*` skill (PATCH is the exception for senpi-trading-runtime). 3.9.1 auto-updates just the same if the upgrader keys on any change; 3.10.0 would match the convention. Not blocking.

betashop (approved):
Approved. All four points addressed on ec3d4f4: identifier and incident figures out of the code comment, test docstring, PR body and #568; the comment now says live edits go through `deploy.py update` and names the two paths that still poll here (the refused changes, and every plain close); the `closed` line names the terminal statuses so a dead package cannot read as a close that just completed; 3.10.0 with the README row synced. Re-ran on this head: `test_close.py` 30 passed, full suite 887 passed. CI green.

PR #571, opened 2026-09-10
trade: an empty trade-tool result is a denied approval, not a placed trade (1.1.0)

betashop:
## Why

`senpi-trade`'s guardrail table still says an "approval / approve again" step **does not exist**. It does now: trade tools run behind the agent's trade-approval gate, and a call the user does not approve in time is **denied with an empty result** — nothing reached the venue. Read with the old row, an agent either keeps polling `strategy_list` for a trade that never happened or reports the empty result as a placed trade.

## What changed

- The guardrail row now says what an empty result is (a denied approval → not executed) and keeps the original discipline for errored calls (surface the verbatim text, read `strategy_list` only to confirm what exists).
- The matching red flag says the same in one line.
- `senpi-trade` 1.0.0 → 1.1.0 (frontmatter + README).

Skill text only, rule only — no scripts touched.

## Runtime side

Prompt text is the weakest control for this — at the call site an empty result is indistinguishable from a successful no-op. The durable fix is the runtime returning an explicit denial on timeout: Senpi-ai/senpi-trading-runtime#402 (fix-list item 4). This PR is the skill half; that issue is the runtime half.


shnoodles:
**Review: yes — four lines, and they are the right four.**

"An empty tool result is a denied approval, not a placed trade" is the precise inversion of what happened in the incident: four ratchet stops waited 584 s on the gate, were denied, and two sessions later were reported as active. The new table row makes the distinction the old text missed — an **errored** call carries its own message, an **unapproved** one comes back empty — and the remedy ("say the approval timed out and nothing was placed; offer to re-run") is actionable rather than a warning.

Keeping `strategy_list` in the remedy but demoting it to "read only to confirm what actually exists" is the right call; the old row's "poll `strategy_list` for the real status" is what turned a denial into a polling loop.

**One thing to change, and it is about scope rather than content:** this is prompt text, and the failure it describes is a model that confidently reported a trade it had not placed. Prompt text is the weakest available control for that. The Slack list marks item 4 "Skill side done · runtime side not started", which is the right reading — but worth saying here so nobody closes the incident item off this PR: an empty result is *structurally* indistinguishable from a successful no-op at the call site, so the durable fix is the runtime returning an explicit denial on timeout rather than an empty result. Please link this PR to that runtime issue so the two do not drift apart.

Minor: the new table row is long enough to wrap awkwardly in a rendered table. Consider moving the parenthetical ("An errored call is different: it carries its own message") into the remedy column, which already makes the same split.

Ponytail pass: `+4/-4` for a behavioural correction with a version bump. Nothing to cut. `net: 0 lines`.

betashop:
Row rebalanced: the errored-vs-unapproved split now lives in the remedy column. And the runtime half is filed so this cannot close the incident item on its own: Senpi-ai/senpi-trading-runtime#402 — explicit denial on timeout, pending approval visible in chat, "Deploy all" counts as consent. Linked from the PR body.

PR #572, opened 2026-09-10
author: advisory warnings (stop distance, multi-slot sizing, daily cap) + shadow-testing reference (3.3.0)

betashop:
## Why

A package can pass every check we have and still behave, from the user's side, like a broken strategy in its first week: a hard stop that sits inside a wick, three slots the account can only fund one of, a daily entry cap that silently pauses the runtime. None of that is an *error* — the package runs — so the validator had no way to say it. And the most expensive way to "test" a strategy (scheduling the agent to watch it) had no reference telling authors why not.

## What changed

**`scripts/validate_strategy.py` — a second channel, `warnings()`.** Advisory, printed under the verdict, never the exit code. Three findings, each worded to be relayed to the user:

| tag | fires when | says |
|---|---|---|
| `[stop]` | `max_loss_pct ÷ leverage ≤ 1.0%` of price (named preset from `dsl-presets.yaml` or inline; leverage from `default_leverage`, `leverage_multipliers`, or scanner `leverage`/`maxLeverage`/`leverageTiers` in any shape) | the distance in price, that it is inside intraday noise, and the three ways out |
| `[sizing]` | `slots × margin > 100%`, or a multi-slot scanner with no free-margin read (`withdrawable` / `free_margin` / … — the camel `_get_positions` gate) before it emits | which slots can never fund; that the 2nd+ open lands `position_open_failed` every tick |
| `[cap]` | `risk.guard_rails.max_entries_per_day` is set | that the runtime logs `Runtime paused: Max Entries/Day` and opens nothing until 00:00 UTC — its own rule, not a fault; extra note when the cap ≤ slots |

Catalog sweep at the 1.0% threshold: 21 of 109 packages get a `[stop]` note (e.g. 8% ROE at 17x = 0.47% of price), 1 is over-committed, 24 multi-slot scanners carry no free-margin gate. All advisory.

**`references/shadow-testing.md`** — the two cheap shadows (`senpi validate` tick, floor-budget wallet), why an `openclaw cron` job is a model call per firing (a 5-minute monitor = 288/day, each paying the whole context), read-on-demand commands, `nohup … & disown` if a process is unavoidable, and the paper-vs-live labeling rule.

**`SKILL.md`** — two lines extended in place (lint step relays the warns; the edit section points at shadow-testing instead of cron polling). Body stays at the 386-line budget. `creating-a-strategy.md` §9 gets the same pointer. `senpi-strategy-author` 3.2.1 → 3.3.0 (frontmatter + README).

## Tests

`tests/test_validate_warnings.py` (13 cases): each warning on/off, the arithmetic in the message, unknown leverage/preset stays silent, and the wall between channels (`validate()` green while `warnings()` is loud; `main` exits 0 with them printed).

```
python3 -m pytest senpi-strategy-author/tests senpi-strategy-ops/tests/test_skill_surface.py -q
201 passed
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | [`senpi-account-status`](senpi-account-status/) | 1.2.0 | Points, loyalty tier, fees, referrals |
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.19.0 | Conversational picker — rank the catalog against your worldview |
-| [`senpi-strategy-author`](senpi-strategy-author/) | 3.2.1 | Build/edit a DSL-protected strategy package, one decision at a time |
+| [`senpi-strategy-author`](senpi-strategy-author/) | 3.3.0 | Build/edit a DSL-protected strategy package, one decision at a time |
 | [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.7.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.0.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
 | [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.1 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -245,7 +245,7 @@ the catalog entry, then unit-test → lint → `senpi validate` → hand to ops.
 8. **Lint — advisory, instant, no credentials** (pass the package's absolute path,
    `/data/workspace/strategies/<id>`, so they hit the authored copy from any CWD):
    (a) **authoring lint** → `python3 senpi-strategy-author/scripts/validate_strategy.py /data/workspace/strategies/<id>`
-   (candle keys, null-in-schema, mandate description, retention/cooldown bounds) **+ advisory warns you relay to the user**: the stop's distance in price at the recipe's leverage, multi-slot sizing with no free-margin gate, the daily entry cap and what it does;
+   (candle keys, null-in-schema, mandate description, retention/cooldown bounds) **+ advisory warns you relay to the user**: the stop's distance in price at the recipe's leverage, multi-slot sizing with no free-margin gate, a daily entry cap at or below the slot count;
    (b) **universe gate** → `python3 senpi-strategy-ops/scripts/validate_universe.py /data/workspace/strategies/<id>`
    — every hardcoded ticker you TRADE must be a live HL instrument (derived universes, and names under an exclusion key, pass trivially);
    (c) **deploy contract** → `python3 senpi-strategy-ops/scripts/deploy.py validate /data/workspace/strategies/<id>`
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -13,7 +13,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.2.1"
+  version: "3.3.0"
   platform: senpi
   exchange: hyperliquid
   requires:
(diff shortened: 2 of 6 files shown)


shnoodles:
**Review: the mechanism is right and two of the three warnings are excellent. The third makes the channel unusable, and I can show that with numbers.**

I ran this PR's `warnings()` over all **109 shipped catalog packages**:

```
packages scanned          : 109
packages with >=1 warning : 109   (100%)
total warnings            : 184
   [cap]    134
   [sizing]  25
   [stop]    25
```

**Every package in the catalog warns.** A channel that fires on 100% of inputs stops being read by the third package, and it will train the author skill to relay warnings as boilerplate — which is precisely how the stop-distance finding gets ignored.

### `[stop]` and `[sizing]` are doing real work — keep them exactly as they are

23% hit rate each, and the hits are the incident:

```
bison   : hard stop 0.80% of price (max_loss_pct 8% ROE at 10x)
cheetah : hard stop 0.71% of price (max_loss_pct 10% ROE at 14x)
```

That is the −15% ROE at 5–10x pathology, caught statically, before funding. `max_loss_pct ÷ leverage`, with "more leverage is a TIGHTER stop" stated in the constant's comment, is the single most valuable line in this PR — it is the thing nobody reasons about correctly under pressure.

### `[cap]` is the problem: it is unconditional

```python
cap = _daily_cap(rt_doc)
if cap:
    msg = (...)  # always appended
```

Setting `max_entries_per_day` is **normal, correct practice** — the catalog does it 134 times. The message even concedes this ("That is the strategy's own rule, not a fault"), which is the tell: it is a **narration reminder**, not a warning, and it does not belong in the same channel as "your stop sits inside the noise floor."

The genuinely warnable case is already computed two lines below — `cap <= slots`, where the cap fills the book once and every re-entry after a stop-out waits for UTC midnight. Narrowing to that:

```
[cap] today (any cap set)     : 134
[cap] narrowed to cap <= slots:  23   (17% of current)
   ant/main      cap=5 slots=5      condor/main   cap=1 slots=1
   armadillo     cap=3 slots=4      caribou/long  cap=8 slots=8
   bison/main    cap=3 slots=3      chimp/main    cap=8 slots=8
```

`condor: cap=1 slots=1` is a package that can open exactly one position per UTC day — that is worth saying out loud. `albatross: cap=5` with room to spare is not.

**Suggested change:** make `[cap]` fire only when `cap <= slots`, and move the unconditional "say the cap in the How-it-runs summary" line into the summary template where it belongs. That takes the channel from 184 warnings on 109 packages to ~73 on maybe 40 — a number an author will actually read.

### Two smaller notes

**`WARN_STOP_PRICE_PCT = 1.0` is a crypto-perp heuristic applied to everything.** The constant's comment says "on most perps", but the catalog now carries `xyz:` HIP-3 equities and commodities, where 1% of price is a different animal from 1% on a perp. Advisory-only means a false positive is cheap, so this is not blocking — but the message asserts "inside ordinary intraday movement" as fact. Consider softening the wording for non-crypto assets, or scaling the threshold by asset scope.

**`_FREE_MARGIN_RE` only reads `scan.py`.** I tested the false-positive risk — a multi-slot package that computes free margin in a sibling module would warn spuriously — and it is **zero across all shipped packages today**, so this is not a current defect. Flagging it only because customer-authored packages routinely split into `scoring.py` (I read one this week doing exactly that), so the first false positive will come from an authored package rather than a catalog one. Globbing `scanners/*.py` instead of `scan.py` is a one-line hedge.

### One cross-PR inconsistency worth resolving

This PR adds `references/shadow-testing.md` and legitimises "shadow-testing" as the supported workflow (validate = the shadow run). senpi-agent #63 and skills #575 use "shadow" as the **thing to ban** ("never build a sim/backtest/'shadow' harness"). Both are pushing the same direction, but the word now means two opposite things across always-loaded docs. Pick one — I would keep "shadow" for the supported cheap path here, and have #63/#575 ban "cron-driven watching" by mechanism rather than by that word.

Ponytail pass: `+442/-5`, and the helper ladder (`_num` / `_nums` / `_preset_max_loss` / `resolved_max_loss_pct` / `max_leverage` / `slot_plan` / `_daily_cap`) is justified — each reads a different real catalog shape, and `_nums` handling scalar/list/dict exists because the catalog genuinely uses all three. No speculative flexibility. `net: 0 lines` (the `[cap]` change removes warnings, not code).

betashop:
Taken, with numbers. `[cap]` now fires only when `max_entries_per_day <= slots` (the book fills once, every re-entry waits for UTC midnight); the unconditional "say the cap in the summary" line lives in the How-it-runs template (ops #573) where it belongs. Free-margin idiom is searched across every module in `scanners/` (authored packages split the read into a sibling file — test added). The `[stop]` message now says "for a crypto perp this is inside ordinary intraday movement" and tells the author to judge an `xyz:` equity/index/commodity against that market's own daily range.

Catalog sweep, same 109 packages:

| | before | after |
|---|---|---|
| packages with ≥1 warning | 109 (100%) | 48 (44%) |
| total warnings | 184 | 65 |
| `[cap]` | 134 | 23 |
| `[sizing]` | 25 | 17 |
| `[stop]` | 25 | 25 |

On the word "shadow": settled the way you suggested — `shadow-testing.md` keeps it for the supported cheap path (`senpi validate` IS the shadow run); senpi-agent #63 and #575 now ban the mechanism (anything that re-enters the model on a timer) and no longer use the word.

PR #573, opened 2026-09-10
ops: deploy.py update wrapper, risk-gate pause + config drift in status.py, approval-gate rule (3.8.0)

betashop:
> **Stacked on #546 + #549** (merged into this branch so the ops edit section is the `senpi update` one). Until those land, this PR's diff shows their commits too; after they land it is only the commits below. **Ship after the runtime release that carries `senpi update`** (senpi-trading-runtime #397 / #342) — same ordering caveat as #546.

## Why

Three things the churn analysis kept asking this skill for, and none of them needs the runtime to change:

1. **An edit path that cannot become a redeploy.** Agents routed to "apply the edit" had one tool that worked, delete-and-create, which market-exits the book and drops the stops — and on a box without the `update` verb the raw CLI answers with a Commander parse error that reads like nothing happened.
2. **"My strategy is dead" when it is paused.** A runtime held by its own daily cap reports `health: healthy` — it ticks, it manages what it holds, it opens nothing. `status.py` printed ✅ and nothing else, so the user's next move was a redeploy that restarts the same gate from zero.
3. **"Did my edit ever land?"** Nothing on the read surface compared what the runtime is *running* with what is on disk.

## What changed

**`deploy.py update <pkg> --id <runtime_id> [--apply] [--code-only] [--json]`** — plans by default, `--apply` commits. Around the verb: the structural preflight `create` runs (refused packages never reach the verb), the instance dir resolved from `--id` (required on multi-instance packages, implied on one-instance ones; a selector is always sent), `--apply` refused locally when no proof is on disk (names the exact `openclaw senpi validate <dir>`), the verb's stdout/stderr and exit codes relayed verbatim (0/1/2/3), and on a pre-verb runtime: exit 1, "the edit is on disk and NOT applied — do NOT close and redeploy". It fetches nothing and can never delete, create, fund or close.

**`status.py`** — two facts beside health on every deep row (`--fast` skips both):
- `paused` — `components.risk` of the fleet-wide `status --json` says `CLOSED`/`COOLDOWN`: the row shows `⏸ CLOSED`, a section quotes each gate's `reason` and when it lets go (daily gates: 00:00 UTC; cooldowns: on their own), and says never to redeploy over it. Health is untouched.
- `config_drift` — one `runtime list --json` for the rendered descriptors; compared with the package on disk (`dsl_preset` name, collapsed `description`, and the recipe hash when the runtime publishes `recipeHash`). Differences print with the `deploy.py update` fix. Fail-open: package not on disk / no descriptor → nothing claimed.

**`_cli.py`** — `risk_pause`, `describe_pause`, `runtime_descriptors`, `recipe_hash` (normalized: blank/comment lines out, trailing space off, `wallet`/`model`/`decision_model` lines out — deploy rewrites those), `config_drift`.

**`SKILL.md`** (body 297 ≤ 300 budget): How-it-runs gains the daily-cap line; Monitor names ⏸ paused and ✎ drift; the trade-approval rule (a trade tool the user does not approve in time is denied and returns **empty** — not executed, never "done", never re-issued in a loop); the edit section routes through the wrapper. `references/editing-a-live-strategy.md` + `lifecycle.md` updated.

**Versions** — `senpi-strategy-ops` 3.7.0 → 3.8.0; also `senpi-trading-runtime` skill 4.0.1 → 4.0.2 (its SKILL.md changed in #546 with no bump, so the fleet would not have picked it up). README table synced.

## Tests

`tests/test_deploy_update.py` (11): gate-before-verb, instance-dir + selector, no-proof refusal, apply/code-only/json passthrough, unknown `--id`, multi-instance needs `--id`, verb refusal/failed relay, the pre-verb box stop, not-on-disk never fetched.
`tests/test_status_risk.py` (14): pause parsing (CLOSED/COOLDOWN/OPEN/absent), health untouched, recipe hash normalization, drift on preset/description/hash, text render (row hold + section + fix command), and `build(deep=True)` making one health call and one descriptor call.

```
python3 -m pytest senpi-strategy-ops/tests senpi-strategy-author/tests senpi-trading-runtime/tests -q
576 passed, 1 skipped
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,9 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.19.0 | Conversational picker — rank the catalog against your worldview |
 | [`senpi-strategy-author`](senpi-strategy-author/) | 3.2.1 | Build/edit a DSL-protected strategy package, one decision at a time |
-| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.7.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
+| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.8.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.0.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
-| [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.1 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
+| [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.0.2 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
 | **Move money / positioning** | | |
 | [`senpi-deposit-withdraw-transfer`](senpi-deposit-withdraw-transfer/) | 1.1.0 | The money-movement rails (funds in via embedded wallet or in-app USDC purchase; out via the app) |
 | [`senpi-why`](senpi-why/) | 1.0.3 | "Why Senpi / vs. other tools" — the positioning answer |
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -352,10 +352,10 @@ loop every time:
 > deployed package — "make my live strategy more aggressive", re-tune, re-score) — then hand it to
 > **`senpi-strategy-ops`**, which applies it IN PLACE with `openclaw senpi update`: no close, no fresh
 > wallet, no market exit. **Re-running `create` will NOT apply it** — the deploy verb is idempotent, so it
-> adopts the existing wallet and leaves the deployed scanner as it is. Tell the user two things: exit
-> changes are **forward-only**, so a tightened `dsl_preset` governs new entries and never reaches a
-> position already open; and a changed `strategy.wallet`, a renamed or moved scanner or a changed
-> `action_type` still forces a close-and-redeploy — a market exit. Below is the not-yet-live path.
+> adopts the existing wallet and leaves the deployed scanner as it is. Tell the user two things:
+> `dsl_preset` is **forward-only** — new entries only, never a position already open (other `exit:` fields
+> like `order_type` DO reach open ones); and a changed `strategy.wallet`, a renamed or moved external
+> scanner or a changed `action_type` still forces close-and-redeploy — a market exit. Below: the not-yet-live path.
 
 1. **Confirm with the user** — budget + "ready to deploy?" Funding a wallet is real money and one-way, so
    this is an explicit yes, not an assumption.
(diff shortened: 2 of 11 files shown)


shnoodles:
**Review: yes — and this PR quietly supplies the safety property #546 is missing.**

Ran it: `test_deploy_update.py` + `test_status_risk.py` **25 passed**, whole ops suite **396 passed, 59 subtests**.

### The version guard is the important part, and it deserves to be called out

#546 routes every live edit through `openclaw senpi update` — a verb that does not exist until the runtime release lands — and its `requires:` carries no minimum version. This PR is what makes that safe:

```
deploy.py update … → exit 1, "the edit is on disk and NOT applied"
```

instead of a Commander parse error that reads like nothing happened. And the detection is done properly. Matching on Commander's **parse-error grammar**, anchored to the line Commander writes, rather than scanning the payload for `deploy`, with the reason documented:

> The root action is why: Commander dispatches `deploy status` INTO it instead of raising unknownCommand, so neither "unknown command" nor even the word "deploy" appears on any box in the fleet. Matching those (what this did) detected only pre-2026-03-27 runtimes — nothing that exists.

That is a real investigation, and the `_REFUSAL_CODE` guard (a bracketed `[CODE]` means the verb spoke, so a refusal quoting a parser error is never misread as a missing verb) is the kind of thing that only gets written after being bitten once. Keep the comments.

**Merge-order consequence:** #546 → #573, and ideally close together. Between them, the skills tell an agent to run a raw `openclaw senpi update` with no degradation path; after #573 the instruction points at the guarded wrapper. If #546 ships and #573 slips, the gap is exactly the window where an agent on an un-upgraded box gets a parse error and improvises — and the improvisation is close-and-redeploy, a market exit.

### One thing to change: the drift check rests on a cross-repo constant that only one side pins

`status.py`'s **✎ running recipe ≠ disk** is the "did my edit ever land?" answer, and it compares `_cli.py`'s `recipe_hash` against the runtime's `recipeHash` from senpi-trading-runtime #399.

I executed both sides against #399's fixture — they agree today, on the authored form *and* on the CRLF/comment-stripped/wallet-rewritten installed form:

```
TS-pinned digest  : 0ee8ddd8870febccca803a23bb959b98556dbbd93fc26fbeb80af522600fbbb4
python recipe_hash: 0ee8ddd8870febccca803a23bb959b98556dbbd93fc26fbeb80af522600fbbb4  MATCH
```

But the digest is pinned **only in the TS test**, in the other repo. Nothing here asserts it. Widen `_RECIPE_DROP_LINE` (add `builder_fee:`, say) and every test in both repos still passes while `status.py` starts reporting **✎ running recipe ≠ disk** for every runtime in the fleet — a drift detector that cries drift on everything, feeding the exact "your edit never applied" anxiety this PR exists to answer.

Two lines in `test_status_risk.py` pinning the same fixture and digest closes it. I would not merge the pair without it.

### Smaller

- "an unapproved call times out, is **denied**, and returns an **EMPTY** result" duplicates #571's wording deliberately — good, they should not drift. Worth a comment in one pointing at the other so a future edit updates both.
- The `--code-only` refusal when the recipe moved as well is a nice touch: it turns "I only changed scan.py" from a claim into an assertion.

Ponytail pass: `+809/-44`, and the largest single block is the ~40 lines of comment above three regexes. Normally I would call that out; here the comments encode why the obvious implementation (grep for "deploy") detects nothing that exists in the fleet, which is exactly the knowledge that gets lost and re-learned. Keep it. `net: 0 lines`.

betashop:
Parity pin added: `test_status_risk.py` now asserts `recipe_hash` over the identical fixture equals the digest the runtime pins in `runtime-descriptor.test.ts` (`0ee8ddd8…`), so the contract is asserted from both ends. The approval-gate line names senpi-trade's guardrail row as its twin so an edit updates both. Merge order agreed: #546 → #573, close together — and note Sarvesh's finding on runtime #399: an `update --apply` currently re-serialises the recipe, so the ✎ drift check would cry drift after every update until his follow-up (splice only the wallet scalar) lands; the name/description compare is unaffected.

PR #574, opened 2026-09-10
portfolio: one fleet status call for runtime health + risk-gate pause beside health (1.18.0)

betashop:
## Why

Two things the `strategies` step got wrong for a user with several runtimes:

1. **It could not finish inside the exec window.** Runtime health was one `openclaw senpi status -r <id> --json` per runtime, sequentially, at a plugin start each (~2-3 s). The step's cost grew with the runtime count, the exec timed out, and the SKILL's own warning came true: the agent fell back to raw MCP and lost every guardrail.
2. **A paused strategy read as a dead one.** A runtime held by its own daily entry cap reports `health: healthy` — it ticks, it manages what it holds, it opens nothing. Narrated as `live`, the quiet reads as broken; the user's next move was a redeploy that market-exits the book and restarts the same gate from zero.

## What changed

**One fleet call.** `_fleet_statuses(meta)` reads `openclaw senpi status --json` once per run (retried once — the gateway is flaky-empty; a progress line goes to stderr, flushed, so a streaming exec shows what it is waiting on). `_fetch_runtime_status` slices it per runtime **in the shape the per-id call prints**, so `_liveness_from_status` is untouched. A failed fleet read falls back to the per-id read for that runtime — a box where the fleet form misbehaves loses nothing it had. The `SENPI_STATUS_FIXTURE` per-id contract is unchanged.

**`risk_pause`** on each strategy and each `strategy_groups[]` entry, read off the same document as liveness: `{"eligibility": "CLOSED"|"COOLDOWN", "gates": [{gate, id, reason, reset}]}` while the runtime's own guard rail holds entries, else `null`. Never a health downgrade. `meta.paused_by_risk_gate` lists them.

**SKILL.md** — the `risk_pause` narration rule (name the gate, quote its reason, say when it lets go, never redeploy over it), "paper is not live" (a validate tick / signal / estimate is not a trade), and the one-call note. `senpi-portfolio` 1.17.0 → 1.18.0 (frontmatter + README).

## Tests

Five new cases in `tests/test_portfolio.py`: the per-runtime form is never called when the fleet call answers (exactly one `status --json`); a failed fleet read falls back to `-r`; a runtime the fleet document does not list is `unknown`, never `live`; a CLOSED gate is reported beside `live` (OPEN gates not listed, reason + reset carried, group roll-up, meta list); OPEN gates and runtimes without a risk component are not paused.

```
python3 -m pytest senpi-portfolio/tests -q
84 passed
```

Pairs with #573 (`status.py` prints the same pause as ⏸) and #571 (the approval-gate rule).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | Skill | Ver | Role |
 |---|---|---|
 | **Analyze** | | |
-| [`senpi-portfolio`](senpi-portfolio/) | 1.17.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
+| [`senpi-portfolio`](senpi-portfolio/) | 1.18.0 | All-wallet portfolio, positions, DSL protection, per-strategy mandate reads |
 | [`senpi-market-pulse`](senpi-market-pulse/) | 1.2.0 | Daily cross-asset market read (crypto, equities, commodities, macro, funding regime) |
 | [`senpi-smart-money`](senpi-smart-money/) | 1.2.0 | Where the most-profitable wallets are positioned vs. the crowd |
 | [`senpi-trader-research`](senpi-trader-research/) | 1.4.0 | Rank + vet Hyperliquid traders before copying them (mirror-aware: copyability, min-budget, live book) |
--- a/senpi-portfolio/SKILL.md
+++ b/senpi-portfolio/SKILL.md
@@ -16,7 +16,7 @@ description: >-
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "1.17.0"
+  version: "1.18.0"
   platform: senpi
   exchange: hyperliquid
   requires:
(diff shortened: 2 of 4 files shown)


shnoodles:
**Review: yes, all three parts land, and the fleet-status change is the quiet win.**

- **`risk_pause` beside `runtime_health`, never instead of it** — correct modelling. A strategy stopped by its own daily cap is *healthy and paused*, and collapsing that into `degraded` is what produced "hornet was paused ten minutes after going live and the agent could not find the cap when asked, twice". Quoting the gate's `reason` verbatim and naming the reset (daily gates at 00:00 UTC) is what makes it actionable instead of alarming.
- **"Never send the user to redeploy over it — a fresh wallet market-exits the book and starts the same gate again from zero."** That sentence is worth the PR on its own. It is the exact loop the incident user was put through.
- **"Paper is not live"** — right place, right words. `senpi validate` ticks and shadow signals being narrated as live Mako trades is one of the 14 misreports.
- **One fleet `status --json` with per-id fallback** — the fallback is what makes this safe: a box where the fleet call misbehaves loses nothing it had before, and the fixture path keeps its old contract so existing tests still mean what they meant.

Two small things to change:

**1. An empty fleet is indistinguishable from a failed read.** In `_fleet_statuses`:

```python
recs = _status_entries(data) if isinstance(data, dict) else []
if recs:
    meta["_fleet_statuses"] = recs
    return recs
```

A user with genuinely **zero** runtimes returns `[]`, which is falsy, so the loop retries the whole call and then returns `None` — and every subsequent runtime falls through to the per-id path. Harmless today (no runtimes means nothing to look up), but it makes "no runtimes" cost two extra gateway calls and log a misleading progress line. Distinguishing "parsed fine, empty" from "did not parse" is a one-line change and keeps the retry for the case it was written for.

**2. `_record_is` matches four keys, and the docstring says two.** The comment says the record carries `runtimeId` (optional) and `runtimeName` (always), but the match also accepts `id` and `name`. If those two are real aliases the comment should say so; if they are defensive guesses, they are a way for an unrelated record shaped `{name: ...}` to be claimed by the wrong runtime. Given this feeds `runtime_health` for a *funded* strategy, I would narrow it to the two documented keys unless the others are confirmed.

Ponytail pass: `+200/-3`, and the largest chunk is the fleet-status function that removes an O(runtimes) sequential-plugin-start cost. That is a deletion in disguise. `net: 0 lines`.

betashop:
1. `_record_is` narrowed to `runtimeId` / `runtimeName` — the two keys `RuntimeHealthStatus` actually carries; `id`/`name` were defensive guesses, and this decides a funded strategy's health.
2. On the empty fleet: the function is only reached for a runtime the registry lists, so a user with zero runtimes never pays for the call at all. An empty `statuses[]` while the registry lists runtimes is the gateway's known flaky-empty answer (the same one `_cli.list_runtimes` documents) — that is exactly the case the single retry was written for, and after it the per-id read gets its own chance. Kept the behaviour, wrote that down in the docstring so the next reader does not re-derive it.

PR #575, opened 2026-09-11
skills: crons cost tokens, and there is no paper-trading mode — in every skill that could schedule or shadow a strategy

betashop:
> **Stacked on #571, #572, #573 and #574** (merged into this branch so versions stay monotonic and the author body stays at its budget). Until those land the diff shows their commits too. One thing for whoever merges the four: their README rows are adjacent lines, so the second one in needs a one-line README rebase.

## Why

An `openclaw cron` job is an **agent turn**: every firing is a full model call over the whole conversation. A five-minute "shadow scan" is ~288 model calls a day, watching a strategy the runtime already supervises at zero model cost. The transcript behind this shows the agent creating three such jobs on its own initiative — a daily check-in it proposed itself, a 10-minute "compounding engine" halt, a 5-minute shadow scanner — with the runtime, author and portfolio skills loaded and silent on cost. The asks that led there ("monitor it", "a measurement framework", "I don't want to touch production, run it in shadow", "a daily check-in") all route through the skills below.

## The rule, where each ask lands

| skill | where | says |
|---|---|---|
| senpi-strategy-author 3.4.0 | description · funding heads-up · `references/shadow-testing.md` | **There is no paper-trading mode.** Test = `senpi validate` (one real tick, no wallet), then a live run at the $10 floor. A cron-driven "shadow" scan is a model call per tick, not a simulation. |
| senpi-strategy-ops 3.9.0 | description · Monitor | Never an agent-turn cron to watch a strategy; the runtime supervises it at zero model cost, read on demand. A check-in the user wants: at most 1–2/day, cost said first, a yes before creating it. |
| senpi-strategy-discover 2.20.0 | Layer 3 sizing | "Can I try it first / paper trade it?" → the floor budget IS the trial, after `senpi validate`; never a scheduled shadow. |
| senpi-trade 1.2.0 | hero check · mirror reference | The sizing estimate is one-shot, not paper trading; the live test is the $10 floor. |
| senpi-why 1.1.0 | supporting facts | No paper mode by design; the runtime watches at zero model cost — no token bill for "monitoring". |
| senpi-portfolio 1.19.0 · senpi-market-pulse 1.3.0 · senpi-smart-money 1.3.0 · senpi-improve-trades 1.9.0 | run-it / golden rules / hard rules | Asked to run this on a schedule? Say the cost first (hourly = 24 calls/day, 5-minute = 288), offer 1–2/day at most, get a yes; never a cron to watch strategies. improve-trades adds: a "measurement framework" is read from the runtime's own events, never a cron. |
| senpi-trading-runtime 4.0.3 (PATCH) | contract | Zero model cost, the only loop there is; never a producer/shadow/watcher cron beside it. |

Body budgets hold: ops 298/300, author 386/386. README rows synced.

## Pairs with

- senpi-agent #63 — the always-loaded workspace docs (`TOOLS.md` recommended cron for recurring jobs; `AGENTS.md` said sim/backtest harnesses are fine). That is the surface that actually pushed the agent toward crons; this PR is the skill side.
- senpi-trading-runtime #400 — the CLI cheatsheet ("Testing a strategy?") and the liveness reference (a producer must not be an agent-turn cron).

## Tests

```
python3 -m pytest senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-strategy-author/tests senpi-trading-runtime/tests senpi-trader-research/tests senpi-market-pulse/tests senpi-portfolio/tests strategies/tests -q
878 passed, 9 skipped
```


shnoodles:
**Review: the content is right and consistent. The packaging is the problem — this is not a standalone PR, and its stat line hides that.**

### This PR is the tip of a four-deep stack, presented as based on `main`

```
pr572 IS an ancestor of pr575
pr573 IS an ancestor of pr575
pr574 IS an ancestor of pr575
pr546 is NOT an ancestor of pr575
```

GitHub reports `base=main`, so the `+1497/-70` diff against main **contains all of #572, #573 and #574**. I confirmed by content, not just ancestry: #572's `def warnings(pkg` , #573's `_cli_rejected_the_command` and #574's `_fleet_statuses` all appear as additions in this PR's diff.

Two consequences:

1. **Merging #575 first lands #572, #573 and #574 wholesale**, bypassing their own review. #572 in particular has an open issue from me (its `[cap]` warning fires on 100% of the catalog — 134 of 184 warnings) that would ship silently through this door.
2. Anyone reviewing #575 on its own is reviewing ~1300 lines they have already reviewed elsewhere, or — worse — has not.

**What to change:** retarget this PR's base to `feat/ops-update-wrapper-risk-pause` (or whichever is last in the intended chain) so the diff shows only its own contribution, or merge strictly 572 → 573 → 574 → 575 and say so on the PR. Right now nothing in the UI communicates the dependency.

### The content itself is good, and it is genuinely consistent

I checked the numbers across every copy of the recurring paragraph — the arithmetic does not drift:

```
3 × "every hour" is 24 model calls a day
3 × "every 5 minutes" is 288
```

Repeating the paragraph in each skill is the correct call, not a DRY violation: skills load independently, so text absent from `senpi-market-pulse/SKILL.md` does not exist for an agent working in market-pulse. The duplication is the delivery mechanism.

But it does create a maintenance edge: a future correction has to land in every copy, and the first divergence will be silent. The repo already has a good pattern for this — a test pinning a constant that must stay equal across files. A three-line test asserting the cron paragraph's numbers are identical everywhere they appear would make that free.

### One thing to narrow — same point I left on senpi-agent#63

`references/shadow-testing.md:9-10`:

> Say that plainly when the user asks to paper trade, **backtest** or "run it in shadow"

"Backtest" does not belong in that list. A cron-driven shadow scan is a model call per tick — that is the thing that cost the credits. An offline backtest over historical candles is a one-shot computation: no scheduler, no per-tick model call, no venue contact.

The 09-10 funnel report has the counterexample from the same week: M417794 ran a nine-turn authoring interview, built a package, passed `senpi validate`, **backtested 90 days**, then optimised across 21,600 and 58,800 parameter combinations to v1.3.0 — written up as "the best funds-blocked case the series has seen" and "the highest-value outreach in the report". Under this wording the agent talks them out of it.

Ban the mechanism, not the word: *"never anything that re-enters the model on a timer"* covers the failure and leaves offline computation alone.

### Smaller

`senpi-strategy-author/SKILL.md` now says "`senpi validate` is the shadow run" while this file says a "shadow" scan is the thing to refuse. Same word, two meanings, both in always-loaded docs — worth one pass to settle which sense "shadow" carries.

Ponytail pass on #575's own contribution (23 files once the stack is discounted): docs, no speculative structure, and the repetition is load-bearing rather than redundant. `net: 0 lines` — the cut I would make is to the *diff*, by retargeting the base, not to the content.

betashop:
Taken on all three.

**Packaging.** Converted to **draft** and it stays stacked: merge **#571 → #572 → #573 → #574** first (each is self-contained and reviewed on its own), then I flip this one to ready and its diff against `main` collapses to its own 23-file contribution. The branch has been refreshed with the review fixes on all four, so the tip is consistent with what those PRs now say. Retargeting the base was not possible — the base would have to be a branch that already contains all three, which is only this one.

**"backtest" / "shadow".** Removed "backtest" from the refusal list in `shadow-testing.md`; the file now says an offline one-shot backtest over historical candles is fine (no scheduler, no model call per tick, no venue contact) as long as it never wires to a trade tool, and it bans the mechanism — anything that re-enters the model on a timer. "shadow" keeps one meaning fleet-wide: the `senpi validate` run. The ban copies in author, discover, trade, improve-trades and trading-runtime no longer use the word; senpi-agent #63 was narrowed the same way.

**Drift guard.** `test_skill_surface.py` gains `CronCostArithmeticAgreesEverywhere`: it collects every "… is N model calls" number across all skills and references and pins hourly = 24, five-minute = 288, ten-minute = 144. Suite: 881 passed.

PR #576, opened 2026-09-11
feat(deposit): route funding to the card

im-vignesh:


senpi-deposit-withdraw-transfer/SKILL.md
```diff
@@ -44,54 +50,63 @@ Everything the agent does with tools happens **strictly between the user's own S
...
+- **Show the card; don't describe the plumbing.** Call `show_widget` with
+  `widget_type: "fund_user_wallet"` whenever money should arrive in the user's own Senpi wallet — an
+  explicit deposit ask, an empty wallet, or a balance too short for what they asked. The user picks the
+  token and network **in the card**, which shows the matching deposit address and QR.
+- **NEVER write a deposit address, QR code, or network list in chat.** The deposit address is
+  **network-scoped** ("only send USDC on Base to this address") — repeating an address or implying it
+  works on any chain can lose funds. Do not read the address out of `user_get_me` either: the wallet
+  list there is identity data, not a deposit target.
+- **Where deposits land:** as USDC on Hyperliquid in the user's own funding (embedded) wallet, ready to
+  trade — no separate transfer step afterwards.
```
shnoodles, line 79: **The deposit floor is missing, so "what's the minimum to deposit?" gets the wrong number.**

The only minimum in the skill is the **$10** on line 106, and it's scoped to the **Buy USDC** tab — correct there, and it matches the $10/wallet strategy floor. But the **Deposit Crypto** tab's own floor is **$8** (`MIN $8` in the modal), and nothing states it. Asked "what's the minimum I can deposit?", the agent answers $10.

Both numbers are real and they're different things, which is the part worth writing down:

```suggestion
- **Where deposits land:** as USDC on Hyperliquid in the user's own funding (embedded) wallet, ready to
  trade — no separate transfer step afterwards.
- **Two different minimums — don't merge them.** The **deposit** floor is **$8** (the card's own `MIN $8`);
  a **strategy** still needs **$10** to open. So "what's the minimum deposit?" is $8, but "what do I need
  to start trading?" is $10 — quote the $10 whenever the goal is running something.
```

Minor, since the card shows `MIN $8` itself — but a user who deposits exactly $8 on that advice then can't open a strategy, and that's the confusing half.

shnoodles (commented):
One gap on minimums. Everything else on the address rule checks out — repo-wide, every remaining `user_get_me` mention is a prohibition, so the "never write an address" rail is actually closed and not just asserted.

shnoodles (approved):

PR #578, opened 2026-09-11
ops: the walkthrough before the budget question; every template deploys as `<User>'s <Template>` (3.12.0)

betashop:
## What

**Step 0.75 — the walkthrough**, required before the budget question. After `deploy.py validate <id>` has fetched the package, ops reads `strategy.yaml` and each `runtime.yaml` and tells the user, in plain language: what it does and how it is set (the same four lines as the post-live "How it runs" block, moved *before* the money moves), **two levers** worth a look, the **fee load and the design budget**, and **the name**: every template deploys as `<User>'s <Template>` (`PurpleFrog's Starling`, or a name of the user's own), as-is or with levers moved. Then one question: as-is, or make it yours? One word skips it; never a gate; cost classes stated as facts, never as discouragement. Named installs get it too.

**Resident vs reference.** The rule is 14 resident lines in SKILL.md; the depth is `references/walkthrough.md`: the lever table ranked by catalog coverage — `strategy.slots`, `strategy.default_leverage`, `exit.dsl_preset.phase1.max_loss_pct` (ROE *and* price at leverage), `risk.guard_rails`, `interval_seconds` first (100% of the catalog), `strategy.margin_pct` as the biggest felt lever after the stop (about half), scanner-side `maxSlots`/`maxLeverage`/`leverageTiers` read where present and never looked for first — the name rules (possessive, username never an id, a second fork asks), the cost classes, the two consent sentences, and **the fork on disk** (until a `fork` verb ships): copy to `<strategies root>/<template>-<user-slug>/` — the root `deploy.py where` prints, never a CWD-relative `strategies/` — leaving `.deploy-state.json` and `__pycache__` out, then `id`, `catalog.name`, `forked_from` lineage, `group`/`name` linkage, the mandate prefix, values-only lever edits, and the unchanged gate on the **directory**.

**`deploy.py where`** (new verb, review): prints `_pkg.strategies_root()` — `SENPI_STRATEGIES_DIR`, else `$OPENCLAW_WORKSPACE_DIR/strategies`, else `/data/workspace/strategies` — so the fork destination is read, never composed. Dispatched before the flag-reading code (it takes no flags). Two subprocess tests.

**Consent rules** (from the M371275 review): the fee load and the design budget are said in the walkthrough, deploying below the design budget takes an explicit yes, and removing a guardrail or lowering a threshold just recommended gets a one-line consequence and a yes — never "done".

**Post-live**: one ownership line ("this is your strategy, saved at `/data/workspace/strategies/starling-purplefrog/` … tune it any time, it applies in place"). **Live edits**: say *update*, never *redeploy*.

**Body budget**: the ops SKILL body goes 298 → 317 lines; `BODY_BUDGET` set 300 → 317 at the post-edit count with no slack and the dated rationale in the test (the same pattern as the author 368 → 386 bump): the resident lines are what must fire on every template deploy before money moves; everything deferrable went to the reference.

3.11.0 → 3.12.0, README row synced. 407 tests pass.

## Why

Users hear "I deployed Starling for you" and feel they are running our strategy. The walkthrough puts the explanation and the fork before the wallet is funded, and the name makes the version theirs from the first deploy. Author is offered as a peer with its cost class, not a downsell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.22.0 | Conversational picker — rank the catalog against your worldview |
 | [`senpi-strategy-author`](senpi-strategy-author/) | 3.5.0 | Build/edit a DSL-protected strategy package, one decision at a time |
-| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.11.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
+| [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.12.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.3.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
 | [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.1.0 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
 | **Move money / positioning** | | |
--- a/senpi-strategy-ops/SKILL.md
+++ b/senpi-strategy-ops/SKILL.md
@@ -19,12 +19,15 @@ description: >-
   create <id> --budget <usd>` takes a package live end to end (it gates the package,
   then runs the runtime's detached deploy job; watch with `senpi deploy status`);
   close.py tears down (stop runtime + strategy_close → flattens positions,
-  returns funds). The id (spider, polar, kodiak) is the package folder. NOT for choosing WHICH strategy
+  returns funds). Before the budget question ops runs THE WALKTHROUGH (Step 0.75):
+  what the template does, how it is set, the two levers worth shifting, and its
+  name — every template deploys as the user's own version
+  (`PurpleFrog's Starling`, or a name of their own), as-is or with levers moved. The id (spider, polar, kodiak) is the package folder. NOT for choosing WHICH strategy
   (senpi-strategy-discover) or authoring / editing the strategy files themselves (senpi-strategy-author).
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.11.0"
+  version: "3.12.0"
   platform: senpi
   exchange: hyperliquid
   requires:
@@ -92,8 +95,24 @@ fresh `openclaw senpi validate` before the next `create`. **`UNPROVEN` (exit 2) 
 `deploy.py validate <id>` answers the other question, **is the package well formed**. Do not deploy a
 package that has not returned `PASS`. The proof: [`references/lifecycle.md`](references/lifecycle.md).
 
+**Step 0.75 — the walkthrough (REQUIRED before the budget question).** The package is on disk (Step 0.5).
+Before you ask for a dollar, read `strategy.yaml` (`catalog:`) and each instance's `runtime.yaml` and say —
+plain language, no YAML — **what it does and how it is set** (the post-live block's four lines, before the
+money moves), **two levers** worth a look (one for `tier: starter`; each with default, range and effect),
+the **fee load and the design budget**, and **the name**: every template deploys under the user's name,
+levers moved or not — propose **`<User>'s <Template>`** (`PurpleFrog's Starling`); their own word wins.
+Then one question: *"Run it as-is, or shift one of these and make it yours?"* One word skips it; never a
+gate, never re-asked; the budget question comes after the answer. Cost class is a fact beside the choice,
+never a discouragement; named installs get the walkthrough too. Below the design budget, a guardrail
+removed or a threshold lowered: the consequence in one line and an explicit yes — never "done". Which
+levers, the name rules, and **making the fork on disk** — a copy at `<strategies root>/<template>-<user-slug>/`
+(`deploy.py where` prints the root, never a CWD-relative `strategies/`; `.deploy-state.json` left out;
+`id`, `catalog.name`, `forked_from`, linkage, the mandate prefix; then the unchanged gate on the
+**directory**): [`references/walkthrough.md`](references/walkthrough.md).
+
 **Step 1 — start the deploy.** Budget splits across instances by `funding_share`, **min $10 each** (the
-platform wallet floor) — **confirm the amount with the user first**. Two tiers, and only the first
+platform wallet floor) — **ask for the amount now — after the walkthrough, never before it — and confirm
+it**. Two tiers, and only the first
 stops anything: below the $10/wallet floor the deploy **refuses**; a wallet left with less than **its
 own** sizing needs still **deploys**, with a `[W_BUDGET_BELOW_STRATEGY_MIN]` warn to relay.
 ```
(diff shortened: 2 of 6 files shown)


shnoodles:
**Review: achieves the goal — the walkthrough lands before the budget question, the ownership framing is consistent, and "update, never redeploy" is the right word to police.** Jason asked me to check the lever list and the fork mechanics against `deploy.py`'s package-dir resolution, so that is what I did rather than reading the prose.

## 1. The fork destination is the legacy, incident-causing form

```
copy the fetched package to `<strategies root>/<template>-<user-slug>/`
(e.g. `strategies/starling-purplefrog/`)
```

The abstraction is right; the example contradicts it. `_pkg.strategies_root()` resolves in this order:

```
SENPI_STRATEGIES_DIR  >  $OPENCLAW_WORKSPACE_DIR/strategies  >  /data/workspace/strategies
                      >  CWD-relative strategies/   ← legacy, warned loudly
```

I ran it on a host with no env set:

```
⚠ no durable strategies root found (no SENPI_STRATEGIES_DIR, no workspace dir) —
  falling back to CWD-relative strategies/. Packages here may not survive skill updates.
strategies
```

So `strategies/starling-purplefrog/` is the **fallback tier**, and its own docstring names the cost:

> MUST stay outside any managed skill dir: the runtime's skills-manager swap-replaces those dirs on every SKILL.md version bump, destroying anything inside (the 2026-07-30 incident — a CWD-relative dest meant deploys run from a skill dir were wiped on the next bump).

An agent that copies the example literally writes a fork — the user's *named, personalised* strategy, the whole point of this PR — into the directory shape that got wiped once already. On a real agent host the correct path is `/data/workspace/strategies/starling-purplefrog/`.

**Change:** drop the CWD-relative example. Either write `/data/workspace/strategies/<template>-<user-slug>/`, or have the agent ask rather than assume. There is currently **no** `deploy.py` subcommand that prints the root (`create`, `runtime`, `verify`, `status`, `validate`, `update`) — a three-line `deploy.py where` would make this unguessable, and is worth it given a fork now lands there by default.

## 2. Exclude `.deploy-state.json` from the copy

Minor but free. The template package on disk may already carry `.deploy-state.json` — 49% of users deploy templates as-is, so forking a template they already ran is a normal path. Copying it marks the fresh fork as a "legacy DEPLOYED package" for `resolve_pkg_dir`'s tie-break, and arms `ensure_pkg`'s refusal branch (*"carries deploy state but no loadable strategy.yaml — refusing to fetch"*) if the fork dir is ever partially written.

To be fair to the PR: this is **not** a wallet-adoption risk. `deploy.py:64` is explicit that "There is NO local deploy-state file any more" — the sidecar is a legacy marker only. One clause in the copy step covers it.

## 3. The lever list is right on substance, mis-ranked on key names

I counted every key across all **135 catalog recipes**:

| lever key | present |
|---|---|
| `strategy.slots` | **135 (100%)** |
| `strategy.default_leverage` | **135 (100%)** |
| `exit.dsl_preset.phase1.max_loss_pct` | **135 (100%)** |
| `risk.guard_rails` | **135 (100%)** |
| `scanner.interval_seconds` | every scanner |
| `strategy.margin_pct` | 72 (53%) |
| `inputs.maxLeverage` | 66 (48%) |
| `inputs.maxSlots` | 56 (41%) |
| `inputs.leverageTiers` | 34 (25%) |

The PR calls these "the universal ones every runtime.yaml has: max positions (`maxSlots` / `slots`), leverage (`maxLeverage` / `leverageTiers`)…" — but it **leads with the variant that is present less than half the time**. An agent reading in order looks for `maxSlots`, misses it in 59% of templates, and improvises at exactly the moment it is meant to be offering the user a confident choice.

**Change:** lead with the 100% keys and demote the others to what they are —

> max positions (`strategy.slots`; some templates also expose `maxSlots` in scanner inputs), leverage (`strategy.default_leverage`; scanner-side `maxLeverage` / `leverageTiers` where present)

Also consider adding **`strategy.margin_pct`** (53%) to the candidate levers — it is position size, which is a bigger felt lever than cadence, and it is absent from the list entirely.

## Verified correct — do not "fix" these

- `group: <sid>` and `name: <sid>-<instance>` are enforced as **hard errors** in `validate_strategy.py:181-185`. The fork procedure matches exactly.
- `forked_from` is safe: no `additionalProperties`/unknown-key rejection anywhere in `_pkg.py` or `validate_strategy.py`, so "the loader ignores keys it doesn't know" holds.
- `validate_strategy.py <dir>` does take a directory.
- **"pass the directory, not the template id" is right**, and `resolve_pkg_dir` backs it: an explicit path with `strategy.yaml` is honoured first and never second-guessed. Worth keeping that sentence bold — it is the line that stops a fork resolving back to the template.

Ponytail pass: `+69/-6` for a new required step, a fork procedure and an ownership line. No new files, no new structure. The Step 0.75 numbering is slightly odd (0.5 → 0.75 → 1) but renumbering would churn every cross-reference for no reader benefit — leave it. `net: 0 lines`.

betashop:
All four taken, amended onto the same commit and pushed (`579480e`):

1. **Fork destination** — `<strategies root>/<template>-<user-slug>/` where the root is what `deploy.py where` prints (`SENPI_STRATEGIES_DIR` > `$OPENCLAW_WORKSPACE_DIR/strategies` > `/data/workspace/strategies`); the example reads `/data/workspace/strategies/starling-purplefrog/` in both places (fork mechanics and the post-live ownership line), with "never a CWD-relative `strategies/`" and the 07-30 reason beside it.
2. **`.deploy-state.json`** (and `__pycache__`) left out of the copy, with your why in one clause.
3. **Levers re-ranked** — `strategy.slots`, `strategy.default_leverage`, `exit.dsl_preset.phase1.max_loss_pct`, `risk.guard_rails`, `interval_seconds` first as the 100% keys; `strategy.margin_pct` added as the biggest felt lever after the stop (about half the catalog); `maxSlots` / `maxLeverage` / `leverageTiers` demoted to "read where present, never look for first".
4. **`deploy.py where`** — prints `_pkg.strategies_root()`; dispatched right after `parse_args`, before the flag-reading code (it takes no flags, and that code assumes every verb declared `--json`). Two subprocess tests: env root verbatim, workspace fallback.

The "verified correct" list stayed as it was.

One consequence worth your eye: the walkthrough put the ops body at 366 lines against the 300 budget, so the depth (lever table, name rules, cost classes, consents, fork mechanics) moved to `references/walkthrough.md` and the resident rule is 14 lines. `BODY_BUDGET` goes 300 → 317, set at the post-edit count with no slack and the dated rationale in the test — the same pattern as the author 368 → 386 bump. 407 pass.

shnoodles:
**Re-review: all three findings addressed, and one of them better than I asked for.**

**1. Fork destination — fixed.** Now the *durable* root (`SENPI_STRATEGIES_DIR` → agent workspace → `/data/workspace/strategies`), with "never a CWD-relative `strategies/`" stated explicitly, and the ownership line now reads `/data/workspace/strategies/starling-purplefrog/`.

**2. `deploy.py where` shipped** — I suggested it as a nice-to-have; you built it with tests. Ran it:

```
$ python3 deploy.py where
⚠ no durable strategies root found (no SENPI_STRATEGIES_DIR, no workspace dir) —
  falling back to CWD-relative strategies/. Packages here may not survive skill updates.
strategies
```

Prints the root *and* surfaces the warning on a host that has no durable one, which is the right behaviour — the agent now has a way to ask instead of assume.

**3. `.deploy-state.json` left out of the copy** — noted in the fork step.

**4. Lever list — correctly re-ranked.** `references/walkthrough.md:23-30` now leads with the keys that are actually universal:

| lever | key | present |
|---|---|---|
| Max positions | `strategy.slots` | 100% |
| Leverage | `strategy.default_leverage` | 100% |
| Risk per trade | `exit.dsl_preset.phase1.max_loss_pct` | 100% |

…with `strategy.margin_pct` called out as "the biggest *felt* lever after the stop — offer it where it exists", and `maxSlots` / `maxLeverage` / `leverageTiers` demoted to the scanner-side minority they are. That is exactly right.

Tests: `test_deploy_where.py` + `test_skill_surface.py` **11 passed, 44 subtests**; whole ops suite **408 passed** (up from 396).

Extracting the walkthrough into `references/walkthrough.md` was a good call — it kept SKILL.md readable while the step grew.

**No further changes requested.** Ship it.

shnoodles:
Re-checked — all points addressed, nothing further from me. ✅

shnoodles (approved):
Verified on the branch: durable-root fix, `deploy.py where` + tests, lever list re-ranked. Ops suite 408 passing.

PR #579, opened 2026-09-11
discover: a template is a starting point for the user's own strategy — ownership rule, fork name, cost classes (2.23.0)

betashop:
## What

- **New golden rule — it's their strategy.** A template is a quick start to the user's own strategy, never "our strategy" deployed for them. Say "a starting point you can fork"; never "our strategies" / "Senpi's Starling" / "I'll deploy our template". Every template goes live as `<User>'s <Template>` (or a name of the user's own) after ops' walkthrough (Step 0.75, in the companion ops PR).
- **Cost classes as facts, author as a peer.** Four rungs, identical to ops, author and the workspace guardrail: template as-is ≈ the cheapest thing the agent does · a lever fork adds a little · a bespoke edit of a template adds more · scratch ≈ 2–3× a template. Build-custom is a peer route, never a dead-end or a downsell.
- **Card format**: the closing question becomes "start from X (I'll walk you through how it's set and the levers before we fund it; it deploys as your X — or a name of your own), add a hedge, or build your own?", and every card is called a starting point once.
- **Handoff**: ops owes the walkthrough before any budget question.
- The engine fallback line "here are our strategies" → "here's what you can start from".

Docs only. 2.20.0 → 2.21.0, README row synced. **Merge order: #578 (ops) first or together** — this PR promises the walkthrough (ops Step 0.75) that #578 introduces. Pairs with the author PR (3.6.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | [`senpi-improve-trades`](senpi-improve-trades/) | 1.9.0 | Retrospective review + health checks off the **telemetry event log**: exit quality, missed signals, leaks, crashes, "if I'd held" counterfactual |
 | [`senpi-account-status`](senpi-account-status/) | 1.2.0 | Points, loyalty tier, fees, referrals |
 | **Run a strategy** | | |
-| [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.22.0 | Conversational picker — rank the catalog against your worldview |
+| [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.23.0 | Conversational picker — rank the catalog against your worldview |
 | [`senpi-strategy-author`](senpi-strategy-author/) | 3.5.0 | Build/edit a DSL-protected strategy package, one decision at a time |
 | [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.11.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.3.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
--- a/senpi-strategy-discover/SKILL.md
+++ b/senpi-strategy-discover/SKILL.md
@@ -6,16 +6,18 @@ description: >-
   a strategy", "help me pick a strategy", "what's winning?", "set me up", "I have
   a view on the world (a war, the economy, one coin winning) — trade it", "run a
   hedge fund / all-weather / tail-risk book", or wants a strategy but has NOT
-  named a specific one. Surface the closest matching TEMPLATE first — the fastest
-  on-ramp, which the user can then fork/customize — passing their worldview as
-  `--theme` to rank the closest fits. You talk and RANK; a hidden engine
+  named a specific one. Surface the closest matching TEMPLATE first — the quick start
+  to the user's OWN strategy (every template deploys as `<User>'s <Template>` — or a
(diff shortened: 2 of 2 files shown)


shnoodles:
**Review: yes — small, and it changes the thing that actually needed changing.**

The golden rule lands in the right place (first in the list, where it governs every later instruction), and the wording fixes are precise: *"a starting point you can fork"* over "our strategies", and the engine's fallback string going from **"here are our strategies"** → **"here's what you can start from"**. That fallback is the one line a user sees when everything else has failed, so it was worth catching.

"Always offer build-custom **as a peer**; never dead-end, **never downsell**" with its cost class beside it is the correct framing for a skill whose job is to rank — discover is where the user forms their mental model of what the menu even is.

Cost numbers check out against the telemetry in the plan: "scratch ≈ 2–3× a template" brackets the measured ~2.6×.

Two things to change:

**1. The route menu here has three rungs; ops and author have four.** #580 defines template · **lever fork** · **bespoke edit** · scratch, and #578 states four matching cost classes. This PR (and senpi-agent#64) collapse to three — the bespoke edit is missing:

> template as-is ≈ the cheapest thing the agent does · a lever fork adds a little · scratch ≈ 2–3× a template

Discover is where the user **first** hears the menu, and then ops/author offer a rung they were never told about. Since "the routes are peers, each with its cost class" is the whole framing, the menu should read identically in all four places. Add the bespoke-edit rung.

**2. This PR promises behaviour that lives in #578.** The golden rule says every template goes live under their name *"after ops walks them through what it does, how it's set and which levers to shift (ops Step 0.75)"*. Step 0.75 does not exist until #578 merges. Shipped alone, discover promises a walkthrough ops will not perform, and the user hears the ownership pitch and then gets the old budget-first flow.

Not a blocker — just merge #578 first, or together. Worth a line in the PR body so it does not get picked off the queue on its own.

**Minor:** the README row and the `version:` both go 2.22.0 → 2.23.0, so they agree — but the Slack post announcing this called it "discover 2.22.0". Cosmetic, only noting it so the release note matches the artifact.

Ponytail pass: `+24/-11` and the deletions are real (the "our strategies" phrasings). Nothing to cut. `net: 0 lines`.

betashop:
Both taken, thanks.

1. **Four rungs now** — the golden rule states template as-is · lever fork · bespoke edit of a template · scratch, with the note that the menu reads identically in ops, author and the workspace guardrail (senpi-agent #64 updated to match).
2. **Merge order** in the body: #578 first or together; this PR promises the walkthrough that #578 introduces.

The Slack note said 2.22.0 because that was the number before the Phalanx and maker-only catalog bumps landed underneath; version and README row are both 2.23.0 here.

shnoodles:
**Re-review: fixed.**

The menu is now four rungs and matches ops, author and the workspace guardrail:

> template as-is ≈ the cheapest thing the agent does · a lever fork (values only) adds a little · a bespoke edit of a template (a new universe, a different signal — author's edit path) adds more · scratch ≈ 2–3× a template

Adding *"The same four rungs appear in ops, author and the workspace guardrail"* is a better fix than I suggested — it makes the next person editing one of the four copies aware the others exist, which is the thing that would otherwise drift.

The `(values only)` qualifier on the lever fork is a useful addition too: it is the line that separates rung 2 from rung 3.

The only open item is the merge-order one, which is not a code change: this still promises ops Step 0.75, so land #578 first or together.

**No further changes requested.**

shnoodles:
Re-checked — all points addressed, nothing further from me. ✅

shnoodles (approved):
Four cost-class rungs now consistent with ops/author/workspace.

PR #580, opened 2026-09-11
author: four peer routes with cost classes, budget never before the build, let-winners-run exits, the user names it (3.6.0)

betashop:
## What

- **Start here → four peers, template first, each with its cost class**: template as-is · fork a lever before deploy (ops Step 0.75) · bespoke edit of a template on the forked copy · scratch (≈ 2–3× a template). Even-handed: scratch is a peer that works, never a downsell; the user's choice is final.
- **Budget is never asked to start building.** It is asked once, at deploy (Handoff step 1). The funding heads-up stays a one-liner.
- **Pre-fill from the thesis.** A smart-money / cohort / divergence / copy thesis has a derived universe by definition — never ask which tickers. Ask only what the thesis leaves open.
- **Smart-money data source.** The proven cohort (`discovery_get_top_traders` ALL_TIME + `discovery_get_trader_state`, headcount and change), never the 4h gain board as the source of "who is smart" (circular). Name the template that already does it before writing a new one.
- **Exit default — let winners run.** Every stop and rung stated in price at the chosen leverage; a swing floor around 3% of price; lower leverage before tightening the stop. The validator's stop-distance warn is relayed.
- **The name is the user's.** Ask what they want to call it; `id` is its slug, `catalog.name` their words.
- **Editing**: the pre-deploy fork is the copy ops made under the user's name at the durable root `deploy.py where` prints (never a CWD-relative `strategies/`); the mechanics are named as ops' walkthrough reference, not repeated. Execution options are part of the edit, and the validator refuses what the executor cannot place.
- **Removing a protection is a consent question** (daily-loss limit, halt, cap, or a threshold the agent recommended): one line of consequence in the user's numbers, an explicit yes, never "done". Same for deploying below the design budget. From the M371275 review.
- Live edits are called an **update**, never a "redeploy".

**Validator, code (every rule sourced from the runtime, cited in the code):**
- **`[fees]` advisory**: fee spend per day from the recipe's own cap or cadence, slots, margin and leverage at a stated 0.05% per side; warns above 0.5% of the budget per day with the arithmetic in the message (a 10-min scanner with a 12/day cap at 20% margin and 3x reads 0.72%/day, ≈22%/month).
- **`leverageTiers` read by column** (review): rows are `[min_score, leverage, margin_pct]`, maps are `{tier: leverage}`. `max_leverage()` reads the leverage column and `slot_plan()` the margin column; the old flatten read condor's score and margin as leverage (80x). Re-run over the 135 catalog instances on this branch: `[fees]` trips 46 → 45 (wolverine cleared; condor now reads 10x at 80% margin → 0.80%/day at its real numbers).
- **`[exec]` errors** (review): an `OPEN_POSITION` action the executor cannot place. (1) `params.order_type` not in the runtime's own open set `MARKET | LIMIT | FEE_OPTIMIZED_LIMIT` (`constants/order-types.ts`) — the runtime does not refuse it, it warns and falls back to MARKET (`resolve-order-type.ts`), full taker fees while the recipe reads maker-first; (2) `LIMIT` — the MCP requires a `limitPrice` and the open action's `createPosition` call never carries one; (3) `FEE_OPTIMIZED_LIMIT` with `ensure_execution_as_taker: false` — a maker-only entry the executor cannot rest (fails every order, or a partial fill is reconciled as foreign and closed).
- **`[exec]` warn**: `fee_optimized_limit_options` under MARKET or LIMIT — the action forwards the options only for FEE_OPTIMIZED_LIMIT (`open-position.action.ts`), so the maker window is configured and never used.
- Catalog: 0 errors, 0 warns across 110 packages. Twelve validator tests (fees ×4, tiers ×2, exec ×6). 207 pass.

**Body budget**: the author SKILL body goes 386 → 425; `BODY_BUDGET` set at the post-edit count with no slack and the dated rationale in the test (same pattern as the 368 → 386 bump): every added line is a what-to-ask / what-to-claim rule that fires in the interview; the fork mechanics live in ops' `references/walkthrough.md`. Note: #578 changes the same `BODY_BUDGET` line for ops — whichever merges second gets a one-line rebase.

Otherwise docs; 3.5.0 → 3.6.0, README row synced. Rebased on main. Pairs with the ops (3.12.0) and discover (2.23.0) PRs; each stands alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Every analytical skill follows the **hidden-engine pattern**: a vendored, stdlib
 | [`senpi-account-status`](senpi-account-status/) | 1.2.0 | Points, loyalty tier, fees, referrals |
 | **Run a strategy** | | |
 | [`senpi-strategy-discover`](senpi-strategy-discover/) | 2.23.0 | Conversational picker — rank the catalog against your worldview |
-| [`senpi-strategy-author`](senpi-strategy-author/) | 3.5.0 | Build/edit a DSL-protected strategy package, one decision at a time |
+| [`senpi-strategy-author`](senpi-strategy-author/) | 3.6.0 | Build/edit a DSL-protected strategy package, one decision at a time |
 | [`senpi-strategy-ops`](senpi-strategy-ops/) | 3.12.0 | Deploy / monitor / close a named strategy (`deploy.py`, `close.py`) |
 | [`senpi-trade`](senpi-trade/) | 1.3.0 | Direct trade or mirror a specific trader — manual positions + copy trading, one decision at a time |
 | [`senpi-trading-runtime`](senpi-trading-runtime/) | 4.1.0 | The runtime contract reference: `scan(inputs, ctx)`, `runtime.yaml`, DSL |
--- a/senpi-strategy-author/SKILL.md
+++ b/senpi-strategy-author/SKILL.md
@@ -10,12 +10,15 @@ description: >-
   runtime.yaml authored here is the ONLY way to carry a DSL; raw MCP
   strategy_create* / create_position calls cannot, and must never stand up a
   named or protected strategy. Offers the closest TEMPLATE first (via
-  senpi-strategy-discover), then fork-or-scratch as the user chooses. NOT for
+  senpi-strategy-discover) as the quick start to the user's OWN strategy, with
+  fork-before-deploy, bespoke edit and scratch as peers, each with its cost class;
+  never asks for a budget to build (it is asked once, at deploy); exits default to
+  letting winners run, stated in price at leverage; the user names what you build. NOT for
   installing (senpi-strategy-ops) or picking one to run (senpi-strategy-discover).
 license: Apache-2.0
 metadata:
   author: Senpi
-  version: "3.5.0"
+  version: "3.6.0"
   platform: senpi
   exchange: hyperliquid
   requires:
@@ -53,31 +56,35 @@ much risk). Your job is to draw those out, one question at a time, and compile t
 > "succeeds," the position is gone, and the user eats the round-trip. If the user hasn't said which of
 > (A)/(B) they want, **ask before placing anything** — and never route (B) into a managed wallet to save a step.
 
-## Start here — offer the fast path before building from scratch
+## Start here — templates first, as the quick start to THEIR strategy; the other routes are peers
 
-Building from scratch is powerful, but it's the **slow** path (the full interview + compile + smoke-test).
-Most users — especially new ones — are best served by starting from a **proven template** and tweaking it
-if they want. So **before the interview, offer three ways to go** — as peers, with the fast one recommended,
-never as a gate:
+A template is not "our strategy" the user adopts — it is the fastest way to launch **their** strategy, and
+every one deploys under their name (`PurpleFrog's Starling`, or a name of their own) after ops walks them through what it does,
+how it is set and which levers to shift (ops Step 0.75). So **before the interview, offer the routes as
+peers — template first, each with its cost class as a fact — never a gate, never a downsell:**
 
-1. **Start from a matching template** — *the fastest way to get running.* If the user gave any thesis hint,
-   **hand it to `senpi-strategy-discover`** with their words — that skill surfaces the closest matching
-   template(s), which you name in the offer (*"**Cougar** — equity long/short — is close to what you
-   described"*). Discover owns the catalog and the match (and picks + deploys via ops); don't reach into
-   its internals or rebuild the catalog here.
-2. **Start from that template and make it your own** — deploy the template, then **edit it** (this skill's
-   edit path — see "Editing an existing strategy") to change the universe / thresholds / sizing / DSL. The
-   bridge for *"close, but I want changes."*
-3. **Design your own from scratch** — first-class, fully supported; you run the interview below.
+1. **Start from a matching template** — *the quick start.* If the user gave any thesis hint, **hand it to
+   `senpi-strategy-discover`** with their words — that skill surfaces the closest matching template(s),
+   which you name in the offer (*"**Cougar** — equity long/short — is close to what you described"*).
+   Discover owns the catalog and the match; ops deploys it under the user's name, as-is or with levers
+   moved. Don't reach into its internals or rebuild the catalog here. Cost: the cheapest thing the agent does.
+2. **Fork it before it goes live** — the walkthrough's levers (ops Step 0.75): a threshold, the slots, the
+   leverage, the stop, the daily cap. Values only, no new code. Cost: a little more than as-is.
+3. **Make a bespoke edit to a template** — a new universe, a different signal, a changed exit shape: this
+   skill's edit path ("Editing an existing strategy") on the forked copy, **before** deploy. Cost: more than
+   a lever fork, still well under scratch.
+4. **Design your own from scratch** — first-class, fully supported; you run the interview below. Cost:
+   roughly two to three times a template. Worth it when nothing close exists, or when the user wants it.
 
-**Tone — encourage without discouraging:** template-first is *"the fastest way to get running,"* **never**
-*"the right way"* — scratch is a **peer**, not a downsell. **The user's choice is final**: if they pick scratch
-(or already gave a specific thesis), go straight into the interview — **never re-pitch or nag**. Calibrate to
-the signal (vague ask → lean template-first; clear custom thesis → surface the closest match **once**, then
-build). No close fit → say so and go straight to scratch; never force a bad-fit template.
+**Tone — even-handed, ownership first:** template-first is *"the quick start,"* **never** *"the right way"*;
+scratch is a **peer**, not a downsell, and it works — say so. **The user's choice is final**: if they pick
+scratch (or already gave a specific thesis), go straight into the interview — **never re-pitch or nag**.
+Calibrate to the signal (vague ask → lean template-first; clear custom thesis → surface the closest match
+**once**, then build). No close fit → say so and go straight to scratch; never force a bad-fit template.
+**Never ask for a budget to start building** — see the funding heads-up below; the number is asked once,
+at deploy.
 
-Everything below is the **scratch / customize** path — the interview you run once the user chooses to build
-(or to tweak a template they just deployed).
+Everything below is the **scratch / bespoke** path — the interview, once the user chooses to build or to change more than a lever.
 
 ## ⛔ Never guess syntax — get it from the source (your memory is NOT authoritative)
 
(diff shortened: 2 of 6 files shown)


shnoodles:
**Review: the four-route framing achieves the goal, and the `[fees]` warning is a genuinely good addition — well calibrated, unlike the `[cap]` one beside it. But it is built on a helper that misreads `leverageTiers`, and I can show that from the scanner's own source.**

## `max_leverage()` reads score thresholds and margin percentages as leverage

`strategies/condor/main/scanners/scan.py:189` documents the shape:

```python
tiers = inputs.get("leverageTiers")   # optional [[min_score, lev, margin_pct], ...]
```

condor's recipe: `leverageTiers: [[15, 10, 80], [13, 10, 70], [11, 10, 50]]` — so `min_score` 15/13/11, **leverage 10/10/10**, `margin_pct` 80/70/50.

`max_leverage()` calls `_nums()`, which flattens the whole nested structure, and returns `max()` = **80** — a *margin percentage*. `strategy.default_leverage: 10` confirms the true figure. Same for grizzly: `[[14, 10], [12, 10]]` → returns 14, which is a score threshold; real leverage 10, `default_leverage: 7`.

I ran the new estimator over all **135 catalog instances**:

```
trips [fees] > 0.5%/day : 44  (32%)      ← good calibration
fee_drag returns None   :  7  ( 5%)

worst by estimated drag:
   4.20%/day  grizzly/main     6 entries/day — the daily cap
   4.00%/day  condor/main      1 entries/day — the daily cap   ← 1 entry/day at 4%?
   3.90%/day  pangolin/main   12 entries/day — the daily cap
```

condor at **4%/day on a single entry** was the tell: that implies notional 40× the budget for a 1-slot strategy. With leverage 10 instead of 80 the estimate is ~0.50%/day — right at the threshold rather than top of the list. An 8× overstatement on the recipes that use tiers (25% of the catalog).

**Both warnings over-fire from this.** `[stop]` (from #572) computes price distance as `max_loss_pct ÷ leverage`, so an inflated leverage makes the stop look *tighter* than it is. And `slot_plan()` has the mirror problem: it takes `marginPct: 50` from inputs while the tiers carry margin up to 80, so margin is under-read while leverage is over-read — from the same flattening.

**Fix is small** — parse the documented column instead of flattening:

```python
if k == "leverageTiers":
    found += [t[1] for t in (v or []) if isinstance(t, (list, tuple)) and len(t) >= 2 and _num(t[1]) is not None]
else:
    found += _nums(v)
```

Worth doing here rather than as a follow-up: this PR is what promotes the helper from an advisory stop-distance hint to a **money** estimate the agent reads aloud before the user funds.

## The route taxonomy diverges across the four PRs

This PR defines **four** routes (template · lever fork · bespoke edit · scratch). #578 (ops) matches with four cost classes. But **#579 (discover) and senpi-agent#64 both state three** — "template as-is · a lever fork adds a little · scratch ≈ 2–3×" — the bespoke edit is missing.

Discover is where the user *first* hears the menu, so they hear three options and are then offered four. Since "the routes are peers, each with its cost class" is the product framing this whole body of work exists to establish, the menu should be identical everywhere it appears. Add the bespoke-edit rung to #579 and #64, or drop it here — but pick one.

## Smaller

**The fork path repeats #578's CWD-relative form.** `strategies/<template>-<user>/` is the legacy fallback tier of `_pkg.strategies_root()`, which warns loudly and caused the 2026-07-30 wipe. Detail in my #578 comment; same one-word fix applies here since the two texts must agree.

**The maker-first note is a good catch** — "a pure maker-only entry (taker fallback off) passes validation and then fails every order at the executor" is exactly the accepted-but-unreachable config class. Consider making it a `validate_strategy.py` error rather than prose, since it is statically decidable from `order_type` + `ensure_execution_as_taker`; prose is the weakest place to put something a check can catch.

**It is `CONFLICTING`** — needs a rebase before anything else.

Ponytail pass: `+174/-31`. The four-route block replaces a shorter template-first pitch and earns the length — it is the product change. `fee_drag_pct_per_day` is ~20 lines with the assumed rate stated in the message rather than hidden, which is right. `_interval_seconds` correctly scopes to `external_scanner` only. No speculative structure. `net: 0 lines` — the change I want is a corrected line in `max_leverage`, not fewer lines.

betashop:
Taken, rebased on main (README row re-synced) and pushed (`a4464db`):

- **`leverageTiers` read by column.** A `tier_leverages()` helper: rows `[min_score, lev, margin_pct]` → index 1 (your snippet, with non-row items still flattened so nothing regresses), maps `{apex: 5, …}` → the values. `slot_plan()` reads the margin column (index 2) through `tier_margins()`, so condor's sizing is 80%, not the input's 50. Tests: condor `[[15,10,80],…]` → 10x / 80%; grizzly `[[14,10],[12,10]]` → 10; starling dict → 5.
- **Fleet re-run on this branch** (135 instances, catalog as of today): `[fees]` trips 46 → 45 with the fix, `wolverine/main` the one cleared. Condor now reads 10x at 80% margin → 0.80%/day, so it still trips, at its real numbers rather than 80x. (Your 44 was on the pre-rebase catalog — phalanx and the whalehunter change landed since.)
- **Maker-only entry is a `validate()` error** now — `[exec]`: an `OPEN_POSITION` action with `ensure_execution_as_taker: false` — not prose; the prose points at the check. Catalog count: 0 (the #584 sweep). Two tests.
- **Fork path** — `/data/workspace/strategies/<template>-<user>/`, the root `deploy.py where` prints (#578), never CWD-relative.
- **Four routes everywhere** — #579 and senpi-agent #64 now carry the bespoke-edit rung, so the menu is identical in discover, ops, author and the boot file.

203 pass. Now `MERGEABLE`.

shnoodles:
**Re-review: the `leverageTiers` bug is fixed, and I re-ran the numbers to confirm it.**

`tier_leverages()` now reads the documented leverage column instead of flattening, and handles both catalog shapes (row list and `{apex: 5, good: 4}` map). Measured before → after:

| | before | after | recipe says |
|---|---|---|---|
| condor `max_leverage` | 80× | **10×** | `default_leverage: 10` ✓ |
| condor `[fees]` | 4.00%/day | **0.80%/day** | 1 entry/day |
| grizzly `max_leverage` | 14× | **10×** | ✓ |
| grizzly `[fees]` | 4.20%/day | 3.00%/day | |

Trip rate holds at **31% of 135 instances** (was 32%), so the calibration survived the correction — and condor is no longer sitting at the top of the warn list on a single entry per day, which was the tell.

You also added `tier_margins()` for the index-2 column, which closes the mirror half I mentioned (margin under-read while leverage was over-read). condor's margin now resolves to 80 from the tiers rather than 50 from `inputs.marginPct`.

**Maker-first is now a validator error, not prose** — with the reasoning inline ("Decidable here, so it is an error here, not prose") and real evidence in the comment (68 of 79 signals over five days). I checked it does not break the catalog: **0 of 110 packages** trip it, and 0 have any `validate()` error at all. Using `is False` rather than a falsy test is the right call — an absent key is the safe default, not a violation.

**Fork path fixed** to `/data/workspace/strategies/<id>/`, with the reason stated ("so they hit the authored copy from any CWD"). Four route rungs now consistent across all four PRs. `CONFLICTING` → `MERGEABLE`. Author tests **203 passed**.

---

**One adjacent gap, for a follow-up rather than this PR.** Your new check closes `ensure_execution_as_taker: false`. Its sibling is still open, and I verified the asymmetry in the runtime schema:

```ts
// runtime-schema.ts:258-268  — the ACTION path
const actionEntrySchema = z.object({
  params: z.record(z.string(), z.unknown()).optional(),   // ← anything goes
}).passthrough();

// runtime-schema.ts:270-285  — the EXIT path
const exitBlockSchema = z.object({
  order_type: z.enum(['MARKET', 'FEE_OPTIMIZED_LIMIT']).optional(),   // ← constrained
  fee_optimized_limit_options: z.object({ ensure_execution_as_taker: z.boolean().optional() }),
});
```

The exit block enum-constrains `order_type`; the action block does not, and the action block is where the catalog actually puts it — all **135** OPEN_POSITION actions carry `params.order_type: FEE_OPTIMIZED_LIMIT`. `validate_strategy.py` does not check `order_type` anywhere.

So an authored fork setting `params.order_type: LIMIT` on an open passes the recipe schema, passes the skill validator, and fails at the executor — the same class your new check just closed, one key over. The catalog is uniform today so nothing is broken now; the exposure is new precisely *because* this body of work encourages users to fork and edit.

Mirroring the exit block's enum for action params is ~3 lines beside the check you just added. Worth filing.

**Otherwise: no further changes requested.**

shnoodles:
Re-checked — the `leverageTiers` fix is confirmed: condor `max_leverage` 80× → **10×** (matches `default_leverage: 10`), `[fees]` 4.00%/day → 0.80%/day, trip rate holds at 31%. Maker-first check flags 0 of 110 catalog packages. ✅

**One follow-up, not for this PR.** Your new check closes `ensure_execution_as_taker: false`; its sibling is still open:

```ts
// runtime-schema.ts:258-268 — the ACTION path
params: z.record(z.string(), z.unknown()).optional()   // anything goes

// runtime-schema.ts:270-285 — the EXIT path
order_type: z.enum(['MARKET', 'FEE_OPTIMIZED_LIMIT']).optional()   // constrained
```

The action path is where the catalog puts it (all 135 OPEN_POSITION actions carry `params.order_type`), and `validate_strategy.py` doesn't check it. So a fork setting `params.order_type: LIMIT` passes the schema, passes the validator, and fails at the executor — one key over from what you just fixed. Catalog is uniform today; the exposure is new because this work encourages forking. Mirroring the exit enum is ~3 lines beside your new check.

betashop:
Landed here after all — it is a few lines beside the check — but reading the runtime changed its shape, so flagging the differences from the suggestion:

- **The open set is not the exit enum.** `VALID_OPEN_ORDER_TYPES` is `MARKET | LIMIT | FEE_OPTIMIZED_LIMIT` (`src/constants/order-types.ts:7`); mirroring the exit enum would have refused `LIMIT`, which the open path accepts.
- **An unknown value does not fail at the executor.** `resolveOpenOrderType` logs a warning and falls back to MARKET (`src/utils/resolve-order-type.ts:9`) — every entry a taker order while the recipe reads maker-first. That silent fallback is the failure, so the validator refuses an unknown `params.order_type` and says what would have happened.
- **`LIMIT` is refused too**, for its own reason: the MCP requires `limitPrice` for LIMIT and the open action's `createPosition` call never carries one (`open-position.action.ts:550`), so the order cannot be placed as written.
- **`fee_optimized_limit_options` under MARKET/LIMIT is a warn, not an error**: the action forwards the options only for FEE_OPTIMIZED_LIMIT (`open-position.action.ts:300`), so they are inert rather than rejected. The maker-only error is now gated on FEE_OPTIMIZED_LIMIT for the same reason — under MARKET the flag does nothing.

Catalog: 0 errors, 0 warns across 110 packages. 207 tests.

Separately: **main's CI has been red since #581** — the min-budget golden had no phalanx entry, so every PR's checks failed at that step before pytest ran; the red on #578/#579 was that, not their content. Fixed on main (the golden entry, nothing else); the four branches are rebased so their checks run against a green base.

Two smaller things in this push: the author SKILL body went 386 → 425 lines, and `BODY_BUDGET` is set at that count with the dated rationale in the test (the fork mechanics point at ops' walkthrough reference instead of repeating it). #578 touches the same line for ops — whichever merges second gets a one-line rebase from me.

shnoodles (approved):
Verified: `leverageTiers` column fix (condor 80x→10x, fees 4.00%→0.80%/day), maker check flags 0/110 catalog, 203 tests passing. The `params.order_type` item is a follow-up, not a blocker.

PR #590, opened 2026-09-12
docs(scan-contract): a tool response is not guaranteed to be a dict

shnoodles:
## The trap

`call_tool` normalizes a result in the order `isError` → `structuredContent` → first text content parsed as JSON → **the raw text**. That last step is deliberate — it mirrors `producer.mjs` and is pinned by `test_unwrap_falls_back_to_raw_text_when_text_is_not_json` — so a *successful* call can hand back a plain `str` when the upstream answers with a message rather than a payload.

Nothing in the author guide or the scan contract said so. Worse, the worked example demonstrated the guard that does not work:

```python
if not md: continue
c = (md.get("data", md) or {}).get("candles", {})
```

A non-empty string is **truthy**. It passes `if not md`, then raises `AttributeError: 'str' object has no attribute 'get'` — unhandled, so it kills the entire tick rather than skipping one asset.

## Cost, last 14 days

Four users, 348 events, across four separate days. Three of the four hit the raw-text case; one hit the sibling — a bare JSON array coming back as a `list` — which the same `isinstance(dict)` guard covers. The most recent shipped `md.get("data", md)` — the template's own idiom — onto a freshly funded wallet within a minute of it being minted, and logged 0 candidates in 231 ticks. (Per-user detail is in the internal telemetry note, not here — this repo is public.)

## Why no gate caught it

It is **intermittent by construction**: it depends on what the upstream returned on that tick. the most recent user's run threw on 12 of 12 ticks in some hours, 4 of 12 in another, and then stopped entirely at ~13:00 when the upstream shape changed — the scanner has ticked clean since, still on the buggy code. So it passes `senpi validate`, passes every liveness test, and surfaces hours later.

`preflight._payload_ok()` does not catch it either: it only looks for a `success: false` flag, so `_payload_ok("some text")` returns `True`.

## The change

- both worked examples in `creating-a-strategy.md` now guard with `isinstance(..., dict)`
- a new **"The response envelope is not guaranteed to be a dict"** section in `scan-contract.md`, placed next to the existing `_f()` note — the same class of problem at a different depth
- names the sibling trap: a JSON payload parsing to a scalar (`"text"`, `123`) lands in the same place, so the test is the type, never the truthiness

## Not changed

`_unwrap` keeps the raw-text fallback. It is a pinned contract the Python and JS clients are required to share; changing one side alone would drift them. **The gap was the documentation, not the behaviour.**

Suite: **907 passed**, 8 skipped.

Review follow-ups (2026-09-12): runtime skill bump changed to a PATCH (4.1.1) per the runtime-skill convention; the list shape named alongside the raw-text one; per-user telemetry moved out of this public body.


betashop (approved):
Verified and approved.

- The claim holds: the runtime bridge's pinned test asserts `call_tool` returns the plain string when the text payload is not JSON, so a truthiness check passes and `.get` dies.
- The cost holds against telemetry: same four users, same counts, 348 events over the fortnight. One of the four hit the `list` shape rather than raw text; the same `isinstance(dict)` guard covers it, and the section now names it.
- The catalog is clean: all 136 catalog scanner files that call a tool already guard on the type — the exposure was user-authored scanners copying the guide's example, which is what this changes.

Two things adjusted in review: the runtime skill's bump is a PATCH (4.1.1) since its version follows the runtime contract, and the per-user telemetry moved out of this public body. Follow-up worth doing: a validator warning when a scanner calls `call_tool` without an `isinstance` guard anywhere in the file, so the next one is caught at validate time.

PR #593, opened 2026-09-12
ops 3.13.0: a DSL change has two targets — the file and the open positions — and the user picks

betashop:
## What

When a user asks to change the DSL (tiers, stop, cuts), the agent has been editing `runtime.yaml` and applying it with `update`, which reaches only positions opened afterwards — and telling the user the change is in while every open position runs the old ladder. The runtime snapshots the exit ladder at position creation; open positions change only through `ratchet_stop_edit`, one call behind the trade-approval gate per position (that gate is the user's yes — the right experience, kept).

The ops skill now carries the rule resident (five lines): **read both** — the file and `ratchet_stop_list`, which shows the ladder each open position actually carries — **show the drift** and each open position's new floor in its own numbers, and **ask (a) future positions, (b) the open ones, or (c) both**; never assume (a), never touch an open position without its own approval. `references/editing-a-live-strategy.md` gets the protocol: the mechanism, the drift table, the floor math (`trigger_pct` is ROE, `lock_hw_pct` is the share of the high-water kept; floor price from entry, leverage and the lock), a worked example from a live book at 3x, the (a)/(b)/(c) question verbatim, the note that an edit re-evaluates the ladder immediately against the current high-water, and verification.

3.12.0 → 3.13.0, README row synced. Body budget 317 → 322 with the dated rationale in the test; one surface test pins the rule and the reference.

## Why

From a live session on 2026-09-12: the agent audited the drift correctly (three ladders on one package), then on "make that config fix" reached for the open positions when only the file was meant — the approval gate caught it — and on other days has done the opposite, file only, leaving the open book on the old ladder. Both failure modes come from picking a target silently. The question is the fix.


shnoodles:
**Review: it achieves what it sets out to — the two-target rule is correct and the mechanism claims hold. One formula is inverted, and it is the one the agent reads out loud.**

## The sign is backwards in the floor-price conversion

`references/editing-a-live-strategy.md:128-129`:

> floor price = entry × (1 − floor ROE ÷ leverage ÷ 100) for a long, + for a short

The runtime is the opposite — `src/dsl/engine/floors.ts:18-23`:

```ts
/** Convert floor ROE % to price (LONG: entry*(1+roe/100/lev), SHORT: entry*(1-roe/100/lev)). */
const pct = floorRoe / 100 / state.leverage;
if (state.direction === "LONG") return roundTo(state.entryPrice * (1 + pct));
return roundTo(state.entryPrice * (1 - pct));
```

**LONG is `+`, SHORT is `−`.** Which is also the only shape that makes sense: a profit-lock floor on a long sits *above* entry — that is what locking a gain means. As written, the reference puts it below.

It matters because step 3 says to price each position "in its own numbers **before asking**", so this number is quoted to the user. Entry 100, 3x, floor 7.6% ROE: correct floor **102.53**, the reference's formula gives **97.47**. The agent tells someone their profit lock is underwater when it is 2.5% in the money — and then asks them to approve a change based on it.

One-character fix per direction; the rest of the math section is right (below).

## Everything else verified

- **`floors.ts:46`** — `const floorRoe = state.highWaterRoe * (tier.lockHwPct / 100)` — exactly the reference's "Floor ROE = high-water ROE × lock". ✓
- **`trigger_pct` is ROE** — confirmed by `config.test.ts:217`: *"accepts a tier with trigger_pct > 100 (ROE can exceed 100% under leverage)"*. ✓
- **Worked example arithmetic** — 30.5 × 0.25 = 7.6, 30.5 × 0.55 = 16.8. ✓ ("twice as much of the peak" is 2.2×, close enough to survive paraphrase.)
- **Retrace room** — `(high-water ROE − floor ROE) ÷ leverage` equals `entry × Δ / 100 / lev` as a share of entry. ✓
- **Immediate re-evaluation** — consistent with the engine deriving tier from current price (`tierIndexFromPrice`, `tick.ts:41`) rather than only on a crossing.
- **Tests** — `test_skill_surface.py` 10 passed / 45 subtests; whole ops suite **409 passed**. Budget 322 holds.

The core claim — file and open positions are two different edits, and the agent must ask rather than assume (a) — is right, and "the approval gate catches the second failure mode; only the question catches the first" is the sharpest line in the PR.

## Ponytail pass

```
tests/test_skill_surface.py:L52-56: shrink: 5-line dated rationale for a +5 budget bump; one
  paragraph per version means this comment grows without bound. One line — "ops 322: DSL
  two-target rule, 5 resident lines" — git blame holds the rest.

tests/test_skill_surface.py:L287,L290: shrink: 9 assertIn needles pin exact prose ("never
  assume", "one approval each"). Any reword of a doc breaks a green test. Three that cannot be
  reworded away — ratchet_stop_edit, ratchet_stop_list, senpi://guides/ratchet_stop — catch a
  deletion just as well and survive an edit.

SKILL.md:L340-341: shrink: the parenthetical explaining what ratchet_stop_list shows repeats
  reference step 1, linked two lines below. Name the tool, drop the gloss.

references/editing-a-live-strategy.md:L144-146: shrink: closing paragraph restates the opening
  and step 4. Keep only its last sentence — "The approval gate catches the second; only the
  question catches the first" — which is the one claim made nowhere else.

net: -7 lines possible.
```

No `delete:` findings. The 6-step protocol is a procedure, not padding; the floor math needs both formula and worked example because the formula alone does not tell an agent what to say; and the five resident SKILL.md lines are load-bearing since the rule fires before the reference is ever opened.

Fix the sign and this is good to go.

betashop:
Sign fixed — thank you, that was the one number the agent reads aloud. The reference now says entry × (1 + floor ROE ÷ leverage ÷ 100) for a long, − for a short, with your worked case beside it (entry 100, 3x, 7.6% ROE → 102.53).

All four shrinks taken: the budget rationale is one line at the measured count (321), the surface test pins only the three needles a reword can't remove, the resident rule names `ratchet_stop_list` without the gloss, and the reference closes on the one sentence made nowhere else. Ops suite green.

betashop:
Correction to the line above: the measured body count is **322**, which is what the test's one-line rationale says (`ops 322: DSL two-target rule, four resident lines`) — 321 was my typo, not the file's.

shnoodles (approved):

PR #594, opened 2026-09-12
ops 3.14.0: a template deploys under the user's name — the verb forks it (ignas-phalanx / "Ignas's Phalanx")

betashop:
**Why.** A live template deploy this evening: the walkthrough's "What it does" was a paragraph (skipped), the two levers were quoted by config key (`tiltThreshold`, `marginPctBase` — meaningless to a user), the name was asked as a question ("Your Phalanx, or something else?"), and when the user came back with "did you finish?" the agent ran the bare template id — so the strategy is named **phalanx** in the app. The fork that carries the user's name was a manual recipe in a reference, and it was skipped.

**What changes.**
- **`deploy.py create <template> --owner <username> --budget <usd>`** forks the template to `<owner>-<template>` under the durable root — `id`, `catalog.name` ("Ignas's Phalanx"), runtime `name`/`group`, the first line of `description` (the mandate portfolio reads back), `forked_from` — validates and proves the copy (`openclaw senpi validate --stage live`), and deploys **the fork**. The strategy in the user's list is `ignas-phalanx`; the agent says "Ignas's Phalanx". `--name "<their words>"` → `shield-wall`. Same owner twice = same fork reused.
- **`deploy.py fork <template> --owner …`** makes the copy and stops, for the levers-first path (edit the fork, then `create <dir>`).
- A bare **catalog template id** on `create` with neither flag is **refused (2)**, nothing created — the message names the command. An explicit directory (an author's own package) is never second-guessed; a fork deploys by directory with no owner.
- The two backend constraints are why there are two spellings: the deploy verb derives the strategy name from the package id (`[A-Za-z0-9_-]`) and the backend lowercases it and forbids spaces — so the possessive is spoken and stored in the fork, the slug is what the list shows.
- **Walkthrough rewrite (Step 0.75 + `references/walkthrough.md`)**: every block is bullets; what it does = three bullets; each lever named by what it controls, never by its key, with its current setting in the user's terms, what moving it does, and what it means at their budget; read the username first (`user_get_me`); the name is a stated fact, never a question; "go"/"deploy"/"yes" = as-is under their name; a status check is not a yes. Post-live ownership line leads with the name.
- New `_fork.py` (textual rewrites — comments and layout survive); 4 tests in `tests/test_fork.py` (names, rewrite, reuse, refusal of fork-of-fork and name collisions, flat packages, the CLI paths) + a surface class pinning the rule; ops body budget set at the measured count with a dated line.

Not in this PR: lever metadata inside the templates themselves (a `levers:` block each with plain-language copy) — worth doing next so every agent says the same words; and the discover skill's hand-off already says "it deploys as {user}'s {top}".


shnoodles (commented):
**It does what it says — I ran it rather than read it. One wording change before merge, then a ponytail list that is all optional.**

### Verified

Forked `phalanx` end-to-end and tested the refusal predicate directly:

| case | result |
|---|---|
| bare template id, no flags | `SystemExit(2)`, message names the command ✓ |
| explicit directory, no flags | passes through ✓ |
| a fork, by directory, no flags | passes through ✓ |
| `runtime` resume, no flags | no refusal ✓ |

Fork output is exactly as described — `ignas-phalanx` / `"Ignas's Phalanx"` / `name: ignas-phalanx-main` / `group: ignas-phalanx` / description prefixed / `forked_from` recorded. Reuse-by-lineage and fork-of-fork refusal both fire. 414 tests pass.

**The claim I most wanted to check held:** `phalanx/strategy.yaml`'s 8-line header comment survives the rewrite byte-for-byte. That is the textual-rewrite approach earning its keep.

I also checked the load-bearing assumption, since the whole guarantee rests on it: if any template lacked `catalog.tier` it would silently deploy bare. Through `_pkg.load()`, **110/110 carry it**. (A naive `grep` says three are missing — `dragonfly`, `spider`, `vulture` put a trailing comment on the `catalog:` line. They're fine.)

### Change before merge — the refusal message teaches the way around itself

`deploy.py`, the refusal in `_fork_for_deploy`, last sentence:

> Their own authored package deploys by DIRECTORY path instead.

Read by the agent that just got refused, that is an instruction for how to get past the refusal. And it works: a catalog template already fetched to the durable root **is** an explicit directory, so

```
create /data/workspace/strategies/phalanx --budget 100
```

deploys the bare template as `phalanx` — the exact outcome this PR exists to prevent.

I am not asking you to close that hole. It is forced: `catalog.tier` cannot separate a template from an authored package, because the author skill's own worked example sets `tier: advanced` (`creating-a-strategy.md:372`), so the directory escape hatch has to stay. The ask is only that the sentence point at the **fork** rather than back at the template — e.g. "deploy the fork by directory, or an authored package by its own path". Same escape hatch, but the sentence no longer reads as a workaround for the refusal it is attached to.

### Ponytail — all optional, ~18 lines of 190

```
_fork.py:39-46   shrink  4-separator loop for the title split
                         → re.split(r"\s[—–-]\s|:\s", t, 1)[0].strip()      -3
_fork.py:48-49   yagni   display_name(), one caller (names_for) — inline     -2
_fork.py:85-86   yagni   _set_catalog_name appends a catalog: block when
                         absent; 110/110 templates have one                  -2
_fork.py:103-104 yagni   _add_forked_from falls back to ^id: when ^version:
                         is missing; every manifest has version              -2
_fork.py:113-116 yagni   _prefix_description "description absent" branch;
                         135/135 catalog runtime.yaml have one               -4
_fork.py:127-131 yagni   the quoted / plain-inline description branch;
                         135/135 are block scalars, 0 observed cases         -5
```

The last one is the weakest: it has its own test, and `fork` accepts any package, so an authored one with a quoted description is conceivable. Fine to keep.

**What I deliberately did not flag:** the 190 lines of textual rewriting look like the obvious target and are the right call. `_pkg.py` already vendors a stdlib-only YAML fallback because "agent hosts may lack PyYAML / pip" — boxes genuinely have no pyyaml — so load/modify/dump is not available, and it would destroy the comments I watched survive. Nothing in `scripts/` duplicates it; `_set_top_key` is the first YAML writer here.

Docs are a rewrite, not bloat: walkthrough 93 → 112 lines, prose → bullets, which is the point.

betashop:
Thanks for running it. Pushed d10161f:

- **The refusal no longer teaches the way around itself.** Last sentence is now *"Deploy the fork by directory, or an authored package by its own path. Nothing was created, funded or installed."* Same escape hatch for authored packages (agreed it is forced by `tier: advanced` in the author's own example), but the sentence points at the fork.
- **Shrinks taken:** the title split is one `re.split`; `display_name` inlined into `names_for`; the "append a `catalog:` block" fallback is gone (a manifest without one is refused by name — 110/110 have one); the `^id:` fallback in `_add_forked_from` is gone (validate already requires `version`). Kept the two description branches: the quoted one has its own test and `fork` accepts any package.
- Suite green (parity, golden, pytest over every listed dir).

shnoodles (approved):

PR #595, opened 2026-09-13
senpi-trade 1.4.0: the protection protocol — read, quote, replace with consent, read back

betashop:
**Why.** On a live BTC short this weekend the agent answered "show dsl levels" without reading anything and narrated a DELETED ratchet as active; computed tier floors as `lockRoe × margin` and landed them below the price (long geometry on a short); promised a static SL and a ratchet "both active" when `ratchet_stop_add` replaces the static stop; argued a higher trigger was "tighter" on a short; and later re-added the ratchet inside a "health check" from a saved memory rule, minutes after the user had said to leave DSL alone. The user caught every one of these by hand.

**What changes (SKILL.md, resident):** a "protection protocol" section with six rules —
1. same-turn read before any claim; `DELETED` = no protection;
2. quote, never compute: `tierFloorPrice`; `floorRoe = highWaterRoe × lockRoe / 100`; kept $ = (entry − stop) × size on a short;
3. a short's stop sits above the price; a lower trigger is tighter;
4. leverage from the clearinghouse, never the strategy name; name the ROE base;
5. a change is read → say what it replaces → yes → act → read back (the read-back is the confirmation);
6. a saved "always" is a reminder to ask, a spoken "no DSL" stands across sessions, a status check is not a yes, a health check never changes protection.

The "two separate, uncoordinated stops — offer both" paragraph is replaced with the observed behaviour: one stop order per position, the engine owns it once a ratchet is active, `ratchet_stop_add` replaces a static SL. Four guardrail rows and three red flags added; Branch A step 4 and the mirror DSL paragraph updated to match.

**New:** `references/protection-cases.md` — the four regression cases (DELETED ratchet; short-side math with the exact numbers; "both"; saved rule vs spoken "no DSL") with must / must-not answers, written to double as eval cases. `tests/test_trade_skill_surface.py` pins the rules, the reference and the README version; `senpi-trade/tests` added to CI.

Related, not in this PR: the platform guide `senpi://guides/ratchet_stop` defines `lockRoe` as an absolute ROE and says manual SL/TP "remain" when a ratchet is added — both wrong versus the engine; corrected text is drafted separately. `senpi-portfolio` gets a matching "a health check never changes protection" rule in its own PR.


shnoodles (commented):
**The protocol is right and the arithmetic holds — I checked every number rather than reading it. One fix in `protection-cases.md`, and one cross-PR caveat worth settling before this merges.**

### Verified

The worked example in `SKILL.md` is correct on all four claims: both stops sit above the mark (short geometry ✓), a rising price reaches $77,500 before $78,370 ✓, that makes it the tighter one ✓, and it keeps more — $3,390 × size vs $2,520 × size ✓. So "lower trigger is tighter **and** keeps more" holds as stated.

Case 2 is better than internally plausible — it is self-proving:

```
high-water ROE          = (80,890 − 77,290.5)/80,890 × 20 = 88.997%   (file: 88.997) ✓
ROE at tierFloorPrice   = (80,890 − 78,370.35)/80,890 × 20 = 62.298%  (file: 62.3)   ✓
share formula           = 88.997 × 0.70                    = 62.298%                 ✓
```

`tierFloorPrice` and the quoted ROE agree to three decimals **under the share-of-high-water formula and only under it**. Kept dollars check out too ($2,095 / $1,557). That is the strongest evidence in any of these four PRs that the `lockRoe` re-definition is correct.

4 tests pass, and `senpi-trade/tests` is correctly added to the CI list.

### Fix before merge — the two "must not say" figures don't reconcile

`protection-cases.md` Case 2 says not to say *"locks $3,626 / $4,528 (those are `lockRoe × margin`, the wrong formula)"*. Under the case's own normalised numbers that formula gives different answers:

```
notional = 80,890 × 0.618 = 49,990.02   margin @20x = 2,499.50
lockRoe × margin:  70% → $1,750     85% → $2,125
```

$3,626 / $4,528 appear to be carried over from the raw incident (a different size/margin) into an otherwise normalised case. The file opens by promising *"Numbers are a single fixed short so the arithmetic can be checked by hand"* — these two can't be. Either recompute them to $1,750 / $2,125, or mark them as the incident's actual figures rather than what the wrong formula yields here. Matters because these are eval assertions: anyone regenerating expected outputs from the stated position will not reproduce them.

### Cross-PR caveat

The `lockRoe` = share-of-high-water redefinition is load-bearing for all four PRs — if it is wrong they are wrong together. I tried to confirm it against the engine and **could not**: `lockRoe` appears nowhere in `senpi-hyperliquid-mcp` (client only — it passes the value through and reads `tierFloorPrice` back) nor in `strategy-manager`. The evidence I do have is strong but indirect: the three-decimal agreement above, plus the fact that the old guide's example ladder is incoherent under the new semantics. Worth one confirmation from whoever owns the engine before merge, since four repos change on it.

### Ponytail

`Lean already. Ship.` 57 lines of resident rules, 97 of regression cases that double as evals, 49 of surface test. The cases file is the largest piece and earns it — it is the only place the numbers are pinned. No speculative branches, no abstraction with one caller, nothing to cut.

One observation, not a cut: "a status question is never a yes" and "memory is a reminder, not permission" now exist in four documents (this, `senpi-portfolio/SKILL.md`, `workspace/AGENTS.md`, the MCP guide). That is deliberate — different surfaces load in different contexts and the MCP guide reaches agents that never load a Senpi skill — but it is now four copies that can drift independently, and only this one has a test pinning it.

betashop:
Fixed in 5463ac6. The must-not pair now reads $1,750 / $2,125: `lockRoe × margin` at 70% and 85% of the $2,499.50 margin. Every figure in the file now reproduces from the stated position. I recomputed them rather than labelling them as outside figures, because the file promises numbers you can check by hand.

On the `lockRoe` caveat, there's one more piece of source evidence. The trading runtime passes its YAML `lock_hw_pct` straight through as the backend's `lockRoe`, then hands Phase 2 to the backend engine and adopts the `tierFloorPrice` it returns. The runtime's own local engine computes `floorRoe = highWaterRoe × lock_hw_pct / 100`. Every template ladder in this repo rides on that: if `lockRoe` were absolute, a `{trigger_pct: 12, lock_hw_pct: 40}` tier would put the floor on the wrong side of the price the moment it fired. The engine source still isn't in any repo I can search, so a one-line confirmation from its owner is still worth getting.

On the four copies, agreed. I'm keeping them for the reason you gave, and the senpi-trade test is the reference copy if they drift.

betashop:
The `lockRoe` caveat is settled by the engine's own output. When the runtime hands a position over, it logs the backend's response, including the high-water and `tierFloorPrice` the backend computed. It also logs every floor the backend returns after that. Over Sep 10–13:

- **Handoff snapshots with a non-zero lock:** all 90 match share-of-high-water, within the backend's rounding. The absolute reading matches none of them.
- **Live floor updates after handoff:** all 4,946, across 9 positions and tiers 0–3, match. The absolute reading matches none. One position's floor moved 11 times inside a single tier as its high-water rose, which an absolute lock can't do.

So the four PRs all rest on the correct meaning.

shnoodles (approved):
Approving — fix verified.

`protection-cases.md` now reads $1,750 / $2,125 and names the $2,499.50 margin, so the wrong formula is reproducible from the case's own numbers and the file's 'checked by hand' promise holds.

Re-confirmed the rest still stands: high-water 88.997% and tierFloorPrice → 62.298% agree to three decimals under the share formula, the worked short example is correct on all four claims, 4 tests pass, `senpi-trade/tests` is in CI.

PR #65, opened 2026-03-11
Fox 1.4

betashop:

PR #596, opened 2026-09-13
senpi-portfolio 1.21.0: a health check reads — it never changes protection

betashop:
**Why.** A "run a full health check" turn called `ratchet_stop_add` on a user's position — unasked, at 40x — because a saved memory rule said protection was mandatory, replacing the user's manual stop minutes after they had said to leave DSL alone. A check must never be the thing that changes the book.

**What changes.** One resident rule beside the existing "do NOT hand-roll DSL verdicts" block: this skill reads; it never calls `ratchet_stop_add` / `ratchet_stop_edit` / `ratchet_stop_delete` / `edit_position`; it reports what is there and what is missing, and offers the change as its own question, which then runs senpi-trade's protection protocol (read → say what it replaces → yes → act → read back). "No DSL" stands across sessions; a status question is never a yes.

Pairs with the senpi-trade 1.4.0 PR (the protocol itself).


shnoodles (commented):
**Correct, minimal, and aimed at the right layer. Nothing to change.**

I checked whether this rule is guarding something real or something already impossible: `senpi-portfolio/scripts/` never calls `ratchet_stop_add` / `_edit` / `_delete` / `edit_position`, and the skill doesn't declare them. So the engine was never the problem — the **agent** reached past its own read-only engine during a portfolio turn. A resident rule in `SKILL.md` is exactly the layer that catches that; a change to the engine would have caught nothing.

Placing it inside the existing "do NOT hand-roll DSL verdicts" block is the right spot: that block is already where an agent goes when it is tempted to improve on the engine's protection read, and "don't invent a verdict" and "don't act on the verdict" are the same instinct one step apart.

Version 1.20.0 → 1.21.0 with the README row synced ✓.

### Ponytail

`Lean already. Ship.` 11 lines, one rule, no new file, no test to add — `test_skill_surface.py` already pins the README/version agreement this touches.

One cross-PR note, not a change request: this rule and #595's rule 6 and #65's AGENTS.md block now say the same thing in three places (four with the MCP guide). That is deliberate and I would keep it — a portfolio turn may never load `senpi-trade/SKILL.md` — but only #595's copy has a test pinning its wording, so the other three can drift silently.

PR #597, opened 2026-09-13
discover 2.26.0: a discretionary mandate routes here

betashop:
**Why.** A new depositor said "I don't [have a strategy]. Deploy any trades you think will be profitable. Feel free to be aggressive." Nothing in the skill triggers claimed that sentence, so the agent read the market and opened three hand-picked longs with fixed stops through `strategy_create_custom_strategy` — no runtime, no DSL — instead of the smart-money template it had just described.

**What changes.** The description's triggers now include the discretionary mandate ("deploy whatever you think is profitable", "trade for me", "be aggressive", "put it to work", "you decide") and say it is a strategy request even when it names no strategy and no coin, never answered with hand-picked raw positions. Version 2.26.0 (rebumped after #600 took 2.25.0), README row.

Pairs with: the senpi-trade fork line on #595 (no coin named = a mandate → discover), the senpi-agent guardrail (#66), and the MCP description fix (which called raw positions "the default, primary path").


shnoodles (approved):
Approving. The trigger list closes the gap the incident found — a mandate names no strategy and no coin, which is exactly why nothing fired — and 'never answered with hand-picked raw positions' makes the routing explicit rather than implied. Pairs correctly with mcp#208 and agent#66.

Needs a rebump to 2.26.0 after #600 lands (same `version:` line, as your body notes).

PR #600, opened 2026-09-13
discover 2.25.0: Aegis in the catalog, and offered as Phalanx's pair

betashop:
**Why.** Aegis landed as a template in #599 (a regime-adaptive dynamic hedge built to run beside Phalanx). The catalog copy inside discover already carries it, but installed pickers only pull a new catalog when discover's version moves — this is that bump, plus the pairing.

**What changes**
- `SKILL.md` version **2.25.0** (not 2.24.0: #597 holds 2.24.0 — the version line will conflict so the two can never ship under one number; whichever lands second rebumps).
- "Stack, don't isolate": names the pair that was built for it — Phalanx + **Aegis** (reads the tape where Phalanx reads the crowd), sized ~70/30.
- `test_new_strategy_discoverability.py`: two rows pinning Aegis top-3 for "dynamic hedge regime adaptive risk off" and "hedge for phalanx" (it ranks #1 for both, and for "risk off hedge gold yen" / "hedge that goes to cash in chop").
- README row.

Parity + golden + full CI suites green locally.


shnoodles (approved):
Approving. Aegis's two pinned rows rank it #1 for both probes, the pairing line lands where the stacking guidance already lives, and CI is green.

One sequencing note since this goes first: #597 bumps the same `version:` line from 2.23.0, so it will conflict the moment this merges and needs a rebump to 2.26.0. Your PR body already calls this out — flagging only so it isn't a surprise in the merge queue.

PR #601, opened 2026-09-13
ops 3.15.0: a template forks under the user's Senpi username, read by ops itself

betashop:
**A template now forks under the user's Senpi username, and ops reads that name itself.**

The goal of Make It Yours is that a strategy launched from a template still reads as the user's own: "PurpleFrog's Aegis", `purplefrog-aegis` in their list. The fork name came from `--owner`, which the agent had to fill in, and the docs sent it to `user_get_me` for a username that tool never returned. With only the account ID to go on, an agent just named a live strategy after that ID.

**What changes**
- `deploy.py` reads `userName` from `user_get_me` itself, so `create <template>` and `fork <template>` need no owner.
- A user ID (`M123…`, or the key-scoped `M123…-hash`) is refused as a name, in `--name` and `--owner` alike.
- When no username can be read (no token, or an MCP that doesn't return it yet), a bare template is refused, and the agent asks the user what to call it and passes `--name`. It never falls back to the ID.
- `--owner` stays as an override, so sessions still on the old instructions keep working.
- SKILL.md and `references/walkthrough.md` drop `--owner <username>` from the commands and name the field. Version 3.15.0, README row.

**Tests.** A user ID is refused in `names_for` and through the `fork` CLI. `_username` reads `data.user.userName`, and returns `None` rather than the ID when it's missing or there's no token. A bare `create` forks as `purplefrog-phalanx` / "PurpleFrog's Phalanx". The offline CLI refusal now asks for `--name`. The surface test pins "A user ID is never a name" in place of `--owner`. Exact CI steps: parity ✓, golden ✓, pytest 916 passed.

**Depends on** senpi-hyperliquid-mcp#209, which makes `user_get_me` return `userName`. Either merge order is safe: until #209 is live, ops asks the user for a name instead of stating it.

**Ponytail:** `--owner` could go entirely, since `--name` covers every override. I kept it only for backward compatibility, and the ID guard makes it harmless.


shnoodles (commented):
**The design is right — reading the name instead of asking the agent to supply it is the correct fix, and `_username()` returning `None` rather than the id is the good half. But the headline guarantee has a hole, and it is on the path your own refusal message sends the agent down.**

### Fix before merge — a user ID still gets through `--name`

The body says *"A user ID (`M123…`) is refused as a name **wherever it arrives**."* It is refused in the `owner` branch only, and `name` is tested first and wins:

```
--name  "M401059"        -> ('m401059', 'M401059')                        NOT refused
--owner "M401059-7a611"  -> ('m401009-7a611-phalanx', "M401059-7a611's …") NOT refused
--owner "M401059"        -> REFUSED ✓
```

Two separate misses:

1. **`--name` has no ID check at all.** This is the reachable one, because the new refusal text routes the agent straight to it in exactly the situation where the ID is the only identifier it holds:

   > *"Ask what to call it, then re-run with `create … --name "<their words>" --budget <usd>`. A user ID is never a name."*

   An agent that half-follows that — takes the `--name` instruction, fills it with the ID it has — lands on `m401059-phalanx`, which is the bug this PR exists to close. The sentence warns; nothing enforces.

2. **`[Mm]\d+` misses the key-scoped form.** `M401059-7a611` is a real shape (the `M…-hash` key-scoped MID) and it sails through `--owner` into `"M401059-7a611's Phalanx"`.

Both close by moving the check above the branch so it covers whatever arrives, and widening it a little:

```python
ident = (name or owner or "").strip()
if re.fullmatch(r"[Mm]\d+(-[0-9a-z]+)?", ident):
    raise ForkError(...)
```

The tests presently cover only the owner path — `("M123456", "m123456", " M42 ")` through `names_for`, and `fork --owner M123456` through the CLI. Worth adding the `--name` case, since that is the one the refusal message actively steers toward.

### Also before merge

`README.md` conflicts with `main` — the version table only, no code conflict. Main has moved (#595, #597, #602). Trivial rebase.

### Verified

`_username()` is right: no token → `None`, exception → `None`, missing/blank field → `None`, never the id, and it is only called after the template/explicit-dir guard so a non-template `create` does not pay for an MCP round trip. The `fork` verb reads it too via `a.owner or (None if a.name else _username())`. 416 tests pass in the ops suite.

The "either merge order is safe" claim holds — without #209 the field is absent, `_username()` returns `None`, and the path degrades to asking for a name rather than to the ID.

### Ponytail

On your own note — keeping `--owner` is the right call, and not only for backward compatibility. It is the only way to name a fork for someone whose Senpi username is not what they want on the strategy, and `--name` is a different thing (it replaces the whole id rather than the owner half). Two knobs, two jobs.

`net: 0 lines cuttable.` The ID guard is the one place I would *add* a line rather than remove one.

betashop:
Both fixed. Thanks for catching the `--name` path; it's exactly where my own refusal message sent the agent.

- **One check, above the branch** (fbda17d): `names_for` checks `name` and `owner` alike before either is used. The pattern is `m\d+(-[0-9a-z]+)?`, case-insensitive, so the key-scoped `M…-hash` form is caught too.
- **Tests:** plain, lowercase, padded and hash-suffixed IDs are each refused through both flags. `create … --name M123456` is refused through the CLI, the path the refusal points at. "Mars Base" is still accepted as a name.
- **README:** merged `main`. The table now reads ops 3.15.0 and senpi-trade 1.4.0.

Exact CI steps: parity ✓, golden ✓, pytest 921 passed.

PR #605, opened 2026-09-14
discover 2.29.0: "which templates are proven?" gets the honest answer, never a mirror ranking

betashop:
**"Which of your templates are proven?" now gets the honest answer instead of a mirror-strategy ranking.**

From today's #user-pulse: a user asked whether a template had been backtested, then for "tested proven strategies people are making money from", and was answered twice with `discovery_get_top_strategies`, which is mirror-only. They read that as "Senpi is built only to promote mirror trading" and left. The agent had no better tool: there is no per-template performance surface, so it reached for the one ranked-PnL tool it has.

**What changes**
- *Special paths* gains the answer: a template has no backtest and no per-template performance board; "validated" means it runs (`senpi validate`) and forward-tests at the $10 floor; never answer a templates question with a mirror ranking, and offer copy-trading only as a labelled alternative.
- One few-shot line routes "proven / backtested" asks to that answer with no engine call.
- A surface test pins the line in the skill body. Version 2.29.0, README row.

**Not in this PR:** the real fix is a per-template track record the agent can quote. The backend already stamps every wallet with its template id, so a discovery endpoint grouping user PnL by template would give the agent a true answer. Logged as a product item in the Fix Loop ledger.

Exact CI steps: parity ✓, golden ✓, pytest 927 passed. Ponytail: five lines of skill text, one test, nothing else.


betashop:
Rebased onto main after #609 (discover 2.28.0 landed there); this PR now bumps discover to 2.29.0. CI re-run locally: parity OK, golden OK, suites green.

shnoodles:
**Review — ships what it claims. One required fix, two ponytail trims.**

**Intent ✓.** Notion row (the reporting user's agent answered a "which templates are proven?" question with `discovery_get_top_strategies`, which is mirror-only) → the *Special paths* answer plus the few-shot route are exactly the agent-behaviour half of that. Checked the leak in the other direction too: nothing else in the repo prescribes `discovery_get_top_strategies` for a templates question — only `senpi-trader-research/scripts/research.py` and the `cuckoo` scanner, both legitimately mirror. The guard is in the one place it leaks from. Out-of-scope note in the body (no per-template track record exists) is the right call.

---

### Required

**1. Version mismatch — title/commit/body say `2.28.0`, the diff ships `2.29.0`.**
`SKILL.md` and `README.md` both go `2.28.0 → 2.29.0`. The *bump* is correct (2.28.0 is already on main); the title, commit subject `12453b4d` and the body line "Version 2.28.0, README row" are stale. Fix those three, not the bump. The PR title is the audit trail this loop writes back to Notion — a cycle from now, "what shipped in 2.28.0?" has no answer.

### Ponytail (diff-scoped)

**2. `tests/test_skill_surface_discover.py:13` reinvents frontmatter-stripping.** `body.split("---", 2)[2]` is a new variant of something the repo already has, in `senpi-strategy-ops/tests/test_skill_surface.py:42`:

```python
m = re.match(r"^---\n.*?\n---\n", text, re.S)
return text[m.end():] if m else text
```

Same line count, correct on the edge case — that SKILL.md contains 8 `---`, and the naive split only lands right because none of them sit inside the frontmatter yet. Copy the regex. (Cross-skill *import* is correctly out — skills install standalone — but copying the proven three lines beats inventing a flimsier two.)

**3. Four prose needles pin one paragraph.** `"has no backtest"` and `"mirror-only"` carry the rule; `"per-template performance board"` and `"never as the answer"` just make a reword go red for no behavioural reason. Drop the last two.

---

### Verified locally
Branch @ `12453b4d` (= `origin/main` + 1 commit), python 3.14:

- `test_min_budget_vendor_parity.py` ✓ · `test_min_budget_golden.py` ✓ (110 templates)
- Full CI suite: **922 passed, 9 skipped, 63 subtests passed**, 1 failed.
- The one failure — `senpi-portfolio/tests/test_portfolio.py::test_all_step_is_byte_identical_to_run` — **is not this PR.** It reproduces identically on `origin/main`, and it is a thread race, not a regression: `hydrate` runs under `ThreadPoolExecutor(max_workers=6)` (`portfolio.py:983`) while each worker appends to the shared `meta["warnings"]`, so warning *order* is whatever finishes first, and the test asserts byte-identical JSON. Red on 5 of 8 `PYTHONHASHSEED` values on main. Reported separately — no action here.

betashop:
Review addressed (`75155ab` on top of the amended `8984150`): (1) title, commit subject and body now say 2.29.0 — the bump was already right; (2) the test strips frontmatter with the ops regex (`re.match(r"^---\n.*?\n---\n", …)`) instead of `split("---", 2)`; (3) needles down to `has no backtest` + `mirror-only`. #610 rebased onto the new head. CI: parity ✓, golden ✓, 927 passed / 9 skipped (the portfolio flake is fixed in #607, which is why it is green here without a deselect on this run).

PR #606, opened 2026-09-14
author 3.8.0: a close is a scanner plus a CLOSE_POSITION action, never a direction

betashop:
**A close is a dedicated scanner plus a `CLOSE_POSITION` action, never `direction: CLOSE`. The lint now says so before a wallet exists.**

From today's #user-pulse: an authored package emitted "force-flat" signals with `direction: CLOSE`, and no runtime action consumed them. Nothing surfaced it. The runtime routes a signal only to the actions whose `scanners:` list its scanner and drops a signal with no consumer without a log line, and the open action refuses any direction but `LONG`/`SHORT` (`invalid_direction`). The user discovered it themselves and removed the feature by hand.

**What changes**
- `validate_strategy.py` gains two decidable checks: an external scanner no action lists (in `scanners:` or `context[].scanner`) is an error with the wiring to add; a `direction` literal in scanner source that names a close (`CLOSE`, `FLAT`, `EXIT`) is an error naming `invalid_direction` and the shipped pattern; words like `NEUTRAL` in analysis dicts are left alone, since eleven catalog scanners use them. The first check applies to recipes that declare actions at all; a recipe with none is a different defect.
- `creating-a-strategy.md` §6 documents the pattern from Barracuda's `close_all`: a dedicated close scanner emitting each open position's asset and side, consumed by its own `CLOSE_POSITION` action. §10's checklist carries the one-line rule.
- Tests: Barracuda-shaped wiring passes; an unlisted scanner is refused; a `CLOSE` literal is refused; the helpers are pure. All 110 catalog templates pass the new checks unchanged.
- Version 3.8.0, README row.

**Verified:** exact CI steps (parity ✓, golden ✓, pytest passed). **Ponytail:** two helpers, two error branches, one doc paragraph, one test file; the checks reuse the parsed runtime doc the validator already holds.

**Companion:** a runtime issue asks for a log line when a signal has no consumer, so the same defect is visible on a live box too.


shnoodles:
**Review — the right fix in the right place. Two required changes; one is a hole in the new check itself.**

**Intent ✓.** Notion row (force-flat signals with `direction: CLOSE` that no action consumed, discovered by the user, removed by hand) → doc half in `creating-a-strategy.md` §6 + lint half in `validate_strategy.py`. Both halves land.

The cited pattern is real, not recalled: `strategies/barracuda/main/scanners/close_all.py:69` does emit `"direction": "LONG" if szi > 0 else "SHORT"`, consumed by `close_all_action` (`CLOSE_POSITION`, `scanners: [close_all_scanner]`) at `runtime.yaml:146`. The body's "all 110 catalog templates pass unchanged" also holds — I ran `validate_strategy.py` over every package on this branch: **0 hits** on either new check.

---

### Required

**1. The `actions` guard exempts the worst case — a zero-action recipe validates completely clean.**

`validate_strategy.py:339`:
```python
if isinstance(rt_doc, dict) and rt_doc.get("actions"):   # a recipe with no actions at all is another defect
```
That comment points at a defect **nothing in this validator catches.** Built a package with `actions: []` and again with the key omitted, everything else well-formed:

```
NO actions: key    -> errors: NONE — validates CLEAN
actions: []        -> errors: NONE — validates CLEAN
```

So the one package where *every* scanner's signals are dropped with no log line — the exact failure this PR exists to close — is the one the check skips. The orphan case is caught correctly when actions exist (verified), which makes the gap sharper, not softer.

Drop the `and rt_doc.get("actions")` condition. The error text already reads correctly for a zero-action recipe — *"List it under an OPEN_POSITION action (entries) or a CLOSE_POSITION action (closes), or remove it"* — and it is one less branch. Checked it is safe: of **138 catalog runtimes, 0 have zero actions**, so dropping the guard adds no new error anywhere in the catalog.

**2. Rebase — this PR conflicts with `main`.**

`gh` reports `CONFLICTING/DIRTY`; locally `git merge-tree origin/main pr-606` → `CONFLICT (content): Merge conflict in README.md`. The branch is 1 commit behind; `main` moved the discover row to `2.28.0` and the author row you edit is the adjacent line, so the two land inside one context window. (605/607/611 merge clean against main and against each other in any order — 606 is the only one.)

### Optional

**3. `direction_literal_offenders` only sees the dict-literal form.** The regex requires a `:`, so an assignment slips through:

```python
s = {'asset': 'BTC', 'data': {...}}
s['direction'] = 'CLOSE'        # -> direction errors: NONE
```
The literal form is what telemetry actually showed, so this is a nice-to-have, not a blocker — worth one alternation in `_DIRECTION_LITERAL` if it is cheap.

---

### Ponytail (diff-scoped) — nothing to cut

Checked the two usual suspects and both are house convention, not bloat:

- `test_validate_close_wiring.py`'s `_package(tmp_path, …)` builder duplicates a shape in five sibling files (`test_validate_fees.py`, `_tiers_and_exec`, `_package_id`, `_warnings`, `_wallet_binding`). Per-file fixtures are this directory's pattern and skills install standalone — leave it.
- The `isinstance(rt_doc, dict)` defensiveness in `unconsumed_scanners` matches seven pre-existing uses in the same file (lines 458–552). Consistent.

Two helpers, two error branches, one doc block, one test file is the right size for this. No trims.

### Verified locally
Branch @ `8a02a3fc`, python 3.14: parity ✓ · golden ✓ (110 templates) · full suite **925 passed, 9 skipped, 63 subtests** (+4 from this PR). The single `senpi-portfolio::test_all_step_is_byte_identical_to_run` failure reproduces on `origin/main` and is unrelated — thread race on `meta["warnings"]` under `ThreadPoolExecutor`, reported separately.

betashop:
Review addressed: (1) the `and rt_doc.get("actions")` guard is gone — `actions: []` and a missing `actions:` key now refuse with the same "feeds no action" line per scanner (test added for both variants); it surfaced that the warnings-suite fixture was itself a zero-action recipe with an unnamed scanner, so that fixture now wires its scanner to an entry action, and an unnamed scanner reports as `<unnamed>` instead of `'None'`; (2) rebased onto main — the README conflict was the adjacent discover row, resolved to main's 2.28.0 + author 3.8.0; (3) optional taken: `_DIRECTION_LITERAL` also matches the assigned form `s['direction'] = 'CLOSE'` and still ignores `== 'CLOSE'` comparisons (test added). Catalog sweep: 112 packages, 0 hits on either check. CI: parity ✓, golden ✓, full suite green (count in the run below).

PR #607, opened 2026-09-15
portfolio: the closed record comes from the engine, other templates from discover (1.23.0)

betashop:
## What the user hit

A user asked whether their strategy "only losing" was a problem. The review answered "20 trades, almost all losing and all long, one winner — win rate ~5%". The fills for that wallet: 9 winners out of 20, 7 shorts. Eleven minutes later, when the same user asked what to run instead, the review offered "a non-aggressive strategy (e.g. Kodiak or Polar)" from memory; both are `risk_level: aggressive` in the catalog, and the agent corrected itself only after running discover.

## Root cause

The engine's `closed` block printed only `realized_pnl`, `trade_count` and the five most recent trades, so the narration worked a win rate and a direction split out of the sample. And the portfolio review had no rule about naming other templates — discover's "only name what the engine returned" lives in discover, which is not loaded during a review.

## What changes

- `scripts/portfolio.py`: `closed` now carries the record over the whole pull — `winners`, `losers`, `win_rate_pct`, `longs`, `shorts` (a flat close is neither a winner nor a loser); `strategy_groups[].totals` sums the same counts across a strategy's wallets. A failed read prints `null`, never zero.
- `SKILL.md`: golden rule "Numbers come from the engine; other strategies come from discover" — never state a rate or a split the block did not print; another template is offered only through senpi-strategy-discover with its `risk_level` quoted, never named from memory. The realized-PnL bullet and the output contract describe the new fields.
- `tests/test_closed_stats.py`: the record over a 20-trade pull whose five most recent are all losing longs; null counts on a failed read; no win rate on an empty history; the SKILL needles.
- `run()` sorts `meta.warnings` before returning: the per-wallet reads run in a thread pool, so the warnings landed in completion order and `test_all_step_is_byte_identical_to_run` failed 2 runs in 5 on main. One line; 0/10 after.
- senpi-portfolio 1.22.0 → 1.23.0 (README + SKILL).

## Verification

The exact CI steps from `.github/workflows/tests.yml`, gated on exit codes:
- `test_min_budget_vendor_parity.py` → VENDOR PARITY OK
- `test_min_budget_golden.py` → GOLDEN OK, 110 templates
- `python3 -m pytest` over the listed suites → 930 passed, 9 skipped (4 new in `test_closed_stats.py`)
- the formerly flaky byte-identical test: 0 failures in 10 runs (2/5 on main before the sort)

## Detection

None in the funnel report: the truth is only in the tool results, and comparing a narrated win rate to the fills is not worth a query. The engine printing the record is the guard.

## Ponytail

Counts added in the loop that already walks the rows; the group rollup reuses the existing pattern. Skipped: per-asset breakdowns (nobody asked) and a separate "trades" verb (the block is already on every strategy).


shnoodles:
**Review — right diagnosis, and the warnings fix is a genuine catch. But the new counts have two arithmetic holes, in the one number this PR exists to make trustworthy.**

**Intent ✓.** Notion row (review narrated "win rate ~5%, all long" against fills of 9/20 winners and 7 shorts) → `closed` now carries the record over the whole pull instead of leaving the narration to extrapolate from `recent[]`, and SKILL.md gets the "numbers from the engine, templates from discover" rule. That is the right shape for the fix.

**Credit where it is due — the `sorted(meta["warnings"])` line is real.** I found the same race independently while reviewing #605, before reading this body. Measured: on `origin/main` `test_all_step_is_byte_identical_to_run` fails **5 of 8** `PYTHONHASHSEED` values; on this branch **0 of 12**. And `_all_and_persist` routes through `run()` (`portfolio.py:1888`), so both compared paths get the sort. One line, correct place.

---

### Required

**1. The win-rate denominator counts rows the loop never parsed.** (`portfolio.py:1127`)

The loop skips non-dict entries, then `n = len(rows)` counts them anyway:

```python
for p in rows:
    if not isinstance(p, dict):
        continue          # <- skipped here
...
n = len(rows)             # <- but counted here
"win_rate_pct": round(100.0 * winners / n, 1) if n else None
```

A pull with two malformed rows:
```
trade_count=5  winners=1 losers=1 longs=2 shorts=1  win_rate=20.0
```
The record over the rows actually parsed is 1 of 2 = **50%**; the block prints **20%**. `winners + losers = 2` and `longs + shorts = 3` against `trade_count = 5` — the counts and the denominator are drawn from two different populations, and SKILL.md now instructs the agent to quote the result as *the record*. Count the parsed rows instead of `len(rows)`.

**2. A missing or unparseable `szi` is counted as a LONG.** (`portfolio.py:1109`)

```python
szi = _f(p, "szi", "size", default=0.0)
...
if szi >= 0: longs += 1
else:        shorts += 1
```
A row with no `szi` at all:
```
longs=1 shorts=0   <- unknown side counted as LONG
```
This is the same shape as the bug in the ticket — the agent told that user "all long" when 7 of 20 were shorts. `recent[]` has always used `szi >= 0`, but there it is a display string on five rows; here it becomes a headline count the skill says to quote. Treat an absent/unparseable `szi` as neither side (and, if you want the split to stay honest, keep the remainder visible rather than folding it into longs).

### Minor

**3. "A failed read prints `null`, never zero" is true for five of six fields.** `empty` sets `trade_count: 0` (and the new test asserts it). That is the field feeding the group denominator, so in `strategy_groups[].totals` a failed wallet is indistinguishable from a wallet with no closed trades — one wallet at 10 trades / 5 winners plus one failed read reports `trade_count: 10, win_rate: 50.0` with nothing marking the gap. Either make it `None` like its five siblings, or drop the claim from the body.

### Ponytail (diff-scoped)

**4. Five prose needles in `test_skill_says_quote_the_record_and_route_other_templates_through_discover`,** one of which pins exact markdown punctuation:
```python
"`winners`, `losers`, `win_rate_pct`, `longs`, `shorts`"
```
Reorder that list or change a comma and the test goes red for no behavioural reason. Two needles carry the rule — the golden-rule sentence and `"from memory"`. Drop the rest.

Checked and **not** flagging: `_closed_count` sits next to a near-identical `realized_vals` walk, but one returns `int` and the other a rounded float, so folding them costs more than it saves. Leave it.

### Verified locally
Branch @ `2292dfd4`, python 3.14: parity ✓ · golden ✓ (110 templates) · full suite **926 passed, 9 skipped, 63 subtests, 0 failed** — the portfolio flake is gone, which is the first fully green run I have had across these four PRs. (Body says 930; I collect 926 = main's 922 + your 4. Probably a 3.11/3.14 collection difference, flagging only so the number in the ledger is not treated as exact.)

betashop:
Review addressed (`fetch_closed`, two commits): (1) the denominator is the rows the loop actually parsed — `trade_count = parsed`, never `len(rows)`; (2) a row with no readable `szi` is neither side — new `unknown_side` count in `closed` and in `strategy_groups[].totals`, `recent[].direction` prints `null` for it, so the split stays honest; (3) a failed read now prints `trade_count: null` like its siblings, so a failed wallet cannot masquerade as a wallet with no closed trades in the group totals; (4) needles down to the golden-rule sentence + `from memory`. Tests pin (1) and (2): two junk rows on two losers → 0 of 2, not 0 of 4; a missing size → `longs 0, shorts 1, unknown_side 1`. SKILL.md field lists carry `unknown_side`. #608 rebased on top. CI: parity ✓, golden ✓, 932 passed / 9 skipped, 0 deselected — your 926 vs my 930 was the 3.11/3.14 collection difference you guessed; this run is 3.14.

PR #608, opened 2026-09-15
portfolio: a raw position's protected:false is "no ratchet," never "no stop" (1.24.0)

betashop:
## What changes

A raw position (no runtime, `protected: false`) is narrated as **"no ratchet," never "no stop."** A one-off position the user placed by hand carries its protection as resting orders, which the portfolio engine does not read. Before the agent says such a position has no stop it reads `strategy_get_open_orders` on that wallet: a reduce-only trigger order (`isTrigger`, `triggerPx`) whose `orderType` is a Stop (`Stop Market` / `Stop Limit`) is its stop, a Take Profit order is not, and the agent says what it found — "no ratchet; a static stop rests at $X" or "no ratchet and no stop order."

The pre-existing rule "never infer 'no stop' from the absence of a resting stop order" is now scoped to runtime-managed positions (DSL exits are not venue orders); a raw position is the reverse case. The field pointer at the `protected` glossary entry defers to the HARD rule instead of restating it.

## Why

`protected: false` covers two states — a runtime-managed position below Tier 1, and a hand-placed position with no runtime — and only one of them means "unprotected". Heard as "no stop", the sentence sends a user to change a stop that was fine. The direct question ("do I have a stop?") is `senpi-trade`'s protection protocol, which reads the same orders in the turn that answers; the portfolio view had no such rule.

## Verification

`python3 -m pytest` over the nine test dirs → 943 passed, 9 skipped; `test_min_budget_vendor_parity.py` and `test_min_budget_golden.py` OK. `tests/test_skill_raw_position_stop.py` pins `strategy_get_open_orders`, the reduce-only trigger order, the Stop-vs-Take-Profit clause and the two narrations; each needle occurs once in SKILL.md.

## Ponytail

Prose plus one needle test. Skipped: reading open orders inside the engine (one more MCP call per wallet on every read is a bad trade for a narrow case) — senpi-trade already reads them on the direct question; the SKILL rule covers the portfolio view. Left for a follow-up (ledger row): the engine still flags a hand-placed one-off as `not_running` / "⛔ NOT RUNNING / UNPROTECTED — redeploy via ops" because senpi-trade stamps `skillName` on every custom strategy; that narration needs the same scoping.

Version 1.23.0 → 1.24.0; README + SKILL in one commit. Owner for review: Ignas.


shnoodles:
**Review — good distinction, correctly sourced identifiers. One claim in the write-up doesn't hold.**

**Intent ✓.** `protected: false` genuinely covers two different states, and only one of them means "unprotected". The user harm is the sharp part: they heard "no DSL" as "no stop" and went and moved a stop that was fine. Telling the agent to say "no ratchet; a static stop rests at $X" or "no ratchet and no stop order" fixes the sentence that caused it.

**Identifiers verified** — worth stating, because CLAUDE.md requires them copied from source and none of these appear anywhere else in `senpi-skills`:

| named in the diff | source |
|---|---|
| `strategy_get_open_orders` | real MCP tool |
| `isTrigger`, `triggerPx` | `senpi-hyperliquid-mcp/src/types/senpi.types.ts:275,277`, selected in `clients/senpi.client.ts:666-668`, documented at `resources/discovery-top-traders.resource.ts:214` alongside `reduceOnly` |

So "a reduce-only trigger order (`isTrigger`, `triggerPx`) is its stop" is accurate against the real response shape.

---

### The "Skipped" rationale cites a protocol that isn't there

> Skipped: reading open orders inside the engine (one more MCP call per wallet on every read; **the direct question already triggers senpi-trade's protection protocol, which reads them**).

It doesn't. `strategy_get_open_orders` appears in **no** skill in this repo — this PR is the first reference to it. `senpi-trade`'s protection section covers what a raw position *can carry* (bare / static stop-TP / `ratchet_stop_add` ladder) and how to *set* it; nothing there reads resting orders to discover a stop that already exists.

The decision itself is still right — one extra MCP call per wallet on every portfolio read is a bad trade for a narrow case. But the safety net named in its defence doesn't exist, so this SKILL rule is the *only* thing that will ever make the agent look. Worth saying plainly in the body, since "senpi-trade already covers it" is exactly the sentence a future editor will cite when trimming this bullet.

### Ponytail (diff-scoped)

Prose plus one needle test, which is the right size. Same note as the rest of this batch: four needles, one of which pins punctuation —

```python
'is "no ratchet," not "no stop."'
```

— comma inside the quotes and all. `strategy_get_open_orders` plus one phrase carries the rule.

### Verified locally
Branch @ pr-608 (based on `fix/portfolio-closed-stats`, retargets to main when #607 lands — as the body says): parity ✓, golden ✓ (110 templates), full CI suite **929 passed, 9 skipped, 63 subtests, 0 failed**. Version chain 1.23.0 → 1.24.0 consistent across title, README and SKILL.md, and it stacks correctly on #607's 1.23.0.

betashop:
Right on the "Skipped" rationale — nothing reads resting orders today, so this rule is the only thing that makes the agent look; the body now says so instead of citing a protocol that isn't there. Needles: the two punctuation-pinning ones are replaced by the reduce-only trigger order and the "no ratchet and no stop order" phrase, so the test carries the rule and survives a comma. `pytest senpi-portfolio/tests` → 91 passed. Head 35ab102, still stacked on #607.

betashop:
Second pass, and a correction to my earlier reply: senpi-trade's protection protocol **does** read resting orders on the direct question (1.4.0 rule 1: `ratchet_stop_get` + `strategy_get_open_orders` in the turn that makes the claim) — the original body sentence was right and I should not have adopted the correction; the body now says so. Changes: a resting **Stop** (`orderType` Stop Market / Stop Limit) is the stop, a Take Profit order is not (both are reduce-only triggers; verified on a live `strategy_get_open_orders` read); the pre-existing "never infer no stop from the order book" bullet is scoped to runtime-managed positions; the glossary pointer no longer restates the rule; the incident sentence is gone. Left for a follow-up (ledger row): the engine still flags a hand-placed one-off as `not_running` / "⛔ NOT RUNNING / UNPROTECTED" because senpi-trade stamps `skillName` on every custom strategy. Rebased onto main (1.24.0 on 1.23.0); CI 943 passed, 9 skipped. Head 514466b.

shnoodles:
**Correction — my "Skipped rationale" finding was wrong, and you were right to change it back.**

`senpi-trade/SKILL.md` does read resting orders on the direct question: `strategy_get_open_orders` appears at lines 49, 65, 69 and 82, and line 65 is exactly the protocol you cited — `ratchet_stop_get` + `strategy_get_open_orders` in the turn that makes the claim. The original body sentence was accurate.

The cause, so it is not mysterious: I grepped the working tree while it sat on an old branch that predates `senpi-trade 1.4.0` (#595, `20c4ee16`), instead of `origin/main`. Same mistake I had already made once in this batch with a stale local `main`, which is on me.

Sorry for the churn — you changed the body on my say-so and then had to change it back. The rest of that review stands, and your second-pass changes (a resting **Stop** is the stop, a Take Profit is not; scoping the "never infer no stop" bullet to runtime-managed positions) are both improvements I did not catch.

Re-verified at head `514466b` against current `main`: CI **943 passed, 9 skipped**, version chain 1.24.0 on main's 1.23.0. No further findings.

betashop:
Noted, thanks — the body stands as pushed (senpi-trade reads the resting orders on the direct question; this rule covers the portfolio view).

PR #610, opened 2026-09-15
discover 2.30.0: fee math on micro-specs; link the public source for audits

betashop:
## What changes

Two bullets in the discover skill's intent table:

- **Micro-specs** ("$1 a trade, a stop of a few cents, a hundred small wins") → do the fee math out loud before matching anything: the round-trip cost is Hyperliquid's taker rate on both legs (the wallet's `userFees`, never assumed) plus Senpi's builder fee (`get_loyalty_tiers`, `feePercent` — the only fee that tool carries), put next to the requested stop. A position big enough to make a dollar pays a fee that is a large share of a cents-wide stop, and a stop that tight is hit by normal noise. Say the numbers, then offer what exists: a template's stop and lock ladder sized to its budget. Never save an impossible spec as a "risk profile", never answer "Senpi has exactly that".
- **"Show me the source" / "I want an outside audit"** → the catalog is public: link `https://github.com/Senpi-ai/senpi-skills/tree/main/strategies/<id>`; for a fork, point at their package (`<username>-<template>`) and its diff against that link. Never hand-roll a tarball or a hash-stamped dump in place of the link.

## Why

Two classes of user request the agent answered from memory instead of from the fee table and the public repo: an impossible micro-spec was saved as a risk profile and confirmed as available, and a source request for an outside audit got a local dump instead of the public link.

## Verification

`python3 -m pytest` over the nine test dirs → 943 passed, 9 skipped; `test_min_budget_vendor_parity.py` and `test_min_budget_golden.py` OK. The needle test pins the fee rule ("do the fee math out loud", "the only fee that tool carries"), the impossible-spec rule, the public link prefix and the no-tarball rule.

## Ponytail

Two bullets, one test. Cut in review: needles that pinned markdown bold and the `<id>` placeholder spelling; a user narrative in the test docstring.

Version 2.29.0 → 2.30.0 on top of 2.29.0 (#605, merged); rebased onto main. Owner for review: Ignas.


shnoodles:
**Review — both bullets check out, including the two claims that could quietly be wrong. No change requested.**

**Intent ✓** for both rows. One user was told "Senpi has exactly that" for "$1 a trade, never lose more than 20 cents" and had it saved as a risk profile; another asked for source for an outside audit and got a hand-built tarball. The two bullets answer exactly those.

**The claims I checked rather than assumed:**

- **`get_loyalty_tiers` is the right tool**, not a plausible-looking one — the MCP names it *"THE CANONICAL FEE SOURCE — call it before answering ANY question about Senpi fees … never quote fees from memory (the table changes)"*. Telling the agent to read round-trip cost from it is exactly right, and "out loud" is the part that fixes the row.
- **The public link resolves.** `Senpi-ai/senpi-skills` is `visibility=PUBLIC`, and `strategies/<id>` is a real path (spot-checked `strategies/barracuda` via the API). A link to a private repo would have turned "here's the source" into a 404 in front of someone doing an audit — worth confirming before shipping it as the standing answer.

**Version chain ✓** — 2.29.0 → 2.30.0 on top of #605's 2.29.0, README and SKILL.md agree with the title, and the base is correctly `fix/discover-proven-templates` pending #605.

### Verified locally
parity ✓, golden ✓ (110 templates), CI suite across three hash seeds:

```
seed=0: 924 passed, 9 skipped, 63 subtests
seed=1: 1 failed, 923 passed
seed=2: 1 failed, 923 passed
```

The intermittent one is `test_all_step_is_byte_identical_to_run`, exactly as the body says — the thread-order flake that #607 fixes and that this base predates. The body's characterisation is honest; I reproduced both outcomes.

### Ponytail (diff-scoped)
Two bullets, one test. Nothing to cut. Same standing note as the rest of this batch: five needles, two of which pin formatting rather than rule — `"**Micro-specs**"` carries markdown bold, and the URL needle pins the literal `<id>` placeholder, so re-writing the link as a real example breaks it. `"do the fee math out loud"` plus the bare URL would hold the same ground.

betashop:
Thanks — no change. Noted the needle comment (the markdown bold and the literal `<id>` placeholder); I will swap those for "do the fee math out loud" plus the bare URL the next time this file is touched rather than churn the PR now.

betashop:
Second pass: `get_loyalty_tiers` carries only Senpi's builder fee (`feePercent`), never the Hyperliquid taker rate — the bullet now says where each leg comes from (the wallet's `userFees` for taker, never assumed; `get_loyalty_tiers` for the builder fee only); the fork is named the way ops names it; the needles carry the rule instead of markdown bold and the `<id>` spelling; the test docstring inherited from #605 no longer carries a user narrative. Rebased onto main (2.30.0 on 2.29.0); CI 943 passed, 9 skipped. Head 36da705.

One hygiene item on this thread rather than the diff: the review comment above names two user IDs on a public repo — could the author edit them out?

shnoodles:
**Done — user IDs removed.** Good catch, and it applied to more than this thread: `Senpi-ai/senpi-skills` is the only **public** repo in this batch, and five of my review comments on it named MIDs (#605, #606, #607, #610, #611). All five are edited; the other three repos are private, so the ledger references there are fine as they are.

**Second pass on the fee change — agreed, and it is a real correction.** `get_loyalty_tiers` returning only Senpi's builder fee makes my "canonical fee source" praise half right: it is canonical for *our* leg, not the Hyperliquid taker rate. Naming both legs separately (the wallet's `userFees` for taker, `get_loyalty_tiers` for the builder fee) is what makes the micro-spec arithmetic actually land, since the taker leg is the larger of the two against a 20-cent stop.

No further findings. Re-verified at head `36da705`: rebased on main, 2.30.0 on main's 2.29.0, CI **943 passed, 9 skipped**.

PR #611, opened 2026-09-15
ops 3.18.0: a backend SERR124 is this account's empty wallet pool — route it, never "fix it and re-run", never a loop

betashop:
## What the user hit

A deploy that the backend refused with `SERR124: No approved strategy wallet available in the pool (free=0, unapproved=20, …)`. The ops outcome table's generic exit-3 advice — "read the quoted cause, fix it, re-run" — turned that into a 30-second retry loop while the app told the user to retry; a second agent alternated `deploy.py create` with raw MCP creates to route around it.

## Root cause

Strategy-wallet pools are **per account**, and this code has only been seen on some migrated older accounts (`unapproved=N` means that account's approval step is stuck). Nothing was created or debited, other accounts deploy normally, and no retry refills the pool — someone at Senpi does. The skill had no sentence for a backend refusal the box cannot fix, so the agent treated it as a fixable cause. (An earlier revision of this PR called it a platform outage; corrected 2026-09-15.)

## What changes

- `SKILL.md` exit-3 row (at the line budget, so it rides the existing row): a backend `SERR124` is the third exception to "fix it, re-run" — per-account, nothing created or debited, tell the user, route it to Senpi support, one later attempt after they confirm, never a retry loop.
- `references/refusal-playbook.md`: new section "Backend rejections the box cannot fix / `SERR124`" with the one sentence for the user, the routing step, and the five Nevers (no 30-second loop or timer, no lower budget or template switch, no raw MCP create around the verb, no closing another strategy to "free a wallet", no "Senpi is down").
- `docs/error-code-taxonomy.md`: the `SERR124` row beside `SERR037`.
- `tests/test_skill_serr124.py`: pins the row, the playbook section and the taxonomy row.
- ops 3.17.0 → 3.18.0, README row.

## Verification

`python3 -m pytest senpi-strategy-ops/tests` → 418 passed, 1 skipped; `python3 senpi-strategy-ops/tests/test_skill_serr124.py` → OK. Body stays at 331 lines (`BODY_BUDGET`).

## Detection

Funnel-report F32 (senpi-internal-skills #74): one row per user with a `SERR124` in the window, plus the counter-evidence line (other users creating), so the report pages needs-a-human for that account and never a platform line. The MCP's curated message is senpi-hyperliquid-mcp #213.

## Ponytail

The exit-3 cell carries the rule only; depth is in the playbook. Owner for review: Ignas.


betashop:
CI as the workflow runs it: `python3 senpi-strategy-ops/tests/test_min_budget_vendor_parity.py` → VENDOR PARITY OK; `python3 senpi-trading-runtime/tests/test_min_budget_golden.py` → GOLDEN OK (110 templates); pytest over all nine test dirs → 928 passed, 9 skipped, 63 subtests passed, 1 deselected (`test_all_step_is_byte_identical_to_run`, the thread-order flake fixed in #607). Ops body stays at 331 lines (the budget).

shnoodles:
**Review — correct call, and the scope split is right. One thing to confirm before merge, because the whole fix hangs on it.**

**Intent ✓.** Notion P0 (pool `free=0` from ~13:01Z, 9 `strategy_create_custom_strategy` calls plus repeated `deploy.py` over 50 minutes, every one refused, app still saying "retry") → the outcome table's generic *"read the quoted cause, fix it, re-run"* is named as the sentence that produced the loop, and `SERR124` joins the two existing exceptions. Deferring the curated message to senpi-hyperliquid-mcp #213 and the refill to the platform is the right cut — this is the agent's half and it does not pretend otherwise. The four "never"s in the playbook are each a behaviour actually seen in that transcript, including closing another strategy to free a wallet.

---

### Confirm before merge

**1. The guidance is only in the exit `3` row — is `SERR124` actually exit 3?**

The taxonomy row you add asserts it: *"relayed by `senpi deploy` as a `failed` step"*. Nothing in this repo establishes that. `deploy.py` does not classify it — `exit_code_for` (`deploy.py:645`) reads `state.overall` straight from the verb and its docstring says *"Mirrors the verb's own `exitCodeForDeploy`, exactly"*, and the runtime repo is not in this tree, so I could not check it.

If the verb classifies a rejected create as `refused` instead, the agent reads the exit `2` row — which enumerates bracketed gate codes (`[E_FUNDS_*]`, `[E_VALIDATE_*]`, `[E_UNIVERSE_NOT_LIVE]`, …), none of which match an unbracketed backend `SERR124` — finds no match in either row, and lands back on the generic advice this PR exists to override. The user-visible outcome is the 50-minute loop, unchanged.

Two ways to close it, either is fine:
- confirm the mapping against `exitCodeForDeploy` and say so in the PR, or
- name `SERR124` in both rows. The playbook section is already keyed by code rather than by exit, so it costs one pointer and stops depending on a classification made in another repo.

### Ponytail (diff-scoped)

**2. The exit-3 cell grew 501 → 806 chars (+61%) restating what the playbook says in depth** — in a body sitting *exactly* at its budget:

```
BODY_BUDGET = {"senpi-strategy-ops": 331, ...}      # ops body: 331 lines
```

The guard counts **lines**, and this change is `+2/-2` on one line, so it saw nothing. That is the case for a human looking at it rather than a reason it is fine. Per CLAUDE.md — hot paths carry rules, rationale lives in reference files — the resident cell needs the code, "outage, nothing created or debited", "once after ~30 min, never a loop", and the pointer. *"and never a raw MCP create to route around it"* plus the parenthetical are playbook material and are already there, verbatim. Trim the cell, keep the playbook.

**3. Two test lookups raise instead of failing.** `next(l for l in ... if l.startswith("| `3` |"))` raises `StopIteration` and `_playbook_section`'s `text.index("### `SERR124`")` raises `ValueError` if either surface is reformatted — an error with no message rather than "the row is gone". `next(..., None)` plus an assert costs nothing and says which surface moved.

---

### Verified locally
Branch @ `354aec5e`, python 3.14: parity ✓ · golden ✓ (110 templates) · suite **924 passed, 9 skipped, 1 deselected, 63 subtests** — matching the body's deselect of `test_all_step_is_byte_identical_to_run`, which is #607's thread-order flake (5 of 8 `PYTHONHASHSEED` values red on `main`; #607's `sorted()` fixes it). Version `3.17.0 → 3.18.0` consistent across title, README and SKILL.md ✓. Taxonomy row sits beside `SERR037` as described ✓. Merges clean against `main` ✓.

**One note across all four PRs in this batch** (605 / 606 / 607 / 611): every new test pins agent-facing prose with 4–6 literal substrings, one of which here is a full table row prefix and in #607 an exact backtick-comma list. It makes the suite go red on rewording that changes no behaviour, and it will be the automated author's own next obstacle when it edits these lines again. One or two needles carrying the rule is the same guard with less friction.

betashop:
Confirmed against the runtime source (`src/deploy/orchestrator.ts` @ main): a backend rejection of the create call is recorded as `record(run, "create", { status: "failed", … "wallet creation failed: <backend message>" })` — only the name-rejection class (SERR055/056/058) gets a retry, everything else lands as a failed create step — and `decideOverall` returns `refused` only for the preflight/reconcile gates, else `failed` whenever any step is `failed`. `exitCodeForDeploy` maps `failed` → 3. So SERR124 arrives on the exit-3 row, and the exit-2 row needs no pointer; the taxonomy row's "relayed as a failed step" stands. Ponytail applied (`3d3f…`): the exit-3 cell is back to the rule — code, "outage, nothing was created or debited", "once after ~30 min, never a retry loop", pointer (806 → 592 chars; depth stays in the playbook); the two test lookups now fail with a message instead of raising. Body still 331 lines. CI: parity ✓, golden ✓, 929 passed / 9 skipped. Agreed on the needle note across the batch — #605 and #607 are trimmed to the two that carry the rule; #606's tests already were.

betashop:
Re-pushed with a correction to the premise, from Jason: strategy-wallet pools are **per account**, and SERR124 has only been seen on some migrated older accounts — not a platform outage, other accounts deploy normally. The exit-3 row, the playbook section, the taxonomy row and the test now say per-account: tell the user, route it to Senpi support, one later attempt after they confirm, never a retry loop, never "Senpi is down". Same reframe in senpi-hyperliquid-mcp #213 and senpi-internal-skills #74, so the three still agree.

`python3 -m pytest senpi-strategy-ops/tests` → 418 passed, 1 skipped; `test_skill_serr124.py` → OK; body still 331 lines. Title and body updated.

PR #632, opened 2026-09-15
runtime 4.1.2: hard_timeout fires in either phase — the concepts reference said Phase 1 only

betashop:
## What was wrong

`senpi-trading-runtime/references/runtime-concepts.md` told authors that `hard_timeout` is Phase 1 only and never fires once a position enters Phase 2. The runtime's own `dsl-reference.md` says the opposite: the preset-root time cuts (`hard_timeout`, `weak_peak_cut`, `dead_weight_cut`) "have no phase guard and still fire". An author reading the skill's reference would leave a 6h or 12h clock on a let-winners-run template believing it could never cut a tier-armed winner; it can, and did.

## What changes

- The phase-behaviour bullet, the `hard_timeout` section and the summary-table row now say: counts wall-clock minutes from open, fires in either phase, never resets — a profit-taker on a winner as much as a give-up on a laggard, which is why let-winners-run templates disable it.
- `tests/test_reference_time_cuts_either_phase.py` pins the wording (two positive needles, two negative).
- senpi-trading-runtime 4.1.1 → 4.1.2 (PATCH, per this skill's rule); README row in the same commit.

## How verified

`python3 -m pytest senpi-trading-runtime/tests -q` green; exact CI over the nine test dirs green; `test_min_budget_vendor_parity.py` and `test_min_budget_golden.py` OK. Source of truth: the runtime repo's `docs/runtime-docs/dsl-reference.md:123` (checkout of 2026-09-11), the `hard_timeout` section itself: the timer starts at entry, runs across both phases, never resets, with the single-tick exception on the first Phase-2 crossing. The runtime already made this correction in its own guide and pinned it (`src/cli/guide-commands.ts:240`, `guide-commands.test.ts:108` "guide dsl no longer claims hard_timeout is phase-1 only"); the skills reference lagged it.

## Ponytail

Three sentences replaced, one row, one test. `retrace_threshold (Phase 1 only)` and the `weak_peak_cut` note ("fires in both phases; practically Phase 1 only") were already right and are untouched. Found while removing the time cuts from Raptor, Mantis and Cheetah (#628–#630). Owner for review: Ignas.


shnoodles:
**Review — the correction is right, and better supported than the PR claims. One citation to move, one needle to loosen.**

**Intent ✓.** The old text told authors that entering Phase 2 disarms `hard_timeout`. It does not, and a 6h/12h clock left on a let-winners-run template can cut a tier-armed winner. The rewrite says counts-from-open, either phase, never resets — which is exactly right.

### The source of truth is one line further down than the one you cited

The body cites the `phase1.enabled` row. That row's *"they have no phase guard and still fire"* is scoped to what happens when you set `phase1.enabled: false` — true, but a narrower claim than the one this PR makes. The line that states your rule directly is **`docs/runtime-docs/dsl-reference.md:123`**, the `hard_timeout` section itself:

> Close when the position has been open for at least N minutes. The timer starts at entry (`hardTimeoutStartedAt`) and runs across both phases — it never resets — so `hard_timeout` fires in Phase 1 **or** Phase 2 once `interval_in_minutes` elapses. Exception: on the single tick that first crosses into Phase 2 via a new tier, tier advance wins and that one tick does not emit `hard_timeout`.

That is your new paragraph, including the single-tick exception, almost sentence for sentence. Cite 123 — a reader who checks 108 finds a clause about a different question and has to go hunting.

**Corroborated twice more in the runtime repo:**

- `src/cli/guide-commands.ts:240` — *"fires in either phase once `interval_in_minutes` elapses since entry (the timer …)"*
- `src/cli/guide-commands.test.ts:108` — `it("guide dsl no longer claims hard_timeout is phase-1 only", …)`

That second one is worth knowing: **the runtime already made this exact correction in its own guide and left a regression test behind.** The skills reference simply lagged it. It is also the precedent for your new test, and worth a line in the body — it turns "we think this is right" into "the owning repo fixed the same sentence and pinned it".

**The catalog side is already closed.** Checked every shipped recipe: **0** templates combine `let_winners_run` with a `hard_timeout` that is not explicitly disabled — #628–#630 took care of that. So this doc change is the last piece, not a warning about live packages.

### Ponytail (diff-scoped)

One needle pins a line wrap:

```python
assert "entering Phase 2 does\n  not disarm `hard_timeout`" in text
```

The `\n  ` is the current fold plus its two-space continuation indent. Reflow that bullet — or let a formatter touch it — and the test goes red for no behavioural reason. The two negative needles (`"hard_timeout` is Phase 1 only"`, `"Phase 1 only. |"`) are the ones that actually guard the regression, since they fail if the old claim ever comes back. Drop the wrap-sensitive one or match on `"not disarm"` alone.

### Verified locally
Branch @ pr-632 (3 behind main): parity ✓, golden ✓, full CI suite **939 passed, 9 skipped, 63 subtests, 0 failed**. Version 4.1.1 → 4.1.2 (PATCH, correct for a reference correction) consistent across title, README and SKILL.md.

betashop:
Both taken: the body now cites `dsl-reference.md:123` (the `hard_timeout` section) and the runtime's own correction with its regression test (`guide-commands.ts:240`, `guide-commands.test.ts:108`); the wrap-sensitive needle is gone — the test matches `not disarm \`hard_timeout\`` plus the two negative needles. `pytest senpi-trading-runtime/tests` → 3 passed. Head 3c260a5.

PR #633, opened 2026-09-15
author 3.9.0: a 4h-board read is momentum, not smart money — the validator says so

betashop:
## What changes

- `senpi-strategy-author/scripts/validate_strategy.py` — two checks added to the existing lint, in the section and shape of 3.8.0's close-wiring checks (same `errs.append` reporting, the per-file check next to `direction_literal_offenders` inside the scanner loop, the package-level check after it):
  - **(a) and (b)** — a scanner under `scanners/` references `pct_of_top_traders_gain` / `longPct` / `long_pct`, no scanner in the package calls `discovery_get_trader_state`, and the package text says "smart money" / "smart-money" / "smart_money" (case-insensitive) in `strategy.yaml` `catalog` (name, tagline, belief_plain, thesis, tags, group — what discover ranks on) or `description`, a runtime.yaml `description` (the mandate the runtime registers), or `README.md`. One error per package, naming the scanner file(s) and the first mention's file, field and phrase.
  - **(c)** — a `"LONG" if … <pnl field> … >= 0 else "SHORT"` ternary (either order, any field whose name contains `pnl`, compared with `0` / `0.0`) in a scanner file. One error per file.
  - Helpers `pnl_sign_directions(src)` and `smart_money_mentions(pkg)`; regexes `_BOARD_SIDE`, `_SMART_MONEY`, `_PNL_SIGN`. Grep-level on purpose — no AST, no config, no new module.
- `senpi-strategy-author/tests/test_validate_momentum_not_smart_money.py` (new) — one failing fixture per class (needles: the phrase and the file it names), two passing fixtures (a board read described as "4h leader momentum"; a cohort read via `discovery_get_trader_state` described as smart money), pure-helper cases, and the catalog sweep marked `xfail(strict=True)` (see staging).
- `senpi-strategy-author/references/creating-a-strategy.md` — decision 2 (Data) splits *4h-board momentum* (`leaderboard_get_markets`) from *smart-money* (`discovery_get_trader_state` on the proven cohort); the checklist carries the rule.
- `senpi-strategy-author/SKILL.md` — 3.8.0 → 3.9.0; decision 2's data line lists `leaderboard_*` under 4h-board momentum instead of smart-money (edited in place: the body stays at its 425-line budget). `README.md` row bumped to match.

## Why

`leaderboard_get_markets` ranks markets by who was winning over the last four hours, so a direction taken from `pct_of_top_traders_gain` or `longPct` is a momentum read of the board: the top of the board is a consequence of the move, and calling that "smart money" is circular. The skill already says a smart-money thesis reads the proven cohort through `discovery_get_trader_state`; the lint now holds the name to the read, so what a package says about itself in the catalog and in its registered description matches what its scanner actually does. And a PnL sign is not a side — a short is green when the price falls — the side is the sign of `szi`.

## The rule

```
<scanner files>: momentum read of the 4h board described as smart money (<file> <field>: '<phrase>') — rename to 4h leader momentum, or read the proven cohort via discovery_get_trader_state
<file>.py: direction from a PnL sign (`<field>`) is not a direction — use the position's side (szi sign)
```

Two things worth knowing before the rename pass: the phrase is refused wherever it appears in those fields, negated or not (a grep cannot honor "not smart money" — leave the phrase out and say *4h leader momentum*); and a runtime.yaml `description` counts as self-description alongside the catalog fields, because the runtime registers it as the mandate — 5 of the 47 hits are on that field alone.

## Verification

```
/usr/bin/python3 -m pytest senpi-strategy-ops/tests senpi-strategy-discover/tests senpi-strategy-author/tests senpi-trading-runtime/tests senpi-trader-research/tests senpi-market-pulse/tests senpi-portfolio/tests senpi-trade/tests strategies/tests -q
  947 passed, 9 skipped, 1 xfailed          (main before this change: 942 passed, 9 skipped)
/usr/bin/python3 senpi-strategy-ops/tests/test_min_budget_vendor_parity.py   → VENDOR PARITY OK
/usr/bin/python3 senpi-trading-runtime/tests/test_min_budget_golden.py       → GOLDEN OK — 110 templates
/usr/bin/python3 -m pytest senpi-strategy-author/tests -q                    → 223 passed, 1 xfailed
```

Catalog sweep (`strategies/*`, 112 packages, the same walk as the xfail test): 54 board readers — exactly the set of packages calling `leaderboard_get_markets`, so the field regex is a faithful proxy; 6 exempt because a scanner reads the proven cohort (athena, chimp, gibbon, gorilla, orangutan, phalanx); falcon already names itself honestly; **47 packages hit (a) and (b)**; raptor hits (c) on `main` (the side-from-position fix is on its own branch, #621). The 47 are listed one per line in `LINT_HITS.md` — the worklist for the rename pass; nothing is renamed here.

## Staging

The repo has no staged-rule mechanism, so the catalog sweep test is `xfail(strict=True, reason="lands with the rename pass")`: it turns XPASS → FAIL the moment the catalog is clean, which is the cue to drop the marker. The rule itself is not weakened — `validate_strategy.py <pkg>` exits 1 on those 47 packages today. The deploy gate is unchanged: ops' `_pkg.validate` mirrors the author lint rather than importing it.

## Ponytail

Three regexes, two helpers (one is the YAML string walker that names the field), ~60 lines with comments, in the file the rules already live in. A `"direction" in src` condition and a package-wide (non-scanner) `discovery_get_trader_state` scan were dropped after proving the hit list is identical without them (47/47, same files).

Review: Ignas.

**Update 2026-09-15 22:35Z** — rebased onto main after the rename wave (#635–#640); the catalog sweep test `test_catalog_has_no_board_read_sold_as_smart_money` no longer carries `xfail(strict)`: it passes on main (222 author tests). Full CI green on the rebased head.


shnoodles:
**Review — both checks are sound and I could not make them misfire. One scoping gap worth a line.**

**Intent ✓.** The circularity is real and worth a lint: `leaderboard_get_markets` ranks by who was winning over the last four hours, so a side taken from `pct_of_top_traders_gain` / `longPct` is *momentum*, and calling it smart money is describing the board's winners as though they were a vetted cohort. Tying the naming rule to whether the package actually reads `discovery_get_trader_state` is the right discriminator — it is a fact about the code, not a judgement about the prose.

**No regression.** Ran the validator over **every** package in `strategies/`: **0 hits** on either new check. So this lands as a guard on future packages rather than a pile of work on existing ones.

**The `_PNL_SIGN` regex holds up.** It is the fiddly part, so I probed it rather than reading it:

| source | fires |
|---|---|
| `"LONG" if delta_pnl >= 0 else "SHORT"` | ✓ |
| `"SHORT" if pnl < 0 else "LONG"` (reversed) | ✓ |
| `"LONG" if row["realizedPnl"] > 0 else "SHORT"` | ✓ |
| `"LONG" if p.get("unrealizedPnl") >= 0 else "SHORT"` | ✓ |
| `"LONG" if pnl_usd >= 0.0 else "SHORT"` | ✓ |
| `"LONG" if total_pnl_pct >= 0 else "SHORT"` | ✓ |
| **`"LONG" if szi >= 0 else "SHORT"`** — the correct idiom | **silent** ✓ |
| split across two lines | missed |

The only miss is the line-split form, and the docstring already says "Grep-level on purpose", so that is a stated limit rather than a gap. Not firing on `szi` is the one that matters and it does not.

---

### The cohort-read exemption only looks inside `scanners/`

```python
scn = {str(p.relative_to(pkg)): p.read_text() for p in sorted(pkg.rglob("*.py")) if "scanners" in p.parts}
board = [f for f, s in scn.items() if _BOARD_SIDE.search(s)]
if board and not any("discovery_get_trader_state" in s for s in scn.values()):
```

The board search is correctly scoped to scanners — that is where a direction gets decided. But the **exemption** is scoped the same way, so a package that reads the proven cohort from a helper outside a `scanners/` directory is refused for describing itself as smart money when it has earned the phrase.

Nothing in the catalog is shaped that way today (hence 0 hits), so this is about the next fork rather than a bug now. Widening just the exemption to `pkg.rglob("*.py")` costs one expression and cannot create a false positive — it can only stop one.

### Ponytail (diff-scoped)
Two pure helpers, one regex pair, one error branch each — consistent with how the rest of `validate_strategy.py` is built, and `_strings()` walking parsed YAML is the right way to reach every user-facing field rather than grepping the raw file. `smart_money_mentions(pkg)[:1]` reporting only the first mention keeps it to one line per package, which is the file's convention. Nothing to cut.

### Verified locally
Branch @ pr-633, 0 behind main: `senpi-strategy-author/tests` **222 passed**, full CI **1,096 passed, 9 skipped, 0 failed**, and the validator clean across every catalog package. Version 3.8.0 → 3.9.0 consistent across title, README and SKILL.md.

betashop:
Agreed, and widened in 25335a1 — with one narrowing you may want to push back on.

**Widened, but not into `tests/`.** `pkg.rglob("*.py")` also reaches a package's own tests, and a test that mocks `discovery_get_trader_state` would buy the exemption for a package whose production code never reads the cohort. So the exemption now reads every module **outside `tests/`**. The board search stays scoped to `scanners/`, as you said, since that's where a direction is decided.

Two tests, and each half is proven by reverting just that half:

| mutation | result |
|---|---|
| exemption back to `scanners/` only | `test_cohort_read_in_a_helper_outside_scanners_passes` **fails** |
| exemption reaches into `tests/` | `test_a_test_mock_of_the_cohort_read_does_not_buy_the_exemption` **fails** |

Restored: 8 passed in the file, author suite **224** (was 222). The catalog-wide test is still at **0 hits**. Widening can only remove refusals, and the narrowing only matters for a package with a mock in its tests, which none has today.

shnoodles (approved):
Approved at `25335a19`. The exemption widening went further than I asked and better: excluding `tests/` stops a mocked `discovery_get_trader_state` buying the exemption for a package whose production code never reads the cohort. Both halves proven by reverting each one. Still 0 hits across the catalog, and the `_PNL_SIGN` regex fires on 6 realistic variants while staying silent on the correct `szi` idiom.

PR #641, opened 2026-09-15
discover 2.31.0: the signal-followers list and two glosses after the rename wave

betashop:
## What

senpi-strategy-discover **2.30.0 → 2.31.0** (MINOR per the skill rule). Based on `main` — #610 (2.30.0) merged on Sep 15 at 22:51Z, and this branch is rebased on top of it.

Three text changes that trail the rename wave (#635–#640) and the engine ports (#616/#627/#631):

- **SKILL.md, hands-off copy flavors:** *follow the smart money by signal* now lists **Athena / Phalanx / Starling / Whalehunter / Pilotfish**. Stingray was retired in #635; Pilotfish runs the shared Phalanx engine since #627. "many traders at once" → "many wallets at once" (the engine counts wallets).
- **references/glossary.yaml:** `sm_crowding` and `risk_off_rotation` glosses say "the 4h leaders" instead of "smart money". `risk_off_rotation` is inlined into hyena's `sub_style_label` in both catalog.json copies — the one string the rename wave could not fix from inside a template PR. `sm_crowding` has no catalog user left (egret, dog, lemon moved to `crowd_exhaustion`) and keeps its key for forks that declare it.
- **README row** 2.31.0.

The glosses for the proven-cohort styles (`cohort_divergence`, `sm_rotation`, `sm_hedge_fund`) still say smart money on purpose — those templates read `discovery_get_top_traders` + `discovery_get_trader_state`, which is what smart money means in this catalog.

## Verification

- Exact CI on this branch, rebased on `main` Sep 16: `1090 passed, 9 skipped`; vendor parity OK; golden OK (109 templates). Catalogs regenerated after the rebase — only the `_updated` stamp moved.
- `gen_catalog.py --updated 2026-09-15` regenerated both copies — the only catalog diff against the base is hyena's `sub_style_label`. Regenerate once more on the rebase (the catalog.json copies will conflict with main's, as every template PR's did; the regen is the resolution).
- Hygiene grep of added lines: no user ids, wallets, incidents.

## Ponytail

Three strings and a version. Not touched: the discover ranking, facets, tags (`leader-momentum` already flows in from the templates' `strategy.yaml`), the SKILL's Athena guidance (unchanged and correct).


shnoodles:
**Review — every factual claim checks out. One sentence got mangled in the edit.**

**Verified, since this PR is entirely claims about what is still true:**

| claim | checked |
|---|---|
| Stingray retired (#635) | ✓ not in `catalog.json` |
| Pilotfish runs the shared Phalanx engine (#627) | ✓ present in catalog; #627 merged |
| `sm_crowding` has no catalog user left | ✓ zero packages declare it |
| `risk_off_rotation` is hyena's | ✓ hyena, and only hyena |
| catalogs regenerated, not hand-edited | ✓ re-running `gen_catalog.py` changes only the `_updated` stamp — the label genuinely flows `glossary.yaml` → both copies |

Keeping `sm_crowding`'s key for forks that declare it is the right call, and the note about *why* `cohort_divergence` / `sm_rotation` / `sm_hedge_fund` keep saying "smart money" — those read `discovery_get_top_traders`, which is what the phrase means in this catalog — is exactly the justification clause that stops a future editor "finishing" the rename.

---

### The paper-trade line is now ungrammatical, and it is the line that names a number

`senpi-strategy-discover/SKILL.md:198-200` currently reads:

> The trial IS a live run at **the floor is the card's `min_budget`** — quote that number; the $10 platform floor is only the per-wallet minimum inside it, never the number to give a user, **after `senpi validate` proves it runs**; say so, and never offer to watch it on a timer

Two things broke when the parenthetical `(min_budget, about $10 per wallet)` was replaced:

1. **"a live run at the floor is the card's `min_budget`"** — the verb is orphaned; the sentence no longer parses.
2. **"after `senpi validate` proves it runs"** used to attach to *"a live run at the floor"*. It now trails *"never the number to give a user,"*, so it reads as if validation is a condition on which number to quote.

The substance is right and worth shipping — quoting the card's `min_budget` rather than the $10 per-wallet floor is correct, and matches `catalog.json`'s own `_instructions`. It is only the splice. Something like:

> The trial IS a live run at the floor, after `senpi validate` proves it runs — and the floor is the card's `min_budget`, which is the number to quote. The $10 platform floor is the per-wallet minimum inside it, never the number to give a user. Say so, and never offer to watch it on a timer …

This is agent-facing copy the agent paraphrases to users, and the whole point of the change is which number reaches them, so it is worth a clean sentence.

Minor: that line is now ~250 chars against ~100 for its neighbours — rewrapping it while fixing the grammar keeps the file scannable.

### Ponytail
Nothing to cut. Nine lines changed, three of them one word each, no new test — correct: the catalog guards already cover the generated strings, and a needle test for a gloss would pin prose.

### Verified locally
Branch @ pr-641, 1 behind main: full CI **1,089 passed, 9 skipped, 0 failed**. Version 2.30.0 → 2.31.0 consistent across title, README and SKILL.md. Body says it stacks on #610 (2.30.0) — #610 is still open, so this needs #610 first or a rebase.

betashop:
Sentence fixed, using your wording, in 50c3c11:

> The trial IS a live run at the floor, after `senpi validate` proves it runs — and the floor is the card's `min_budget`, which is the number to quote. The $10 platform floor is the per-wallet minimum inside it, never the number to give a user. Say so, and never offer to watch it on a timer — an `openclaw cron` job is a model call per firing, not a simulation.

Rewrapped to the file's width — 110 / 108 / 105 / 107 characters, down from ~250. Discover tests 26 passed.

**On #610 — that part is out of date.** #610 merged on **Sep 15 at 22:51Z** (`9d0d82e`). The "stacked on #610, merge after it" line was stale text in my description, now corrected. The branch was 1 behind `main`, so I rebased it (now 0 behind) and regenerated the catalogs; only the `_updated` stamp moved. CI after the rebase: **1,090 passed, 9 skipped**, parity ✓, golden ✓ (109). The description's old "943 passed / 110 templates" numbers were from the pre-merge base and are replaced.

shnoodles (approved):
Approved at `50c3c115`. Sentence fixed and rewrapped; every factual claim verified (Stingray retired, Pilotfish present, `sm_crowding` unused, `risk_off_rotation` is hyena's, catalogs regenerate to only the date stamp). Rebased onto merged #610 with the stale stacking note corrected.

PR #642, opened 2026-09-15
senpi-signals 2.0.0: one on-demand market sweep, no compare, no cron (compare over periods is v2)

betashop:
## What
senpi-signals 2.0.0 ships as an MVP: **a one-time, on-demand market sweep with no compare.**

- **One process.** `scripts/sweep.py` does the whole gather (8 MCP reads, zero model tokens) and hands `current.json` to `scripts/score.py` in-process. `--print-feed` prints only the feed the agent presents, plus one plain line naming any source that failed.
- **No compare.** `score.py` runs without `--state`, so nothing is read from or written to history and back-to-back runs rank the same. Only single-reading detectors fire: `sm_divergence` (the standing state), `funding_extreme`, `momentum_event` and `cross_asset_laggard`.
- **No schedule, nothing to fund.** SKILL.md's body and description tell the agent to say no to a cron or any schedule, and never to deploy a strategy or fund a wallet to run it.

## Why
1.x ran the gather as an agent cron, a full model call per firing, and the firings timed out. Keeping signals continuous on the runtime would need a funded strategy wallet on every box, which users should not pay for. So 2.0 runs when asked. Compare over periods moves to v2, where history is kept once on Senpi's side: whale moves, OI and funding changes, the 12h cohort trend, and anti-repeat rotation.

## What moved to v2
- The `strategies/signals` runtime host, `references/scheduling.md`, the scanner tests and the history-file resolver (`_default_state_path`). The pre-rescope branch is saved as `feat/senpi-signals-v2-compare`.
- The compare engine stays in `score.py` behind `--state` for v2 development, and its tests still cover it. `score.HISTORY_DETECTORS` names the detectors 2.0 never fires: `oi_surge`, `sm_conviction`, `funding_flip`, `whale_move`, `sm_positioning_build`, `sm_flow`. `references/detectors.md` marks every detector 2.0 or v2.

## Fixed on this branch along the way (found in live agent runs)
- **Nested payloads.** The leaderboard board and momentum-event payloads are nested one level deeper than the fixtures recorded, on every live response. The sweep unwraps them, and the fixtures now use the real shape.
- **Feed-only output.** `--print-feed` exists because the agent kept turning the sweep's diagnostic lines into a footer for the user.
- **Content-only note.** A user's feed no longer carries "verify before posting", which belongs to the content feed.
- **Funding wording.** Funding is described as a percent of position size, never of margin.

## Verification
- **Signals tests:** `senpi-signals/tests` 35 passed. New tests cover:
  - a second sweep with BTC OI +25% fires no `oi_surge`, back-to-back sweeps rank the same, and only `current.json` and `signals.md` are written;
  - `score.py` without `--state` writes no history;
  - the skill says no to a schedule in both the body and the description, and names nothing to deploy or fund.
- **Mutation-checked:** making the sweep keep history, giving `score.py` a default state file, or removing the no-schedule rule from the body or the description each fails a test.
- **CI pytest:** 1,127 passed, 9 skipped, exit 0. Vendor parity OK; min-budget golden OK.

## Still to do
- One agent run of `sweep.py --print-feed` from this branch after the rescope, including a request for a schedule, which should get a no.
- Merge after #647, since both edit the CI line.
- After merge: the runtime skills-manifest entry.

Review: Ignas.


shnoodles:
**Review — the architecture is right and unusually well-guarded for a package that trades nothing. Three things: one false claim in the body, the tests, and a rebase.**

**Intent ✓, and the economics are the whole point.** ~$0.47 per firing × 23 firings, all timed out, replaced by a clock that costs nothing per tick. Moving a pure gather off the model and onto the runtime is the right call, and "same gather, same ranker, no model in the loop" is exactly the trade.

**The dangerous part of this design is a non-trading package living in `strategies/`, and it is properly fenced.** I checked each guard rather than taking the body's word:

| guard | verified |
|---|---|
| invisible to the picker | **absent from both** `strategies/catalog.json` and `senpi-strategy-discover/catalog.json` |
| the mechanism | `deployable: false` in `strategy.yaml:31`, commented *"hidden from the picker — a host, not a strategy"* |
| cannot open a position | `slots: 0`; `scan()` returns `[]` on every path, including both failure paths |
| author lint still passes | `validate_strategy.py` → **✓ signals v1.0.0** — the one OPEN_POSITION action lists the scanner, so it is not an orphan (#606's check) |
| the package name rule | no bare `@senpi/runtime` anywhere; `runtime.yaml` uses `@senpi-ai/runtime` |

The missing-skill path is also the right shape: `scan.py` fails soft with a stderr line naming `SENPI_SKILLS_DIR`, and the comment states that this makes `validate` report **UNPROVEN** — fail-closed at the gate rather than a package that deploys and silently writes nothing.

---

### 1. The min-budget golden claim is not true

> "min-budget golden regenerated with the one new entry (`signals: 10`)"

`senpi-trading-runtime/tests/fixtures/min_budget_golden.json` has **109 entries and no `signals`**, and this PR does not touch `tests/fixtures/` at all — `git diff origin/main...HEAD -- senpi-trading-runtime/tests/fixtures/` is empty.

The golden passes anyway, and correctly: `deployable: false` keeps the package out of the computed set, so there was nothing to regenerate. Harmless, but the body sends a reviewer looking for an entry that does not exist — and "golden regenerated" is exactly the sentence someone will trust instead of re-running it. Please just drop it.

### 2. Its 33 tests are outside CI

`test_score.py` and `test_sweep.py` live in `senpi-signals/**scripts**/`, not `senpi-signals/tests/`, and the workflow is untouched by this PR — so `senpi-signals` appears nowhere in the CI line. 1,514 lines of new scoring and sweep logic land with 33 tests that only ever run locally.

This is the third instance in this batch: #643's 20 new tests are in `strategies/*/tests` (37 such dirs, none in CI), and #647 fixes exactly this shape by adding its skill's `tests/` directory to the workflow. Cheapest fix here is to follow #647 — move the two files to `senpi-signals/tests/` and add that directory to the same line.

(For the record, since my #643 comment guessed wrong about why the `strategies/*` tree cannot be collected in bulk: the blocker is a module-level `sys.exit()` in copied `test_package_shape.py` / `test_engine.py` files — 8 files across `chimp`, `gibbon`, `gorilla`, `orangutan`.)

### 3. Rebase

Branch is **3 commits behind `main`**. No conflict, but the catalogs were regenerated against an older tree, and `gen_catalog.py --updated 2026-09-15` stamps a date that is now yesterday.

### Ponytail
On a +3,213 diff the honest ponytail verdict is that the diff *is* the deletion: an agent cron that re-derived the universe through the model every 45 minutes becomes a scanner that does it for free. Nothing to cut. The inert `exit:` block is required by the runtime and says so; `slots: 0` plus an action that never receives a signal is the minimum shape that satisfies both the runtime and the author lint.

### Verified locally
Branch @ pr-642: `validate_strategy.py` ✓ signals v1.0.0 · min-budget golden ✓ (109) · `test_score` + `test_sweep` + `test_scan_noop` **33 passed** · full CI **1,092 passed, 9 skipped, 0 failed**.

betashop:
All three done in 1195994.

**1. The golden claim — false, and worse than you found.** Confirmed: 109 entries, no `signals`, no fixture touched. The description also said `GOLDEN OK — 110 templates` in its verification block, and a later paragraph said the golden was *unchanged*, so it contradicted itself. Both false lines are removed, and a dated note at the bottom says what was wrong, so nobody goes looking for the entry.

**2. Tests in CI.** `test_score.py` and `test_sweep.py` moved to `senpi-signals/tests/` via `git mv`, repointed with `SCRIPTS = HERE.parent / "scripts"`: the `sys.path` insert, the two `subprocess` calls running `score.py`, and the `skill_scripts` monkeypatch. The directory is on the workflow's pytest line. Checked basenames first: `test_score.py`, `test_sweep.py` and `test_scan_noop.py` appear nowhere else on `main`, so there's no collision in the shared session. CI grew by exactly **+28**. Proof: forcing `score._num` to return `0.0` fails **17 signals tests inside the CI run**.

`test_scan_noop.py` stays in `strategies/signals/tests/`, where the #649 runner covers it.

**3. Rebase.** Was 3 behind, now 0; catalogs regenerated, date stamp only. After the rebase: CI **1,123 passed, 9 skipped**, golden ✓ (109), `validate_strategy.py` ✓ signals v1.0.0.

**Merge order:** #647 also appends a directory to the same pytest line, so whichever of the two lands second needs a one-line rebase.

betashop:
Scope change, pushed as 2a0773f. 2.0.0 is now an MVP: one on-demand reading with no compare.

- **What changed.** The sweep runs `score.py` without `--state`, so nothing is read from or written to history. Only single-reading detectors fire. `score.HISTORY_DETECTORS` names the v2 ones.
- **What came out.** The `strategies/signals` host, `references/scheduling.md`, the scanner tests and `_default_state_path`. The pre-rescope branch is saved as `feat/senpi-signals-v2-compare`.
- **The schedule rule.** SKILL.md says no to a cron or any schedule, in the body and in the description. A test pins both.
- **Checks.** Signals tests 35 passed, CI 1,127 passed, 9 skipped. Each new test fails when its rule is removed.

The description above is rewritten for the new scope. Your ready-to-merge on the ledger covered the old scope, so this needs a fresh look.

shnoodles:
**Not approving yet — the PR has been rescoped out from under my review, and that is the right call but it needs a fresh look.**

What I reviewed (`a183ad97`) was the runtime-scanner design: `strategies/signals` as a `deployable: false` host, `slots: 0`, a scanner that always returns `[]`, an inert exit block, and the catalog/golden questions that came with it. I verified those guards specifically — catalog absence, the author lint, the `@senpi-ai/runtime` rule.

At `2a0773f6` that package is gone: no `strategies/signals`, no runtime.yaml, no scanner, no host wallet. The PR is now an on-demand sweep with no compare and no schedule. **Every guard I checked no longer exists**, so my approval would be attesting to something that is not in the diff.

**The rescope itself reads right.** "Keeping signals continuous on the runtime would need a funded strategy wallet on every box, which users should not pay for" is a better answer than the one I reviewed — I checked that the host was fenced off from the picker, but I did not ask whether a $10 wallet per box was a reasonable thing to ask of users at all. It isn't, and shipping the read-only half first is the smaller, more honest version.

What a fresh review needs to cover, none of which my earlier one touched:

1. **`HISTORY_DETECTORS` as the 2.0/v2 line.** The compare engine stays in `score.py` behind `--state`. Worth confirming that no 2.0 path can reach a history detector, and that `--print-feed` cannot silently emit one — the split is only as good as the thing that enforces it.
2. **The nested-payload fix.** "The leaderboard board and momentum-event payloads are nested one level deeper than the fixtures recorded, on every live response" — that is a real bug found in live runs, and the fixtures changed shape to match. Worth verifying the new fixtures against a live response rather than against each other.
3. **The tests moved to `senpi-signals/tests/` and the CI line** — that part I did verify and it held; presumably unchanged.
4. **`test_signals_skill_text.py` (new)** and the "say no to a cron, never deploy a strategy to run it" rules — the guard that keeps the deferred design from being re-suggested by an agent.

Happy to do that pass whenever you want it. Flagging rather than rubber-stamping because the two diffs share a number and very little else.

betashop:
Thanks. Here is what your fresh pass asked for, checked against live data and main. Pushed `5f65878`.

**1. The 2.0/v2 line.** The history engine runs only when `score.py` gets `--state`. `sweep.rank()` never passes it, and SKILL.md tells agents to run `sweep.py --print-feed` and nothing else. Two tests hold the line, and I mutation-proved both by making `rank()` pass `--state`:
- `test_a_sweep_is_one_reading_so_a_second_sweep_compares_nothing` runs `run()` in-process. Under the mutation, the second sweep fires `oi_surge`.
- The new `test_print_feed_never_carries_a_history_detector` sends the same two readings through `sweep.main([..., "--print-feed"])`. It asserts that the printed result has no `HISTORY_DETECTORS` item, that `diff_baseline_ts` is `None`, and that only `current.json` and `signals.md` are on disk. Under the mutation it fails with `{'cross_asset_laggard', 'oi_surge', 'sm_divergence'}`.

**2. Nested payloads, checked against live data.**
- MCP main (`7188e52`): `src/tools/leaderboard.tools.ts:286` returns `createResponse(tool, 'read', { markets })` and `:244` returns `{ events }`. `leaderboard.tools.comprehensive.test.ts:751` asserts `result.data?.markets.source_trader_count`.
- Production agent calls in the last 24h: all 44 successful `leaderboard_get_markets` results put rows at `data.markets.markets[]` (0 flat), and all 44 carry `is_dominant_direction`, `trader_count`, `token_price_change_pct_4h` and `pct_of_top_traders_gain`. All 24 successful `leaderboard_get_momentum_events` results put rows at `data.events.events[]`, with `decision`, `top_positions` and `detected_at`.
- Telemetry cuts each result off at about 8,000 characters, so the board's tail (`source_trader_count`) is confirmed from the MCP source above, not from telemetry.
- Live `top_positions` is `null` on blocked events. `momentum_events()` already reads `(e.get("top_positions") or [])`.

**3. Tests moved and the CI line.** Unchanged since your check.

**4. `test_signals_skill_text.py`.** Four guards on the SKILL text: `test_the_run_is_one_on_demand_reading`, `test_a_request_for_a_schedule_gets_a_no`, `test_nothing_to_deploy_fund_or_keep`, `test_compare_detectors_are_v2_and_never_claimed`.

Also in this push, a legend fix. The legend said 🟡 meant 45–64, but news items show from `MIN_SOCIAL` (30) and badge 🟡 below 65, so a news item scored 32 looked mislabelled. The legend now says "under 65" in `score.py`, SKILL.md and `detectors.md`. A new test checks that the 🟡 band never starts above `MIN_SOCIAL`, and it fails on the old text.

CI run locally: vendor parity OK, golden OK (109), pytest 1,129 passed / 9 skipped.

shnoodles (approved):
**Approved at `5f65878`.** Fresh pass done — I verified each of the four points independently rather than reading the reply, and the rescope holds up.

**1. The 2.0 / v2 line — I ran your mutation myself.** Made `rank()` append `--state`:

```
3 failed, 34 passed
  test_a_sweep_is_one_reading_so_a_second_sweep_compares_nothing
  test_print_feed_never_carries_a_history_detector
  test_current_json_carries_the_full_metric_set
```

Restored → 37 pass. So the split is enforced by the tests, not just by convention, and a third test catches it too. This was the point I cared most about: `score.py` still contains the whole compare engine, so "2.0 never fires a history detector" is a claim that needed a guard rather than a docstring. It has one.

**2. Nested payloads — confirmed on live data.** Sampled real `leaderboard_get_markets` results from the last 24h; the shape is exactly as you describe:

```json
"data": { "markets": { "markets": [ { "token": …, "direction": "long",
          "pct_of_top_traders_gain": 17.31, … } ] } }
```

I could not reproduce your 44/44 count — telemetry double-escapes the JSON and my `LIKE` patterns kept returning 0 against the escaped form — so I confirmed the envelope on samples plus the MCP source you cite rather than by count. The shape claim is right either way, and the same nesting is what #653 and #655 document in the phalanx engine, so three PRs now agree on it.

**3 & 4.** Tests are in `senpi-signals/tests/` and on the CI line — unchanged since I checked. The four skill-text guards are present.

**The rescope is clean.** No `strategies/signals` package, and the only catalog change is the `_updated` stamp — `signals` is **not** a catalog entry (109 entries, unchanged). So the picker cannot surface it and there is no host wallet to fund, which was the whole point.

**Verified locally:** vendor parity ✓, min-budget golden ✓ (109), full CI **1,129 passed / 9 skipped / 0 failed** — matches your number exactly.

---

Two merge notes, neither blocking:

- Branch is **13 behind main**, and the conflict is exactly the one file you said: `.github/workflows/tests.yml`.
- That file is now contested by **five open PRs** — #642, #645, #647, #649, #652. Tested pairwise: 647→649 conflicts, 647→652 conflicts, 649→652 clean. Worth landing them in a chosen order rather than discovering it; whichever goes first, the rest need the same one-line rebase.

Good call on the rescope, for what it is worth. I checked that the runtime host was fenced off from the picker and never asked whether a $10 wallet per box was a reasonable thing to ask of users at all — shipping the read-only half first is the better answer.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,455 @@
...
+    """Everything score.py needs, from one client: {"asset_metrics", "events", "coverage", "reads"}."""
+    now = now or datetime.datetime.now(datetime.timezone.utc)
+    c, cov = Client(call_tool), {}
+    metrics = universe(c, cov, top_n)
+    if metrics:
+        cohort(c, cov, metrics)
+        src = board(c, cov, metrics)
+        events = momentum_events(c, cov, metrics, now) + cross_asset(c, cov, metrics)
+    else:
+        src, events = None, []
```
0xsarvesh, line 369: **S1 · reliability — `market_list_instruments` is an unguarded single point of failure**

When the universe read fails, `gather()` skips the cohort, board, momentum and cross-asset reads entirely. Combined with the fact that there are **no retries anywhere** in the skill (`grep -rn "retry|retries|backoff" scripts/*.py` → nothing), one transient 503 produces a zero-information run that still exits 0.

**Failure scenario.** `market_list_instruments` 503s. `reads=1`. The user taps the Signals chip and gets:

```
_Nothing notable stands out right now — a quiet read is a correct answer._
_Not measured this run: market data._
```

SKILL.md trains the agent that a quiet read is a *correct answer* and to add nothing — so the agent reports a calm market during a total data outage. I ran this with the transport hard-down and with a bad token; **both return EXIT CODE 0.** A healthy-but-quiet market and a complete outage are near-indistinguishable in both text and exit status.

**Fix:** retry this read (one idempotent call), return non-zero when the universe is dark, and suppress the "quiet market" line in that state. A dark universe should be an error, not a feed.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,455 @@
...
+
+def run(call_tool, out_dir=None, now=None, top_n=UNIVERSE_TOP_N, top=None, lens="both"):
+    """One sweep: one reading, no compare. Writes current.json + signals.md and returns
+    {"summary", "reads", "coverage", "current", "result", "out_dir"}."""
+    out_dir = out_dir or default_out_dir()
+    os.makedirs(out_dir, exist_ok=True)
+    cur = gather(call_tool, top_n=top_n, now=score._parse_ts(now) if now else None)
+    current_path = os.path.join(out_dir, "current.json")
+    with open(current_path, "w") as f:
+        json.dump(cur, f)
```
0xsarvesh: **S1 · reliability — shared, unlocked `out_dir`: one run can present another run's data**

`run()` writes `current.json` → `rank()` re-reads it from disk → `score.py` writes `signals.md` → `feed_text()` reads it back (L396). Nothing is atomic, nothing is locked. Both callers — the chip's `--print-feed` and market-pulse's `--brief 3` (added in #657) — use the same default dir.

I forced the interleave to prove the window is real. Run B's universe carries an asset run A never gathered:

```
Run A's universe never contained ONLYINB.
ONLYINB appears in the feed run A PRESENTED TO ITS USER: True
   >> ⭐ 🟡 **44.5** · `ONLYINB` — funding_extreme
```

**Failure scenario.** A user asks for a market pulse; while its sweep is mid-flight they tap the Signals chip. The runs cross on disk and one leads its feed with an asset it never read — golden rule 1 broken by a filesystem race, silently, with no error.

#642 alone has one caller, so this is latent here. **#657 adds the second caller and makes it reachable.**

**Fix:** pass the gathered dict to `score.py` in-process (see my note on L362) — the disk round-trip isn't needed. If the file stays, make it per-run (PID/uuid-suffixed).

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,455 @@
...
+    argv = [current_path, "--out", os.path.join(out_dir, "signals.md"), "--lens", lens]
+    if now:
+        argv += ["--now", now]
+    if top:
+        argv += ["--top", str(top)]
+    saved, buf = sys.argv, io.StringIO()
+    try:
+        sys.argv = ["score.py"] + argv
+        with contextlib.redirect_stdout(buf):
+            score.main()
```
0xsarvesh, line 387: **S2 · maintainability — CLI-as-API, and the disk round-trip it forces**

Driving `score.main()` by monkey-patching `sys.argv` and redirecting stdout, then parsing JSON back out, is the single load-bearing joint between the two halves of the system. It's non-reentrant, hostile to testing, and it is **the reason `current.json` has to round-trip through disk at all**: `score.py`'s public API is a file path, so an in-memory dict gets serialized to hand it to a function in the same process.

Worth noting `current.json` isn't serving as a snapshot in 2.0 — `--state` is never passed, nothing is read back, it's overwritten every run, and SKILL.md tells the agent not to open it. It's a pure IPC buffer with a debugging side-benefit.

**Fix:** give `score.py` a real function entry point and keep `main()` as a thin CLI wrapper. That removes the monkey-patch, the disk round-trip, **and** the S1 race on L383 — one change, three findings.

senpi-signals/scripts/score.py
```diff
@@ -0,0 +1,1055 @@
...
+
+
+def trade_read(s):
+    """Actionable framing (trade feed) — names the side and whether price is confirming it."""
+    a, d = s["asset"], (s.get("direction") or "")
+    nums = "; ".join(s.get("numbers") or [])
+    det = s["detector"]
+    early = earliness(s)
+    tag = ("price hasn't moved yet — early" if early >= 0.8
+           else ("price already ran — late" if early >= 0.45 else "price is going against it"))
```
0xsarvesh: **S1 · correctness — a missing 4h price renders as "price already ran — late"**

`earliness()` (L278) returns **0.50** when direction is set but `price_change_pct` is `None`. This band treats `>= 0.45` as "price already ran — late", so *no price data at all* prints as a positive claim that price has moved.

```
dir=short  price=None    earliness=0.50  ->  ... price already ran — late.
dir=short  price=-2.0    earliness=0.67  ->  ... price already ran — late.
dir=short  price=0.0     earliness=1.00  ->  ... price hasn't moved yet — early.
```

**This fires on the healthy path, not just in degraded mode.** `price_change_pct` is only populated for names on the 4h board (sweep.py L251-253). In this PR's own fixture, 2 of 4 universe names have no 4h price:

```
BTC        price_change_pct(4h)=0.4     crowd_source=board_4h
ETH        price_change_pct(4h)=None    crowd_source=funding_sign
SOL        price_change_pct(4h)=None    crowd_source=funding_sign
xyz:TSLA   price_change_pct(4h)=-1.2    crowd_source=board_4h
```

and the healthy feed says:

> 🟡 **61.6** · `ETH` — Smart money long vs the crowd on ETH, **price already ran — late.**

There is no price behind that sentence. It breaks golden rule 1, and `HOW_TO_READ` explicitly teaches the reader that "early" means the move hasn't happened — so "late" reads as its verified converse. It also mis-ranks: earliness is 22% of `trade_score`, and *unknown* currently scores above *contradicted* (0.15).

**Fix:** give unknown price its own band and wording ("price unknown this run"), and don't let it earn the confirmed-signal score.

0xsarvesh (commented):
## Review — senpi-signals 2.0.0

Checked out the branch and ran its own suite: **69 passed, 0 failed.** Everything below is reproduced against that tree using this PR's own fixture (`tests/test_sweep.py::fake_call_tool`), not inferred from reading.

The architecture is sound and the epistemic design is genuinely strong — the circularity analysis, headcount-over-notional, `sample_shrink`, `MTM_SAFE_PCT`, and replacing `confirmation` with `earliness` are real reasoning, not decoration. Most products in this space ship exactly the bug golden rule 11 exists to prevent. The coverage model (a dark source is never "nothing happening") is the right instinct — several findings below are cases where the *implementation* leaks past that principle, which is why they're worth fixing rather than debating.

My findings concentrate in **reliability under partial failure** and **maintainability of the two-script seam**. Inline, by severity:

| # | Sev | Where | Finding |
|---|---|---|---|
| 1 | **S1** | `sweep.py:344` | `market_list_instruments` is an unguarded SPOF; no retries anywhere; total outage exits 0 as "nothing notable stands out" |
| 2 | **S1** | `sweep.py:383` | Shared unlocked `out_dir` — one run presents another run's data (reproduced) |
| 3 | **S1** | `score.py:711` | Missing 4h price renders as "price already ran — late" — fires on the healthy path |
| 4 | **S2** | `score.py:864` | ⛔ engine banner printed verbatim to users; false in the universe-dead state |
| 5 | **S2** | `sweep.py:167` | "the crowd" has two evidential bases, never attributed |
| 6 | **S2** | `sweep.py:360` | CLI-as-API seam; forces the disk round-trip that causes #2 |
| 7 | **S2** | `score.py:925` | Only lock in the codebase is in the half that never runs; ~500 lines unreachable |
| 8 | **S3** | `test_score.py:44` | Test pins the earliness collision rather than catching it |
| 9 | **S4** | `sweep.py:290`, `:162` | `tier 2 Tier 2`; dead `price_change_pct_24h` field |

**S1 = fix before merge · S2 = fix before it's load-bearing · S3 = should fix · S4 = nit**

Two notes on blast radius, since the launch thread merges this first and #657 second:

- Findings 1 and 3 are in the default user path the moment the runtime manifest lands. Finding 2 is latent in this PR (one caller) and becomes reachable when #657 puts a sweep inside every market pulse.
- Rollback is fix-forward only — agents never downgrade — so each of these is live fleet-wide until a version bump propagates (~60 min). I'd fix the S1s here rather than ship and bump.

Leaving this as a comment rather than a blocking review; happy to re-review once the S1s are addressed.

senpi-signals/scripts/score.py
```diff
@@ -0,0 +1,1055 @@
...
+    note = "_Observation, not advice. Every number is from a live read this run"
+    note += " — verify before posting._" if consumer == "social" else "._"
+    out = [f"# 🔭 Senpi Signals — {ts} UTC", "", note, ""] + HOW_TO_READ + ["", "---", ""]
+    cov = cov or {}
+    if cov.get("smart_money_lens") == "NO DATA" and cov.get("flow_lens") == "NO DATA":
+        out += ["> ⛔ **Smart-money lens UNAVAILABLE this run — no cohort positioning was supplied.**",
+                "> Everything below is OI, funding and price: the same inputs as market-pulse. The",
+                "> divergence and flow detectors did not find nothing — they were never fed. Run the",
+                "> proven-cohort fan-out (`senpi-smart-money`) and re-run before reading this as a",
+                "> smart-money report.", ""]
```
0xsarvesh, line 917: **S2 · correctness — this banner is printed verbatim to end users, and is false in one state**

`--print-feed` prints `signals.md` verbatim, so this block reaches the user, telling them to "run the proven-cohort fan-out (`senpi-smart-money`) and re-run".

That's exactly what SKILL.md's own output conventions forbid: *"Talk about the market, never about the engine… Never narrate coverage lines, read counts, detector names."* `not_measured()` already **is** the one plain clause the conventions allow, so the user gets both — a five-line engine-internals banner naming another skill, plus the clean one-liner.

In the universe-dead case the banner is also **factually wrong**: it asserts "Everything below is OI, funding and price" when nothing is below and those weren't read either.

Note the inversion: the machine-readable `[warn]` on stderr **is** suppressed in the user path (sweep.py L445-451), while the prose engine-talk is shown.

**Fix:** gate on `consumer == "social"`, same as the posting reminder two lines above.

senpi-signals/scripts/score.py
```diff
@@ -0,0 +1,1055 @@
...
+
+
+def _commit_state(path, cur_metrics, now, consumer, social_picks):
+    # `social_picks` is every signal that was SHOWN this run (both feeds). It used to be the social
+    # feed alone, which was correct while the feeds overlapped — a trade pick was almost always also
+    # a social pick, so it got recorded anyway. Once dedupe_feeds started promoting items into the
+    # trade feed and out of news, those items stopped being marked as surfaced and would come back
+    # looking fresh forever. Mark what the reader actually saw.
+    """Locked read-merge-write: re-read under an exclusive lock and MERGE this snapshot into whatever
+    ring is current, so a concurrent cron/user run never clobbers the other's baseline; update only
```
0xsarvesh, line 974: **S2 · maintainability — the only concurrency protection in the codebase is in the half that doesn't run**

This docstring promises *"a concurrent cron/user run never clobbers the other's baseline"*, and `_commit_state` does take an `flock`. But it's only reachable via `--state`, which `sweep.py` never passes — so in 2.0 this function never executes, while the path that **does** run (`run()` → `current.json` → `signals.md`, sweep.py L383) has no locking at all.

More broadly: ~500 lines of `score.py` (the ring, `_pick_baseline`, `make_slow_lookup`, whale moves, OI surge, funding flip, `sm_flow`, freshness, locked commits) are unreachable from the skill's own entrypoint, with `HISTORY_DETECTORS` as a comment-enforced boundary rather than a structural one.

**Suggestion:** split into `score.py` (what 2.0 runs) + `history.py` (v2), so the shipped surface is legible and the dead half can't accrue false confidence like this docstring does.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,455 @@
...
+            "dex": _dex(name),
+            "notional_vol": vol,
+            "oi": _num(ctx.get("openInterest")),           # BASE units — immune to the price move
+            "price": price,
+            "price_change_pct_24h": (round((price - prev) / prev * 100, 3)
+                                     if price is not None and prev else None),
+            "funding_annualized_pct": (round(fund * HOURS_PER_YEAR * 100, 3) if fund is not None else None),
+            # the funding-sign proxy (positive ⇒ longs pay ⇒ crowd LONG); the 4h board overrides it
+            "crowd_dir": (None if fund in (None, 0) else ("long" if fund > 0 else "short")),
+            "crowd_source": "funding_sign",
```
0xsarvesh, line 190: **S2 · correctness — "the crowd" means two different things and the output never says which**

`crowd_dir` is set here from the **sign of the funding rate**, and at L255 from the 4h board. `crowd_source` records which — and then never reaches any renderer:

```
$ grep -n "crowd_source" scripts/score.py
→ (no matches)
```

**Failure scenario.** "Smart money short vs the crowd on BTC" is a statement about real trader positioning when the board fired, and a statement about a funding-rate sign when it didn't. On this PR's fixture, half the names are on the funding proxy. When the board read *fails*, every name silently switches evidential basis with no change in wording.

Golden rule 8 requires naming which board and window. **Fix:** thread `crowd_source` into `trade_read()`/`frame()` and label it.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,455 @@
...
+        mk = str(p.get("market") or "")
+        asset = mk if mk in metrics else (f"xyz:{mk}" if f"xyz:{mk}" in metrics else None)
+        if asset is None:
+            continue
+        m = metrics[asset]
+        dp, wallet = _num(e.get("delta_pnl")) or 0.0, str(e.get("trader_id") or "")
+        out.append({
+            "asset": asset, "dex": m["dex"], "detector": "momentum_event",
+            "direction": str(p.get("direction") or "").lower() or None,
+            "numbers": [f"tier {e.get('tier')} {e.get('tier_label') or ''}".strip()
```
0xsarvesh, line 313: **S4 · nit** — `tier` and `tier_label` are concatenated, but the label already contains the tier, so this renders as:

> `tier 2 Tier 2: a top trader's 4h PnL moved $3,000,000`

Reaches the user verbatim in the trade feed.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,455 @@
...
+            continue
+        price, prev = _num(ctx.get("markPx")), _num(ctx.get("prevDayPx"))
+        fund = _num(ctx.get("funding"))
+        keep.append((vol, name, {
+            "dex": _dex(name),
+            "notional_vol": vol,
+            "oi": _num(ctx.get("openInterest")),           # BASE units — immune to the price move
+            "price": price,
+            "price_change_pct_24h": (round((price - prev) / prev * 100, 3)
+                                     if price is not None and prev else None),
```
0xsarvesh, line 186: **S4 · nit** — `price_change_pct_24h` is computed here, documented in SKILL.md and `references/detectors.md` as a declared per-asset output, and consumed by nothing (no detector or renderer reads it). Either wire it up or drop it from both the code and the docs.

senpi-signals/tests/test_score.py
```diff
@@ -0,0 +1,611 @@
...
+    assert 0.45 < mid < 1.0                                # ramps monotonically between floor and full
+
+
+def test_earliness_rewards_flow_before_the_move_not_after():
+    """Replaces the old `confirmation` term, which paid a signal for price having already moved its
+    way. That double-counts one move — a "who gained over 4h" read IS a statement that price moved —
+    and it systematically preferred LATE signals. Early (flow in, price flat) is the valuable state."""
+    e = score.earliness
+    assert e({"direction": "short", "price_change_pct": 0.0}) == 1.0     # flat = early = the alpha
+    assert e({"direction": "short", "price_change_pct": -3.0}) == 0.5    # already ran = real but late
```
0xsarvesh, line 44: **S3 · test quality — this pins the collision rather than catching it**

`0.5` here is asserted as "already ran = real but late". The problem is that a **missing** price also returns exactly `0.5` (score.py L278 returns 0.50 when `price_change_pct is None`), so the two cases are indistinguishable and both render as "price already ran — late" — see my comment on score.py L711.

Once unknown-price gets its own band, this test should assert the two are *different* values, which is the guard that would have caught it.

betashop:
## S1s fixed — `2ff1adc7`, rebased onto main

Reproduced all three against your repro before changing anything; you were right on each. Full CI **1224 passed / 9 skipped** on this branch, **1246 / 9** on #657 with it.

**1 · Missing price claimed the price moved** (`score.py:713`)

`earliness()` returned `0.50` for "side known, no price read" — byte-identical to "price already ran with it". Unread now has its own value and its own sentence:

```
dir=short  price=None   earliness=0.35  ->  ... price not read this run.
dir=short  price=-2.0   earliness=0.67  ->  ... price already ran — late.
dir=short  price=0.0    earliness=1.00  ->  ... price hasn't moved yet — early.
dir=short  price=3.0    earliness=0.15  ->  ... price is going against it.
dir=None   price=None   earliness=0.30  ->  ... no side resolved.
```

`EARLINESS_PRICE_UNREAD = 0.35` — below a confirmed read, above a contradicted one, so it can't earn the confirmed score. While rewriting the bands I found the twin of your bug in the same table: a directionless signal scores `0.30`, which fell under the `>= 0.45` cut into **"price is going against it"** — asserting a direction it doesn't have. `trade_read`'s three bands are now five explicit states, so every branch is one we measured.

Your finding **#8** is folded in: the earliness test asserts the two states *differ* instead of pinning their collision.

**2 · The shared `out_dir` race** (`sweep.py:385`)

Took your L362 suggestion, in the direction that doesn't need the refactor: each run hands `score.py` files only it can see (`.current.<pid>.<uuid>.json`), then publishes under the stable names with `os.replace`. The rendered feed rides back in the result as `feed_md`, so nothing re-reads a file it just wrote.

The guard is your interleave, made deterministic — run B starts and finishes between A's write and A's read:

```
ONLYINB in B's asset_metrics:     True
ONLYINB in A's metrics/result/feed: False / False / False
```

Restore a shared filename and it fails. `current.json` and `signals.md` still land in the stable place for debugging; nothing reads them back.

**3 · The universe SPOF** (`sweep.py:346`)

`READ_ATTEMPTS = 2` — every read here is an idempotent GET. And a dark universe is no longer a feed: plain outage wording instead of *"a quiet read is a correct answer"*, and **exit 3** on both the `--print-feed` and JSON paths.

One extension on #657: `brief_text` closes **every** market pulse, so it carries the same guard — otherwise an outage prints as a quiet market inside someone else's answer.

**Five guards, each proven to fail against the previous behaviour:** `unread == 0.50` · the 3-band renderer · `READ_ATTEMPTS = 1` · `universe_is_dark` disabled · a shared IPC filename.

**Not done, deliberately** — happy to take any of these now or as follow-ups: **#4** (the `⛔` banner gated on `consumer == "social"`), **#5** (`crowd_source` threaded into the renderer — the one I most want, since "the crowd" silently swaps evidential basis), **#6/#7** (the real `score.py` entry point and the `score.py`/`history.py` split — #2's fix removed the race but left the seam), **#9** (`tier 2 Tier 2`, dead `price_change_pct_24h`).

Also rebased onto main — the `tests.yml` conflict you and Ignas both flagged resolved clean, `senpi-signals/tests` is still on the CI line.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,509 @@
...
+
+
+def _traders_of(data):
+    """The row list of a read. The leaderboard tools wrap their client's payload under the same key
+    it already uses — `leaderboard_get_markets` returns data.markets.markets[] and
+    `leaderboard_get_momentum_events` data.events.events[] — so one same-key level is unwrapped."""
+    if isinstance(data, list):
+        return [t for t in data if isinstance(t, dict)]
+    if isinstance(data, dict):
+        for k in ("traders", "markets", "events", "instruments", "data", "results"):
```
0xsarvesh, line 136: **S3 · reliability — payload shape-sniffing degrades silently**

`_traders_of` guesses the row list by trying six candidate keys and then double-unwrapping a same-named nested key. When upstream changes shape, it returns `[]` — which every caller reads as "no data", not "I couldn't parse this".

**Failure scenario.** `leaderboard_get_markets` renames `markets` or stops double-wrapping. `board()` sees zero rows, sets `NO DATA: leaderboard_get_markets returned no markets`, and the sweep carries on. Every asset silently falls back to the `funding_sign` crowd proxy (L167) — a different evidential basis with identical wording in the feed — and nothing in the output distinguishes "the board is empty" from "we can no longer read the board".

The comment on this function documents the current MCP wrapping precisely, which is good; the risk is that the code can't tell a *correct* empty result from an unparseable one.

**Suggestion:** when the payload is a non-empty dict but no candidate key matches, treat it as a parse failure and record it distinctly in `coverage` rather than returning `[]`. A contract drift should look like an error, not like a quiet market.

senpi-signals/scripts/score.py
```diff
@@ -0,0 +1,1080 @@
...
+    if not isinstance(e, dict) or not e.get("asset") or not e.get("detector"):
+        return None
+    det = e["detector"]
+    if det == "whale_move":
+        # MOVES, not holdings — and a P&L SWING IS NOT A MOVE.
+        # `pnl_swing_usd` used to qualify here, which readmitted holdings through the back door: a
+        # months-old position printing hard because price moved is precisely the case SKILL.md warns
+        # about ("a months-old $38.68 entry is holdings, not a move"). Price acting on a static
+        # position is not a decision by the whale. Only a SIZE change is: opened, added, flipped.
+        chg = _num(e.get("change_usd"))
```
0xsarvesh, line 619: **S3 · correctness — `momentum_event` has no staleness gate**

`whale_move` is correctly gated at `WHALE_MOVE_MAX_AGE_MIN` (12h) with a clear rationale: an undated or old change is a holding, not news. No equivalent gate exists on this branch, so a `momentum_event` of any age is kept:

```
momentum_event  age=30 min  -> KEPT
momentum_event  age=9 days  -> KEPT
whale_move      age=30 min  -> KEPT
whale_move      age=9 days  -> DROPPED
```

A 9-day-old event renders as `FOO: x. [216h ago]` and is eligible for the feed — potentially as the ⭐ top item, since `magnitude` is driven by `delta_pnl` and nothing decays it with age.

To be fair on severity: `when()` does print `[216h ago]`, so a careful reader sees it, and `leaderboard_get_momentum_events` most likely only returns recent events in practice — which is exactly the point. The freshness of this feed currently rests on an **undocumented upstream guarantee** rather than on anything in this code. `sweep.py:275` filters on `decision == "sent"` but not on `detected_at`.

**Suggestion:** apply an explicit max age to momentum events too, or decay `magnitude` with age.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,509 @@
...
+            resp = self.call_tool(tool, kw)
+        except Exception:
+            self.failed += 1
+            raise
+        if tool == "discovery_get_trader_state":
+            self.trader_states.extend(_traders_of(_ok(resp)))
+        elif tool == "discovery_get_top_traders":
+            self.top_traders.extend(_traders_of(_ok(resp)))
+        return resp
+
```
0xsarvesh: **S4 · maintainability — this branch is dead with the real client**

`_ok` returns `None` when the payload carries `success: false`, and `_read` turns that into `coverage[key] = "failed: <tool>: tool reported success=false"` (L138).

But `mcp_client._unwrap` already **raises `MCPError`** on `success: false` — deliberately, with the comment *"silence here is how a missing/renamed tool masquerades as 'no data'"*. So with `MCPClient` the exception path always wins and this branch never executes; that coverage string is unreachable in production and only appears for a test double that returns the envelope raw.

Harmless today, but it's two error models for one condition, and the dead one is the more forgiving. Worth collapsing to one so the next person doesn't reason from the wrong branch.

0xsarvesh (commented):
### Follow-up: five smaller findings from the same pass

Second pass over my own notes — these didn't make the first review but are worth recording. All **S3/S4**; none of them should hold the PR on their own, and they're separate from the S1s above.

| # | Sev | Where | Finding |
|---|---|---|---|
| 10 | S3 | `sweep.py:107` | `_traders_of` shape-sniffing returns `[]` on contract drift — unparseable looks identical to empty |
| 11 | S3 | `score.py:600` | `momentum_event` has no staleness gate while `whale_move` does; a 9-day-old event is still feed-eligible |
| 12 | S4 | `sweep.py:99` | `success: false` handled twice, and the `_ok` branch is dead because `mcp_client` raises first |
| 13 | S4 | `sweep.py:224` | Tied cohort resolves to LONG arbitrarily instead of "side unresolved" |
| 14 | S4 | `sweep.py:320` | Upstream strings interpolated unescaped into text the agent presents verbatim |

The through-line with #10 and #11 is the same one as the S1s: the coverage model is a genuinely good idea, and the gaps are all places where *"we could not read this"* and *"there is nothing here"* collapse into the same output. That distinction is the skill's main epistemic claim, so it's worth being strict about.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,509 @@
...
+                                       or "; ".join(meta["warnings"]) or "empty cohort read")
+        return
+    smart_set = set(smart)
+    for t in c.top_traders:          # the ranking rows the cohort was cut from carry lifetime realized PnL
+        w = str(t.get("address") or t.get("trader_address") or t.get("wallet") or "").lower()
+        if w in smart_set and w not in c.wallet_pnl:
+            c.wallet_pnl[w] = round(sm._realized(t), 2)
+    positions = {}                    # coin → {wallet: signed BASE size}
+    for t in c.trader_states:
+        w = str(t.get("address") or t.get("traderAddress") or "").lower()
```
0xsarvesh, line 230: **S4 · correctness — a tied cohort resolves to LONG arbitrarily**

`ln >= sn` means an exact tie is reported as `smart_dir: "long"` with no indication the split was even. Golden rule 7 has a "side unresolved" escape hatch for precisely this and it isn't used here.

Low impact in practice — on a tie `smart_share` is `max(ln,sn)/len(smart)`, which will usually sit under `SMART_COHORT_PCT_MIN` (8%) and never fire `sm_divergence` — so this is more latent than live. But it's reachable on a small cohort, and the failure is invisible: a coin-flip direction presented with the same confidence as a 429-vs-40 rout.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,509 @@
...
+        cov["momentum"] = f"ok ({len(out)} sent events on universe names of {len(rows)} returned)"
+    return out
+
+
+def cross_asset(c, cov, metrics):
+    data = _read(c, "market_get_cross_asset_flows", {}, cov, "cross_asset")
+    if not isinstance(data, dict):
+        cov.setdefault("cross_asset", "NO DATA: market_get_cross_asset_flows returned no payload")
+        return []
+    lead = data.get("leader") if isinstance(data.get("leader"), dict) else {}
```
0xsarvesh, line 332: **S4 · robustness — upstream strings are interpolated into text the agent presents verbatim**

`lead.get('asset')` here, and `tier_label` at L290, are interpolated straight into `numbers`, which flows into `signals.md`, which `--print-feed` hands to the model, which presents it to the user.

Low risk because it's Senpi's own MCP, so I'd not hold the PR for it. But it's worth being deliberate: this is a path where upstream field content becomes part of an LLM turn and then user-facing copy with no escaping or length bound. A garbage or overlong `tier_label` becomes malformed markdown in the feed at minimum.

**Suggestion:** clamp length and strip markdown control characters on any upstream string that reaches `numbers`.

betashop:
## Finding 5 fixed too

`crowd_source` now reaches both renderers:

```
Smart money short vs the crowd on ETH (crowd per the 4h board), …
Smart money short vs the crowd on ETH (crowd per the funding sign), …
Smart money short vs the crowd on ETH (crowd per an unstated basis), …
```

`CROWD_SOURCE_LABEL` mirrors `SMART_SOURCE_LABEL`'s contract, `None` row included — it refuses to borrow the stronger basis rather than going quiet. The label lands in the trade read **and** in `numbers`, since the news lens prints numbers verbatim and one detector is rendered by both lenses. The field rides on the `whale_move` shape too, whose `conflict` flag is itself derived from the crowd.

Two guards, both proven to fail with the labels reverted. Full CI **1226** on #642, **1248** on #657.

That is S1×3 + finding 5 + finding 8. Still open and yours to call: **4** (banner gated on `consumer == "social"`), **6/7** (the real `score.py` entry point and the `score.py`/`history.py` split), **9** (`tier 2 Tier 2`, dead `price_change_pct_24h`). Happy to take any of them now — none is in the user-visible correctness class, so I stopped here rather than keep growing the PR under review.

senpi-signals/scripts/sweep.py
```diff
@@ -0,0 +1,509 @@
...
+    try:
+        sm = _smartmoney()
+    except ImportError as e:
+        cov["cohort"] = f"unavailable: {e}"
+        return
+    sm.MAX_PAGES = COHORT_PAGES
+    meta = {"warnings": []}
+    smart, _crowd = sm.build_cohorts(c, meta)
+    per = sm.cohort_bias(c, smart, meta, "smart") if smart else {}
+    if not smart or not per:
```
0xsarvesh, line 219: **S2 · reliability — the retry covers 4 of 8 reads, and the cohort's 4 are the unprotected ones**

`_read()` retries, and it's the right guard. But `build_cohorts` / `cohort_bias` call `client.mcp_call` directly through the vendored engine, bypassing `_read()` entirely — so the four `discovery_*` reads get no retry. Measured by counting attempts per tool with each one failing in turn:

```
market_list_instruments          attempts = 2  -> RETRIED
leaderboard_get_markets          attempts = 2  -> RETRIED
discovery_get_top_traders        attempts = 1  -> NO RETRY
discovery_get_trader_state       attempts = 1  -> NO RETRY
```

**Failure scenario.** A single transient 503 on `discovery_get_top_traders` still silently downgrades the run from a smart-money read to "market-pulse with extra steps" — no retry, no second chance. That's the exact failure the S1 fix was written for, and it survives on the lens the whole skill is named after: the cohort read is what fires `sm_divergence`, what `coverage` reports as `NO DATA`, and what triggers the ⛔ banner.

The asymmetry is an artifact of vendoring rather than a decision — the engine was copied in as-is, so it never learned about `_read`. Nothing here is wrong; it's just that the least-protected path is the most valuable one.

**Fix:** pass a retrying callable into the `Client` so `mcp_call` itself retries, rather than retrying one layer up in `_read`. That covers all 8 reads and keeps the vendored copy byte-identical (which `test_vendored_parity.py` requires anyway).

**Related, worth a number:** `READ_ATTEMPTS = 2` with no overall deadline moves the worst case from ~152s to ~200s. Reliability up, tail up. #657 puts that tail in every market pulse, so a wall-clock budget that returns whatever landed is worth pairing with this.

senpi-signals/scripts/score.py
```diff
@@ -0,0 +1,1101 @@
...
+    note = "_Observation, not advice. Every number is from a live read this run"
+    note += " — verify before posting._" if consumer == "social" else "._"
+    out = [f"# 🔭 Senpi Signals — {ts} UTC", "", note, ""] + HOW_TO_READ + ["", "---", ""]
+    cov = cov or {}
+    if cov.get("smart_money_lens") == "NO DATA" and cov.get("flow_lens") == "NO DATA":
+        out += ["> ⛔ **Smart-money lens UNAVAILABLE this run — no cohort positioning was supplied.**",
+                "> Everything below is OI, funding and price: the same inputs as market-pulse. The",
+                "> divergence and flow detectors did not find nothing — they were never fed. Run the",
+                "> proven-cohort fan-out (`senpi-smart-money`) and re-run before reading this as a",
+                "> smart-money report.", ""]
```
0xsarvesh, line 917: **Re-confirming finding 4 is still open — and it *is* user-visible**

Small correction to the status note: this was listed as "none user-visible", but this block is the one remaining finding that a user actually reads. `_render_md` writes it into `signals.md`, and `--print-feed` prints `signals.md` verbatim. Re-ran it against this head:

```
$ sed -n '/⛔/,/smart-money report/p' signals.md      # cohort dead, --print-feed output
> ⛔ **Smart-money lens UNAVAILABLE this run — no cohort positioning was supplied.**
> Everything below is OI, funding and price: the same inputs as market-pulse. The
> divergence and flow detectors did not find nothing — they were never fed. Run the
> proven-cohort fan-out (`senpi-smart-money`) and re-run before reading this as a
> smart-money report.
```

So on a cohort-dark run the user is told to "run the proven-cohort fan-out (`senpi-smart-money`) and re-run" — against this skill's own *"Talk about the market, never about the engine"* rule, and alongside the `Not measured this run: proven-trader positions.` line that already says it properly.

It's also still wrong in the universe-dark case, where it asserts "Everything below is OI, funding and price" with nothing below.

Given it's one line — `if consumer == "social"`, mirroring L918 four lines down — I'd take this one before launch rather than after. It's the last place the output says something untrue to a user.

0xsarvesh (commented):
## Re-review — all three S1s verified fixed, one gap and one correction

Re-ran every repro against `9d3b202` rather than reading the diff. Suite green: **58 passed** here, **80** on #657.

### Verified fixed — each reproduced against the new code

| Finding | Before | After (measured on this head) |
|---|---|---|
| **S1** missing price claims a move | `price=None → earliness 0.50 → "price already ran — late"` | `earliness 0.35 → "price not read this run"`. The twin is fixed too: `direction=None` now gives `"no side resolved"` instead of `"price is going against it"` |
| **S1** out-dir race | repro printed `ONLYINB … PRESENTED TO ITS USER: True` | **`False`** — per-run IPC files + `os.replace`. Repro no longer reproduces |
| **S1** universe SPOF | exit 0, *"nothing notable stands out right now"* | **exit 3**, *"could not read the market this run — the universe read failed after 2 attempts … This is an outage, not a quiet market."* Same on the brief path |
| **5** crowd basis unstated | `crowd LONG` either way | `crowd LONG per the 4h board` vs `crowd SHORT per the funding sign` — verified both appear in one feed |
| **8** test pinned the collision | `== 0.5` only | now asserts `unread == EARLINESS_PRICE_UNREAD`, distinct from the confirmed-read case |

The `os.replace` publish and returning the feed in the result rather than re-reading from disk are both better than what I suggested. Extending the dark-universe guard to `brief_text` unprompted was the right instinct — that was the path I was most worried about, since #657 makes it close every pulse.

Also confirmed the #671 rebase duplication is gone and the test now counts occurrences rather than checking presence — good catch, that was mine to own.

### One new finding, one correction

- **S2 (new), `sweep.py:209`** — the retry covers 4 of 8 reads. The four `discovery_*` reads bypass `_read()` through the vendored engine and are **not** retried, so a single blip on `discovery_get_top_traders` still silently downgrades the run to OI/funding/price. That's the failure the S1 fix was for, surviving on the one lens the skill is named after. Measured attempt counts inline.
- **Correction on finding 4** — it was listed as "none user-visible", but the ⛔ banner is precisely the user-visible one: `--print-feed` prints it verbatim, telling users to run `senpi-smart-money`. Re-confirmed on this head, inline.

### Where I land

The S1s are genuinely closed and I'm satisfied with the fixes. Neither item above is a correctness bug in the ranking, so I don't think either blocks step 1 — but **finding 4 is one line and is the last place the output says something untrue to a user**, so my preference is to take it here rather than "straight after". The retry gap (S2) is a follow-up; it degrades honestly today, it just degrades more often than it needs to.

6/7 (entry point + `history.py` split) and 9 (nits) are fine as follow-ups — agreed on not growing a PR under review.

betashop:
## Finding 4 taken, and you found a real hole in my own fix

**Finding 4 — you were right and my scoping was wrong.** I filed it as "not user-visible"; `--print-feed` prints `signals.md` verbatim, so that banner is exactly the user-visible one. Gated on `consumer == "social"`, the same gate the posting reminder four lines above already had. The ⚠️ partial-coverage banner gets it too, for the same reason. A user's one allowed clause stays `not_measured()`.

**The retry gap is the better catch.** My S1 fix went into `_read()`, which wraps four market/leaderboard calls — the cohort is gathered by the vendored engine calling `Client.mcp_call` directly, so I covered 4 of 8 reads and left the smart-money lens as exposed as it started. Moved into `mcp_call`; all eight retry now.

One deliberate narrowing: a `success: false` envelope is still **not** retried. It is returned as-is so `_read` degrades that source exactly as before and the vendored engine keeps seeing the response shape it expects rather than an exception — restricting the retry to transport failures leaves the engine's contract untouched, which is what I want on a launch path. Happy to widen it as a follow-up if you disagree.

Two guards, both proven: a single-attempt `mcp_call` fails the cohort-retry test — which counts attempts, since the run succeeds either way and merely loses a lens — and an ungated banner fails the engine-talk test.

**Full CI 1228 on #642, 1250 on #657**, and #657 is rebased on this head.

Remaining follow-ups, all yours: **6/7**, **9**, **10-14**, the `mcp_client` sequential-reads tail, and the deployed-vs-source parity question on `test_vendored_parity.py` — agreed that last one is a decision rather than a fix, and it is the one I would put next.

0xsarvesh (commented):
## Round 3 — `a67b68d` verified. Both remaining asks are closed.

Re-ran everything against the new head. **60 passed** (signals + pulse). No regressions: the three S1s and finding 5 all still hold.

### The two from last round, verified

**Finding 4 — the banner is now content-only.** Gated on `consumer == "social"` at L911/918, exactly mirroring the posting reminder. Measured on a cohort-dark `--print-feed` run:

```
⛔ banner blocks:                0   (was 1)
'senpi-smart-money' mentions:    0   (was 2)
honest degradation line:      True   → "_Not measured this run: proven-trader positions._"
```

The user now gets the one plain clause the conventions call for, and nothing about the engine. The debugging signal is still there on the `social` path where it belongs.

**Finding 15 — the retry now covers all 8 reads.** Counted attempts per tool, failing each in turn:

```
market_list_instruments          attempts = 2  -> RETRIED
discovery_get_top_traders        attempts = 2  -> RETRIED   (was 1, NO RETRY)
discovery_get_trader_state       attempts = 2  -> RETRIED   (was 1, NO RETRY)
leaderboard_get_markets          attempts = 2  -> RETRIED
leaderboard_get_momentum_events  attempts = 2  -> RETRIED
market_get_cross_asset_flows     attempts = 2  -> RETRIED
```

The cohort lens — the one the skill is named after — is no longer the unprotected path.

### Regression check on the earlier fixes

```
S1#3 unread price : "... price not read this run."          ✓
S1#3 twin no side : "... no side resolved."                 ✓
S1#2 race repro   : ONLYINB presented to user = False       ✓
S1#1 outage       : EXIT CODE 3 + "an outage, not a quiet market"  ✓
```

### Verdict: LGTM, ready for step 1

Every S1 and S2 I raised is closed. Nothing outstanding is a correctness bug or says anything untrue to a user.

**Still open, all agreed follow-ups** — recording them so they don't get lost, not asking for them here:

| # | Finding | User-visible |
|---|---|---|
| 9a | `tier 2 Tier 2` label duplication | yes, cosmetic |
| 11 | `momentum_event` has no staleness gate (`whale_move` does) | yes |
| 13 | tied cohort → LONG rather than "side unresolved" | yes, rare |
| 14 | upstream strings unescaped into feed text | yes, edge |
| 10 | `_traders_of` can't tell unparseable from empty | no |
| 6 / 7 | `sys.argv` seam; ~500 dead v2 lines → `history.py` split | no |
| 9b / 12 | dead `price_change_pct_24h`; dead `_ok` branch | no |
| 15b | no wall-clock deadline (tail now ~200s with retries) | no |

Of those, **11** is the one I'd pick up first — it's the same age gate `whale_move` already has, and a stale momentum event can currently lead the feed.

Nice turnaround on all of it.

0xsarvesh (approved):
**Approving.** `a67b68d` — every S1 and S2 I raised is closed, and I verified each by re-running the repro rather than reading the diff:

- missing price → `"price not read this run"` (and the directionless twin → `"no side resolved"`)
- out-dir race → repro no longer reproduces (`ONLYINB presented to user = False`)
- universe SPOF → exit 3 + `"an outage, not a quiet market"`
- crowd basis → `per the 4h board` / `per the funding sign`
- banner → content-only (0 blocks, 0 `senpi-smart-money` mentions on a user-facing dark run)
- retry → all 8 reads, verified by attempt counts per tool

60 passed locally, CI green.

Open follow-ups are recorded in my round-3 comment above — none are correctness bugs or user-facing untruths. **11** (`momentum_event` has no staleness gate) is the one I'd pick up first.

Not merging — leaving step 1 sequencing to you and Jason.

PR #643, opened 2026-09-16
aegis 1.2.0 / athena 1.2.0: size to the venue minimum on small wallets instead of emitting orders the engine refuses

betashop:
## What
Aegis 1.2.0 — and Athena 1.2.0, whose aegis leg is the same file (drift test) — size to the venue minimum on small wallets. `floor_to_venue_min()` in `scoring.py` raises a regime-scaled margin percent to the smallest order the engine fills ($12 notional, its bump), rounded up at 2 dp so the stored percent still clears, and the emit's reasons say "sized to the venue minimum". When even that floor would exceed the per-position cap, the scanner skips the asset and logs "wallet too small for this regime".

## Why
`margin_pct_for()` sizes down to a quarter of the base margin in a mild regime (12% × 0.25 = 3% of equity). On a wallet funded at the catalog minimum that is a $4–$8 notional, which the engine refuses every tick ("notional value $4.89 (margin $1.63 × 3x) is below minimum $10"): no hedge at all, and hours of skipped entries that read to the user as a broken runtime. A size the venue rejects is not a smaller hedge; it is none. Across all runtimes there were about 4,900 such refusals in the last four days.

The catalog floor stays at $40 — with this change it is honest: a $40 wallet's smallest entry now clears.

## Verification
- `strategies/aegis`: 17 passed — new: the floor arithmetic (untouched on a large wallet; 10% at $40; 13.34% at $30 after the round-up; skip at $20 where the floor would be 20% > 15% cap), small wallets at $30/$40/$60 all emit orders that clear $12 notional and never exceed the cap, a $20 wallet emits nothing and says why.
- `strategies/athena`: 3 passed (legs verbatim).
- Full CI: 1090 passed, 9 skipped; vendor parity OK; min-budget golden OK; validator ✓ aegis v1.2.0 ✓ athena v1.2.0; hygiene 0.
- Catalogs regenerated (`strategies/catalog.json`, `senpi-strategy-discover/catalog.json`).
- **CI coverage.** The aegis and athena suites this PR adds tests to run in CI once #649 merges, because its runner loops over `strategies/*/tests`. Until then they run locally only.


shnoodles:
**Review — the fix is right and the arithmetic holds exactly. Two things about what keeps it right.**

**Intent ✓, and the evidence checks out.** I reproduced the helper independently at 3x against a 15% cap:

| equity | pct | margin | notional | |
|---|---|---|---|---|
| $20 | — | — | — | SKIP (floor 20% > 15% cap) |
| $30 | 13.34% | $4.00 | **$12.01** | clears |
| $40 | 10.0% | $4.00 | **$12.00** | clears |
| $60 | 6.67% | $4.00 | $12.01 | clears |
| $1000 | 3.0% | $30.00 | $90.00 | untouched, `floored=False` |

Every example in the body reproduces, including the 2-dp round-up being what saves $30 (13.33% would land at $11.99). The `~4,900` claim is exact too — `below_min_notional` in `otel_logs` for Sep 12–15: **1,710 + 1,593 + 760 + 862 = 4,925**, across 8–13 distinct runtimes a day.

Also verified: both legs are byte-identical (`cmp` on `scan.py` and `scoring.py`), and the catalogs are genuinely generated — re-running `gen_catalog.py` changes only the `_updated` date stamp, and the two copies match each other.

---

### 1. `$12` now lives in four places, and the one that matters is not this one

`senpi-trading-runtime/scripts/min_budget.py:40`:

```python
BUMPED_NOTIONAL = 12.0     # the engine bumps a small order up to this notional
```

That constant is what computes aegis's advertised **`min_budget: 40`**. This PR's central claim — *"the catalog floor stays at $40 … with this change it is honest: a $40 wallet's smallest entry now clears"* — is true only while `VENUE_MIN_NOTIONAL == BUMPED_NOTIONAL`. Nothing couples them. Change either one and the catalog goes on advertising $40 as the budget where the design works, while the scanner sizes to a different floor — silently, because both sides still pass their own tests.

The other two copies, for completeness: `senpi-trader-research/scripts/research.py:42` (`MIN_NOTIONAL_USD = 12.0`, same rationale) and `strategies/caracal/*/scanners/scan.py:250` (`venueMinNotionalUsd`, defaulting to **10**).

A scanner cannot import from `senpi-trading-runtime/scripts`, so the literal fix is out. The cheap one is a single assert in the aegis suite, where both are importable:

```python
assert scoring.VENUE_MIN_NOTIONAL == min_budget.BUMPED_NOTIONAL
```

One line, and the day someone retunes the engine floor it fails loudly instead of un-honesting the catalog.

### 2. The 20 tests this PR adds never run in CI

`.github/workflows/tests.yml:38` runs `strategies/tests` — **not** `strategies/<id>/tests`. `strategies/tests` collects 297 tests; `strategies/aegis/tests` and `strategies/athena/tests` are not among them.

This is not something this PR introduced — there are **37** `strategies/*/tests` directories and CI runs none of them. But it is what makes this diff's verification story local-only: "aegis 17 passed, athena 3 passed" is true on your laptop and nothing will keep it true. The floor arithmetic is exactly the kind of thing that rots quietly.

Worth knowing before fixing it: those 37 dirs cannot currently be collected as one batch — `pytest strategies/*/tests` dies with an `INTERNALERROR` during collection. **Correction to my first pass:** I put that down to duplicate basenames, which was a guess and is wrong. The actual cause is a **module-level `sys.exit()`** in a copied `test_package_shape.py` / `test_engine.py` — **8 files across 4 packages** (`chimp`, `gibbon`, `gorilla`, `orangutan`). The CI workflow comment names only `chimp`, so it under-reports it too. Individually the dirs are fine. So it is "add the dirs plus a collection fix", not a one-word change — probably its own PR, but this one is the reason to prioritise it.

### Ponytail (diff-scoped)
Nothing to cut. `floor_to_venue_min` is pure, small, returns `(pct, floored)` so the caller can say *why*, and the `(None, True)` skip case is documented at the definition. The stderr line names the wallet, the cap and the leverage — that is the line someone will actually need. `margin_pct_for` never returns `None`, so the new `is None` branch can only mean the cap case; no mislabelled skips.

### Verified locally
Branch @ pr-643, 0 behind main: aegis+athena **20 passed**, full CI **1,086 passed, 9 skipped, 0 failed**, vendor parity ✓, min-budget golden ✓ (109 templates), `validate_strategy.py` ✓ aegis v1.2.0 / athena v1.2.0. (Body says 1,090; I count 1,086 on the same base — cosmetic, but the ledger copies these.)

betashop:
Thanks — both points are right, and both are done.

**1. The `$12` coupling — fixed in 9627ae1.** Added the assert you suggested, in the aegis suite where both modules load:

```python
def test_the_venue_floor_is_the_same_number_the_catalog_minimum_is_built_from():
    ...  # loads senpi-trading-runtime/scripts/min_budget.py by path
    assert scoring.VENUE_MIN_NOTIONAL == min_budget.BUMPED_NOTIONAL
```

Checked that it bites: forcing `VENUE_MIN_NOTIONAL = 10.0` fails it. aegis is now 18 passed, athena 3. I left the other two copies you found (`research.py`'s `MIN_NOTIONAL_USD`, caracal's `venueMinNotionalUsd` default of 10) alone — they aren't on this PR's claim path, and caracal's 10 is a separate question about whether it should be 12.

**2. The 20 tests never running in CI — opened as its own PR, #649.** You were right that it's not a one-word change. What I found once I ran all 37 directories:

- 33 pass under pytest, one process each. They can't share a session, as you saw — duplicate basenames plus each suite putting its own `scanners/` on `sys.path`.
- The other 4 are the chimp family, and they aren't failing — they're **script-style**: checks run at import and the file ends `sys.exit(1 if FAIL else 0)`. pytest can't import them (the exit aborts collection), but `python3 <file>` passes (chimp 26/26) and exits 1 on a real break. So all 37 are green on `main` today; CI just couldn't see them.

#649 adds a runner that classifies each file by content (`def test_` / `class Test` → pytest per directory, anything else → `python3 <file>`), gates on every exit code, and stops with exit 2 if pytest is missing. Proven green on `main` in ~5 s and red on a break in either style.

One thing the aegis mutation turned up, noted on #649: bumping the high-conviction multiplier 1.5× → 9.9× was caught by **athena's drift guard, not aegis's own suite**. `test_margin_pct_for`'s only high-conviction case sits where the 15% cap clips both values to the same answer, so that branch is never exercised uncapped. Small follow-up, not a blocker here.

**On the count:** with the exact CI command on this branch I get `1090 passed, 9 skipped`. I can't reproduce 1086 here — most likely a different base between your checkout and mine. Happy to take your number for the ledger if it's from the merge ref.

shnoodles (approved):
Approved. Findings addressed and verified at `253c07a4`: the `VENUE_MIN_NOTIONAL` / `BUMPED_NOTIONAL` coupling assert exists and bites (forcing 10.0 fails 3 tests), and the CI gap is split out as #649 rather than bolted on here. Floor arithmetic reproduces exactly ($40→10%, $30→13.34% at $12.01, $20→skip), legs byte-identical, catalogs genuinely generated, and the ~4,900 evidence claim is 4,925 in telemetry. CI green.

PR #644, opened 2026-09-16
portfolio 1.25.0: read a closed trade's side from the row's type, never from the sign of an at-close size

betashop:
## What
`senpi-portfolio` 1.25.0: a closed trade's side is read from the row's side label (`type` — "Close Short" / "Close Long" — or `dir`/`side`/`direction`/`positionSide` containing "long" / "short"). A row without a label is `None` and counts as `unknown_side`; the side is never guessed from the size's sign, from realized PnL against the price move, or from a buy / sell side. One helper, `_closed_direction()`, drives both the per-row `direction` and the longs/shorts/`unknown_side` counters.

## Why
`discovery_get_trader_history` returns a POSITIVE `szi` on a closed short (the side lives in `type`). The old rule `szi >= 0 → long` therefore labelled every closed short as a long, so long/short counts, "shorts are your winners" and the recent-closes table were wrong for anyone who shorts. #607 fixed only the missing-size case; the sign rule and its fixture ("signed szi, >0 closed a long") were still wrong. Correction: an earlier version of this body said `senpi-improve-trades` already resolves the side this way. It does not: its discovery path never reads `type` and reads the unsigned size's sign, so it still labels closed shorts as longs. That fix is #650.

## Verification
- `senpi-portfolio/tests`: 92 passed (was 91: one fixture corrected to the real row shape; the missing-size test replaced by two — a closed short with positive size and `type` reads short; an unlabelled row is `unknown_side` and stays in the record whatever its size, PnL or buy / sell side suggests). Each removed guess, restored, fails the second test.
- `senpi-improve-trades/tests`: 71 passed, tree untouched.
- CI steps: vendor parity OK; min-budget golden OK (109 templates); the skills suite 1091 passed, 9 skipped.
- README skill table synced to 1.25.0.


shnoodles:
**Review — correct, and confirmed against production data. One fallback I'd drop, and the doc that caused this.**

**The premise is right, and the blast radius was total.** I pulled real `discovery_get_trader_history` results from telemetry. Last 2 days: **57,784** results containing `Close Short`, **49,964** containing `Close Long`. A real row, verbatim:

```json
"szi": "100000.3",
"realizedPnl": "3925.23971",
"type": "Close Short",
```

`szi` is **positive on a closed short**, and the side is in `type` — exactly as this PR says. So the old `szi >= 0 → long` rule did not mislabel an occasional row; it labelled **every closed short as a long**, on roughly half of all closed trades.

Independent probe against the new code with that real shape (one closed short, one closed long):

```
new:  longs=1 shorts=1 unknown=0   recent dirs=['short','long']
old:  longs=2 shorts=0
```

**Worth saying plainly:** this is the actual fix for the M416086 complaint. I reviewed #607, flagged that a missing `szi` defaulted to long, and accepted the sign convention itself — the real defect was one layer under the thing I checked. #607 fixed the denominator; this fixes the answer.

---

### 1. The document that caused this is still wrong

`senpi-hyperliquid-mcp/src/resources/discovery-top-traders.resource.ts` documents the closedPositions example with:

```json
"type": "market",
```

That is the field that actually carries `"Close Short"`. The MCP's `ClosedPosition` interface types it as a bare `string`, so nothing contradicts the example either. An author reading that page has no way to discover the side is in `type`, which is precisely why reading it off `szi` looked reasonable for as long as it did. Worth a two-line PR on the MCP side, or this comes back.

### 2. The third fallback can be confidently wrong where `None` would be honest

```python
move = (exit_px - entry_px) if (...) else None
if move and pnl:
    return "long" if ((move > 0) == (pnl > 0)) else "short"
```

Tier 1 (the `type` label) resolves **100%** of real rows — both production values match it. So tiers 2 and 3 are already unreachable on live data. Tier 2 (signed `szi`) is a harmless shape-guard. Tier 3 is not: it infers the side from whether PnL and the price move share a sign, and `realizedPnl` is net of fees (`totalFees` is a sibling field, and the sampled row carries `100.23822` of them). A small winner that fees turn negative gets labelled backwards — a wrong side reported with full confidence, where returning `None` would have put it in `unknown_side` and stayed honest.

Given this PR exists because a confidently-wrong side reached a user, I would drop tier 3 and let those rows fall to `unknown_side`. It is also one less branch.

### Ponytail (diff-scoped)
Otherwise clean. The label loop checking five field names is justified by the fail-open contract, `unknown_side` keeps the gap visible, and the `REAL_CLOSED_SHORT` fixture pinned to the real payload is exactly the right test to have written — it is what makes this diff verifiable at all. Minor: the exchange-style aliases (`"a"`, `"s"`, `"ask"`, `"bid"`) match nothing in the real feed; harmless, but they are speculative.

### Verified locally
Branch @ pr-644, 0 behind main: `senpi-portfolio/tests` **94 passed**, full CI **1,093 passed, 9 skipped, 0 failed**. Version 1.24.0 → 1.25.0 consistent across title, README and SKILL.md.

betashop:
Both points taken, and I went one step further on the same principle.

**1. The document.** Senpi-ai/senpi-hyperliquid-mcp#216 fixes the guide. `szi` is documented as unsigned at close and `type` as the side. The example row reads `"type": "Close Long"`, and the line saying `szi: "-2.0"` would be a closed short is gone. The resource test pins the new wording and the absence of the old. Telemetry behind it:

| window (UTC) | result |
|---|---|
| Sep 15 17:00 to Sep 16 17:00 | 649,058 `szi` fields on closed rows, 0 negative |
| Sep 16 14:00 to 17:00 | 95,409 closed rows, every one `Close Short` or `Close Long` |

**2. The fallbacks (c5d2e23).** Tier 3 is gone. I also dropped tier 2 and the buy / sell aliases, because each can be confidently wrong where `None` is honest:
- **Tier 2:** on this feed `szi` is always positive, so a row that lost its `type` would read long whatever its side. That is the original bug again.
- **Aliases:** on a closed row, a buy / sell side can be the closing fill, which is the opposite side.

A row without a label now counts as `unknown_side`. One test covers the four unlabelled cases: a signed size, an unsigned size, a long from 100 to 100.2 that fees turn into a −0.05 loss, and `side: "sell"`. Restoring each removed guess fails it. If you'd rather keep tier 2 as a shape guard, it is one line to restore.

**Correction to the body.** It said senpi-improve-trades already resolves the side this way. It does not. Its `_direction` never reads `type` and falls back to the sign of `szi`, so it still labels every closed short a long on the real feed. A real-shaped closed short comes back `long`. That fix is a separate PR, and the body now says so.

**Verified:** `senpi-portfolio/tests` 92 passed. The full CI pytest command gives 1,091 passed, 9 skipped, exit 0. Vendor parity and the min-budget golden both exit 0.

betashop:
**Correction to my reply above, and to one premise in the review.** `realizedPnl` on a closed row is gross, not net of fees. I checked four live rows, two closed longs and two closed shorts. On each, `realizedPnl` equals `szi × (exit − entry) × side` exactly, and the fees sit separately in `totalFees`. Fees therefore can't flip its sign. On those rows, the PnL-vs-move fallback would have agreed with the label.

The fallback stays removed, for the reason that does hold. The label is on every closed row the feed returns (95,409 of 95,409 over three hours), so the fallback never ran on real data. An unlabelled row is a shape the feed doesn't send, and `unknown_side` is the honest answer there.

61b41ee corrects the docstring and renames the test row to what it pins: no PnL inference. There is no behaviour change. My earlier reply and the c5d2e23 commit message repeated the fee premise without checking it. The same wording is corrected on #650 (d1d3d24).

**Verified:** `senpi-portfolio/tests` 92 passed. With the inference put back, the test fails. The CI pytest command gives 1,091 passed, 9 skipped, exit 0.

shnoodles:
**Recheck — everything addressed, and you were right to correct me.**

**`realizedPnl` is gross of fees. My stated reason for dropping tier 3 was wrong.** I checked your correction against a live row rather than taking it:

```
BTC  entryPx 77580.2686445333   exitPx 78141.5007836   szi 75   type "Close Short"
szi × (exit − entry) × −1  =  −42,092.41
realizedPnl reported       =  −42,097.56      difference −5.15
totalFees                  =   3,503+
```

If `realizedPnl` were net of fees the gap would be ~3,500, not ~5 (the −5.15 is averaging across 316 fills). So "a small winner that fees flip gets labelled backwards" was not a real failure mode, and the docstring correction in `61b41ee9` is right.

The conclusion still holds for the reason you kept: the label is on every row the feed returns, so the fallback never ran on live data — a guess that can only fire when the reliable field is absent is a guess that can only be wrong.

**Verified the rest:**
- Tiers 2 and 3 and the buy/sell aliases are gone; `_closed_direction` is the label loop and `return None`.
- Probe with a real-shaped short + long + three unlabelled rows (signed size, fee-flipped long, `side: "sell"`) → **longs=1, shorts=1, unknown_side=3**, and `recent[].direction` is `None` for all three. Exactly the intended behaviour.
- `senpi-portfolio/tests` 92 passed; with `senpi-strategy-ops/tests` + `strategies/tests`, 799 passed / 0 failed.

**And your `senpi-improve-trades` finding is real — I reproduced it.** `review.py:746` reads `_field(rec, "dir", "side", "direction", "positionSide")` with **no `type`**, so on the live feed tier 1 never matches and it falls to the `szi` sign, which is always positive at close:

```
_direction(real closed SHORT)  ->  'long'      (correct: short)
_direction(real closed LONG)   ->  'long'
```

Its docstring also states the opposite premise — *"the at-close size is often 0"* — where production shows it positive. Worth carrying the same fix and the same test row across; that skill is narrating the same mislabelled sides today.

Nothing outstanding from me on this PR.

betashop:
Thanks for checking the correction against a live row.

The improve-trades half is #650 (senpi-improve-trades 1.10.0). Its `_direction` reads the side only from a label, `type` included, and returns `None` when there is none. The "at-close size is often 0" docstring premise is gone, and `test_a_closed_trade_reads_its_side_from_the_label_never_the_size_or_pnl` pins a `Close Short` row with a positive `szi`.

#650 and #652 both edit the improve-trades version lines in `README.md` and `SKILL.md`, and `git merge-tree` shows the conflict. So #650 merges first and #652 rebases.

shnoodles (approved):
Approved at `61b41ee9`. Premise confirmed against production (57,784 `Close Short` vs 49,964 `Close Long` in 2 days; a live row carries positive `szi` with the side in `type`). Tiers 2 and 3 and the buy/sell aliases are gone; my probe with a real-shaped short, long and three unlabelled rows gives longs=1 / shorts=1 / unknown_side=3. You were also right to correct me on `realizedPnl` being gross — I verified that against a live row. 92 portfolio tests, CI green.

PR #645, opened 2026-09-16
account-status 1.3.0: AI credits are the usage meter — say what they are, never that chat is free

betashop:
## What
`senpi-account-status` 1.3.0: a section "AI credits — the usage meter" — what credits are (plan usage credits, shown as the meter in the app header), what spends them (every message and tool turn; long threads cost more because the whole conversation is re-sent), what they are not (trading money — never tradable, withdrawable or fundable), and the one fact that decides the answers: no Senpi tool returns the balance, so the agent points at the meter and never invents a number. A say / never-say table ("chatting is free" is on the never side), routing for the meter/bubble/cost/"can I trade my credits" questions, and the frontmatter description updated so the router matches them.

## Why
Users ask what the header meter is and why it goes down; the agent had nothing on the subject and answered that chat costs nothing, then guessed. A closed ticket from July ("agent unable to answer questions about free credit usage") is the same gap; this is the skill-side half of closing it. The MCP side (exposing the balance to the agent) is a separate, small backend ask.

## Verification
- `senpi-account-status/tests`: 15 passed (new needle tests pin the section, the "can't read it from here" statement, the never-"free" rule, the not-tradable rule, and the description; the README row is pinned to the frontmatter version).
- CI: vendor parity OK; min-budget golden OK; the skills suite 1090 passed, 9 skipped.
- **CI change:** this skill's tests were never in the workflow's pytest list and cannot join it (its `test_status.py` / `test_skill_surface.py` basenames and `import status` collide with strategy-ops' under a rootdir-less run), so `tests.yml` gains one step that runs them on their own. Mutation check: flipping the "chatting is free" needle fails 2 tests.


shnoodles:
**Review — the load-bearing claim checks out, and the CI fix is the right one. No change requested.**

**Intent ✓.** A trial user asking to delete their account 11 minutes after watching the meter drop is the failure, and this answers it from the other side than `senpi-hyperliquid-mcp` #212: that one gives the overview guide one paragraph naming the bubble; this gives the skill the whole rule set. They agree — meter is the balance, spend is usage not a trading loss, never quote a plan size.

**"Credits are explained, never read" is true, and it is the claim everything else rests on.** I checked rather than assuming: **no MCP tool name contains "credit"**, `user_get_me` returns id / wallets / createdAt / referralCode, `user_get_senpi_points` returns points / tier / fees, and the engine has no `credits` section. So "no tool returns the balance; point at the meter" is accurate, and the hard prohibition on inventing "a number, a percentage or a 'roughly'" is the correct consequence rather than an over-cautious one.

The sharpest line is **"Chat is not free of credits … including the turn that asks what a turn cost."** That is the exact question the incident user asked, and "never 'nothing'" closes the specific wrong answer they got.

**The CI addition is correct and well-diagnosed.** A separate invocation for `senpi-account-status/tests`, with the reason in a comment: `test_status.py` / `test_skill_surface.py` share basenames with `senpi-strategy-ops`' tests and both `import status` from their own `scripts/`, so a shared rootdir-less run fails collection with "import file mismatch". That is a real pytest failure mode and its own invocation is the cheapest fix.

Worth knowing: this is the **third** PR in this batch touching test wiring, and the repo now has two distinct collection blockers — this basename/`import status` collision, and a **module-level `sys.exit()`** in copied `test_package_shape.py` / `test_engine.py` (8 files across `chimp`, `gibbon`, `gorilla`, `orangutan`) that aborts the `strategies/*` tree. #647 adds a skill's `tests/` dir the simple way; #642 adds 1,514 lines whose 33 tests are in `scripts/` and reach no CI path at all. Between the three there is a coherent one-PR cleanup here for someone.

### Ponytail
Nothing to cut. The Say / Never-say table is this repo's established shape for copy an agent reproduces under pressure, and every row pairs a real user question with the wrong answer that was actually given. The rules are short and none of them restate another.

### Verified locally
Branch @ pr-645, 0 behind main: `senpi-account-status/tests` **15 passed** (its own invocation), full CI **1,090 passed, 9 skipped, 0 failed**. Version 1.2.0 → 1.3.0 consistent across title, README and SKILL.md.

betashop:
Thanks — nothing to change here. For the record, the two CI-hunk PRs in this batch (#645 adding its own step, #647 appending to the pytest line) plus #642 and #649 all touch `tests.yml`; none conflicts textually except #642 and #647 on the same pytest line.

betashop:
Correction to my note above — I stated it from reading, then checked with real merges (`git merge-tree` on each pair) and it was incomplete. Two pairs conflict in `tests.yml`, not one:

| pair | result |
|---|---|
| #642 + #647 | conflict |
| **#647 + #649** | **conflict** — the one I missed |
| #642 + #645, #642 + #649, #645 + #647, #645 + #649 | clean |

So #647 is the hinge: whichever of #642 / #649 lands before it, #647 needs a rebase. Merging #647 first and rebasing the other two avoids both.

betashop:
Pushed 23ba18d after your recheck. A new case exposed two gaps in the credits section: a question about whether a *strategy* uses credits, and "credits" asked in another language.

- **A running strategy is not what spends them.** The runtime ticks without a model call (`senpi-trading-runtime/SKILL.md`). The exception is a `decision_mode: llm` action, which calls a model when it fires. A cron the agent set up to check a strategy does spend credits, so the agent names any it runs.
- **Say / Never say** gains a row for that question. The agent never says "running the strategy costs nothing" while a check-in cron runs, and never offers a wallet or strategy balance as proof about credits.
- **Routing.** "Does my strategy use credits?" joins the questions, in the frontmatter too, in any language. "Credits" means these AI credits unless the user says leverage, a loan or funding; if it is unclear, the agent asks.

`senpi-account-status/tests`: 16 passed. The new test fails if any of the three is removed; I removed each in turn to check. CI: 1,090 passed, 9 skipped, and the GitHub `test` check passes. No version change, since 1.3.0 is unmerged. This text postdates your ready-to-merge, so I moved the ledger row back to in review.

shnoodles (approved):
Approved at `23ba18d1`. The load-bearing claim holds — no MCP tool returns a credit balance — and the added rule is sharper than the original: a running strategy's runtime ticks without a model call, with `decision_mode: llm` correctly called out as the exception. The separate invocation for the basename collision is the right fix and its reason is in the comment. CI green.

PR #646, opened 2026-09-16
trade 1.5.0: the three numbers before a hand-placed order — share of account, dollars at the stop, applied leverage

betashop:
## What
`senpi-trade` 1.5.0 — three additions to the manual-position branch:
- **The three numbers before the replay.** Share of the account the margin takes, dollars at the stop (and as a share of the account), and the leverage that will actually apply (the asset's `max_leverage` when lower than asked — say the clip). Above 50% of the account in one position or 20% of the account at the stop, the number goes in its own sentence and gets a fresh yes to that sentence. Re-run on every add, stop move and leverage change.
- **Read back the leverage.** The position's `leverage.value` is the one that applies; quote it, not the request, and fix any label that carries the asked-for number.
- **"Watch it for me" is the ladder, never a cron.** `ratchet_stop_add` is the engine-side watcher at zero model cost (profit-lock only, said plainly); a cron is a full agent turn per firing and is not a stop.
Three guardrail rows carry the same rules.

## Why
A hand-placed position went almost entirely into one thin coin, then doubled, then had its stop widened to over 40% of the account — each step was asked for, and nothing ever said the total. The same session asked for 10× on an asset the venue caps at 3×; the position opened at 3× and kept the "10x" name. Elsewhere the only "watch it for me" the agent offered was a paid cron. The protocol already says "quote, never compute" and "leverage is the exchange's"; this makes the sizing readout and the fresh-yes threshold explicit.

The 50% / 20% thresholds are a proposal — tune in review.

## Verification
- `senpi-trade/tests`: green (one needle test added; the README row is pinned to the version).
- The CI skills suite: green.


shnoodles:
**Review — the rules are right and well-evidenced. One real defect in the test file.**

**Intent ✓.** All three additions trace to the incident: a wallet that went almost entirely into one thin coin one small request at a time with nothing ever stating the total; 10× asked and 3× applied while the position kept the "10x" label; and a cron offered as "watch it for me". The sharpest clause is **"a yes to the setup is not a yes to the concentration"** — that is the exact step that was skipped, and putting the number in its own sentence is what forces it to be answered.

*"Re-run the three numbers on every add, every stop move and every leverage change, not only at open"* is the other load-bearing half — the incident was cumulative, not a single bad order.

---

### 1. `import os` landed above the shebang — the shebang and the docstring are both dead

```python
1  import os
2  #!/usr/bin/env python3
3  """The protection protocol is resident in senpi-trade/SKILL.md, …"""
```

Two consequences, both verified by executing the file:

- Line 2 is no longer a shebang, just a comment. `./test_trade_skill_surface.py` stops working.
- The docstring is no longer the module's first statement, so it is a discarded expression: **`module.__doc__` is `None`**.

The file already imports at lines 6–8 (`re`, `unittest`, `pathlib`). Move `import os` there and both problems go away.

### 2. The new test skips the helper its siblings use

`_body(path)` is defined at line 15 and used by the two sibling tests (lines 23 and 41). The new `test_the_sizing_readout_and_its_thresholds_are_in_the_body` (line 31) opens `SKILL.md` raw instead:

```python
body = open(os.path.join(os.path.dirname(__file__), "..", "SKILL.md"), encoding="utf-8").read()
```

That is what pulled `import os` in. Using `_body(SKILL)` like its neighbours removes the import entirely, keeps the frontmatter stripped (so a needle cannot match in metadata), and makes the three tests read the same way. One-line change that resolves both findings.

### Ponytail (diff-scoped)
Ten needles on the new test is the most in this batch, and several pin prose rather than rule — `"share of the account"`, `"dollars at the stop"`, `"leverage that will actually apply"` are the three numbers and are worth pinning; `"Then the three numbers, before the replay"` pins a sentence opener. The thresholds (`50%`, `20%`) and `"a yes to the setup is not a yes to the concentration"` are the ones that must not silently disappear. The rest could go.

The three guardrail-table rows are not duplication — that table is the scannable index for the same rules, and it is the file's existing pattern.

### Verified locally
Branch @ pr-646, 0 behind main: full CI **1,091 passed, 9 skipped, 0 failed**. Version 1.4.0 → 1.5.0 consistent across title, README and SKILL.md.

betashop:
Both confirmed by running the file, and both fixed in ccb2fde.

**The shebang and docstring.** Before: line 1 `import os`, line 2 the shebang, and `module.__doc__ is None → True`. The new test now reads through `_body(SKILL)` like its siblings, so the import is gone entirely. After: shebang on line 1, `__doc__` restored, frontmatter stripped before matching.

**Needles.** I checked discrimination before cutting anything: every one of the ten was absent on `main` and present on the branch, so none was a dead guard. Your prose-versus-rule point stands on its own, though. Dropped:
- `Then the three numbers, before the replay` — a sentence opener
- `Place or grow a hand position without the three numbers` — a table-row label restating the rule

Kept one needle per rule. That's the six you named (the three numbers, both thresholds, the fresh-yes clause), plus `Read back the leverage` and `is the ladder, never a cron`. Those two are rules in their own right, and without a needle either could vanish silently.

Proof: changing *"50% of the account in one position"* to *"most of the account"* → 1 failed. Passes under `pytest` (5) and `python3 <file>` (OK).

shnoodles (approved):
Approved at `ccb2fde9`. Both defects fixed and verified by executing the file: shebang back on line 1, `__doc__` restored, and the new test reads through `_body(SKILL)` like its siblings so the stray import is gone. Needle trim is well-judged — you checked discrimination before cutting, and kept one per rule. CI green.

PR #647, opened 2026-09-16
deposit-withdraw-transfer 1.4.0: a FAILED top-up leaves the money in Spot — recover it, say where it is, never loop

betashop:
## What
`senpi-deposit-withdraw-transfer` 1.4.0: the money rails learn what to do when a top-up fails, and what "everything" means.
- **Topping up a strategy: precheck, poll, recover.**
  - **Gate:** check the amount against the funding wallet's free perps, `withdrawable` on the `main` side of `strategy_get_clearinghouse_state`, and say that figure next to the amount. `total_in_hyperliquid` is an upper bound only, because it includes margin locked in open positions.
  - **Poll:** poll the request and never re-submit while PENDING.
  - **On FAILED:** locate the money before saying anything. If Spot rose by the amount (`account_get_portfolio` with `forceFetch`), the first leg ran and the second did not. Move it back with `transfer_spot_to_perps` and say which balance it was in.
  - **Retries:** at most one re-submit, only if the user asks, and never for the same strategy twice. A second FAILED goes to Senpi support with the request id, never to a fresh deploy, which would charge another wallet-creation fee and leave the old strategy running.
- **Withdrawals.**
  - `SERR037` with `details.available` → retry once with exactly that figure.
  - "Withdraw everything" means the exact available figure, never rounded.
  - Afterwards, confirm ~$0 and say that an ACTIVE strategy with dust keeps scanning; offer to close it.
- **Costs before money moves.**
  - The creation-fee reserve (~$1, budgeted $1.50) goes beside the $10 floor.
  - Multi-wallet plans get the per-wallet minimum and fee up front.
  - Quote only fees a tool returned.

A say / never-say table carries the sentences.

## Why
A top-up could fail after its first leg and report "no funds moved … still in the funding wallet" while the money sat in Spot. Agents re-submitted in loops and sent users to support without locating the money. The backend side of that failure has since been fixed. The rules still hold for any failed money operation: find the money, name the balance, stop after one retry.

Other problems this fixes:
- The old precheck quoted account value as free capital, which over-states what a top-up can draw whenever a position is open.
- Withdrawals that failed by cents were bounced back to the user for a new number.
- "Withdraw everything" left dust that kept a strategy scanning.
- Wallet-funding costs surprised users after the fact.

## Verification
- `senpi-deposit-withdraw-transfer/tests`: 14 passed. These are needle tests for each rule, the say/never-say table, README↔SKILL version parity and a no-user-ids check, plus guards that fail if the account-value precheck or the fresh-deploy remedy comes back. Both guards were proven against the previous text.
- Every `run:` step of `tests.yml` passes on the branch rebased onto main (548c9a1): vendor parity, min-budget golden, strategy package suites, and the full suites at 1,191 passed / 9 skipped.
- **CI change:** this skill had no tests before; its directory is added to the workflow's pytest list.
- Field names were checked against live `account_get_portfolio` responses and the MCP tool descriptions (`account_get_portfolio`, `transfer_spot_to_perps`).


shnoodles:
**Review — the best-shaped PR in this batch, and it contains one field that undoes its own precheck.**

**Intent ✓, and it goes further than asked.** This is the deposit-side counterpart to `senpi-hyperliquid-mcp` #210, and it fixes the thing I flagged there: **"FAILED — find the money before you say anything"**, with the status message explicitly demoted to *"a hint, not a fact"* and three concrete branches (Spot rose → recover and say so; both unchanged → name perps; anything else → don't re-submit, hand over the request id). *"Never say 'still in the funding wallet' without naming the balance — perps or Spot"* is the sentence that closes the original complaint.

The two-FAILED rule is also better than the generic "stop": naming that **some older strategy wallets refuse the second leg every time** and offering *deploy fresh for a new wallet* vs *support with the request id* gives the user a route instead of a dead end. And the Say / Never-say table is the right format for copy an agent has to reproduce under pressure.

---

### 1. The precheck reads the one field the MCP says is not the available amount

Rule 1 says:

> the funding wallet's perps USDC is `total_in_hyperliquid` … **The amount must not exceed the perps figure**

`account_get_portfolio`'s own description, **already on `main`**, says the opposite:

> **CRITICAL: `total_in_hyperliquid` includes open-position margin, so it is NOT the withdrawable amount** — never present it as available to withdraw (use `withdrawable` from `strategy_get_clearinghouse_state`).

And the open `senpi-hyperliquid-mcp` #211 restates it: *"what can fund a NEW strategy is the free perps USDC in the funding wallet (its clearinghouse `withdrawable`, which this response does not expose)"*.

So a funding wallet holding any open perps position reports a `total_in_hyperliquid` larger than its free USDC. A top-up sized to that figure **passes this precheck and then FAILS** — which is the exact outcome the precheck exists to prevent, now with an added line of false assurance to the user ("Your funding wallet holds $X in perps").

This PR already knows the right read: it uses `withdrawable` from `strategy_get_clearinghouse_state` for the "withdraw everything" rule a few lines up. Same call, on the funding wallet, for the precheck — `withdrawable` (+ `total_spot_usd_in_hyperliquid` for the Spot leg). `total_in_hyperliquid` is fine as an upper bound, and saying it as one would still be honest; it just cannot be the gate.

Worth noting this is also the second contradiction between these two repos on this exact field ( #211's `idle_in_embedded` finding was the first ) — the ledger's P1 row for that would cover this too.

### 2. Adding a tests directory to CI — right move, and it is the template for a bigger gap

`.github/workflows/tests.yml` gains `senpi-deposit-withdraw-transfer/tests`. Good: the needle guards actually run.

It is also the fix-shape for something I raised on #643 — **37 `strategies/*/tests` directories that CI never runs**, including the 20 tests #643 adds. That tree needs one extra step first, and this PR's own workflow comment is close to it but under-reports: the blocker is a **module-level `sys.exit()`** in a copied `test_package_shape.py` / `test_engine.py`, and it is **8 files across 4 packages** — `chimp`, `gibbon`, `gorilla`, `orangutan` — not `chimp` alone. Worth correcting that comment here while you are in the file, since the next person to try will grep for `chimp` and conclude it is already handled.

### Ponytail (diff-scoped)
Large but earned — it is a rules surface, and the four numbered rules are ordered by when the agent needs them. The Say / Never-say table is not duplication; it is the scannable form of the same rules, which is this repo's pattern.

One trim: rule 4 restates "recover the Spot leg (rule 3), re-run the precheck (rule 1)" in full rather than pointing at them, and the row `"Withdrew the exact available figure…"` repeats the withdrawal rule that already has its own bullet above the table.

### Verified locally
Branch @ pr-647, 0 behind main: CI **as this PR redefines it** (including the new directory) → **1,102 passed, 9 skipped, 0 failed**. Version 1.3.0 → 1.4.0 consistent across title, README and SKILL.md.

betashop:
You're right, and this one is on me. The subagent that built this PR listed the caveat — *"total_in_hyperliquid can include position margin"* — in its assumptions, and I shipped the precheck anyway.

**Confirmed before fixing.** `account_get_portfolio`'s live description: *"CRITICAL: total_in_hyperliquid can include open-position margin, so it is NOT the withdrawable amount"*. MCP `main` is more specific: *"total_in_hyperliquid is the perps ACCOUNT VALUE (free balance PLUS margin locked in open positions), so it OVER-states fundable capital … treat it as an upper bound only."* And `strategy_get_clearinghouse_state` states it *"works for any account (a strategy wallet or any trader's wallet)"*, so the read you proposed is callable on the funding wallet.

**Fixed in 8a73d11.** Rule 1 now takes two reads:
- **Free perps, the gate:** `withdrawable` on the `main` side of `strategy_get_clearinghouse_state` for the funding wallet.
- **Spot:** `total_spot_usd_in_hyperliquid` from `account_get_portfolio`, `forceFetch: true`.

`total_in_hyperliquid` stays only as an upper bound, in the MCP's own words: if even it is short, the top-up will certainly fail, so go straight to the funding card. Rule 3's re-read takes both reads again. The say line is now *"$X free in perps"*, and quoting account value as free moved to the never-say column.

One trap I had to step around: this skill already says never read a deposit address out of `user_get_me`. The funding wallet's address does come from there, so the text says so outright — *"read it to check the balance, **never** to hand out as a deposit address"*.

**The test was guarding the defect.** `test_perps_precheck_before_the_top_up_call` asserted the string *"perps USDC is `total_in_hyperliquid`"*. Replaced with a test that pins the `withdrawable` gate and the upper-bound language, plus a second that fails if any of the three account-value phrasings returns. Proof: the pre-fix `SKILL.md` with the new tests → **both fail**; with the fix → 13 passed.

**Ponytail, both taken.** Rule 4's second path is now *"apply rule 3 again"*. The say-table row that repeated the withdraw-all bullet is gone; the withdraw needles still hit the bullet.

**The workflow comment** — correct, it's 8 files across chimp, gibbon, gorilla and orangutan. I landed that edit on #649 instead (e33e8cf), since #649 is the PR that actually runs that tree, and two PRs rewriting the same comment would just conflict.

CI as this PR defines it: **1,103 passed, 9 skipped**; parity ✓, golden ✓.

shnoodles (approved):
Approved at `8a73d11d`. The precheck now gates on `withdrawable` from `strategy_get_clearinghouse_state` with `total_in_hyperliquid` demoted to an upper bound, which is what the MCP's own description requires. Good catch that the old test was guarding the defect. 13 tests pass. The FAILED-top-up branch — locate the money before you say anything, and name perps or Spot — is the best-shaped rule in this batch.

shnoodles:
**Yes — push it. But one part of it is now wrong advice, and it is the expensive part.**

**The backend fix checks out.** Last `No funds moved` in telemetry: **2026-09-16 06:19:15Z**. Since then, **23 completed top-ups across 10 distinct users, zero failures** — about 29 hours of real traffic. The daily shape backs it up (Sep 14: 37 failures / 3 users → Sep 15: 11 → Sep 16: 8 → Sep 17: 0).

**The three independent parts are worth shipping on their own**, and the precheck is the strongest of them: it fixes an **agent** error, not the backend one. Quoting `total_in_hyperliquid` (account value) as free capital over-states what a top-up can draw whenever a position is open, and no backend fix touches that. Same for the exact withdraw-all figure and fees-before-money-moves — neither has anything to do with the two-leg worker.

I would also keep the **"find the money before you say anything"** rule. It is harmless when the bug is absent (it only fires on a FAILED status), and "never say 'still in the funding wallet' without naming the balance — perps or Spot" is good practice for any failed money operation, not just this one.

---

### Change the second-FAILED remedy before pushing

`SKILL.md:205-206` and the say-table row at `:216`:

> a second FAILED for the same strategy means **this strategy wallet cannot receive top-ups right now** (some older strategy wallets refuse the second leg every time; **a fresh deploy gets a wallet that accepts it**)

> "This strategy wallet can't receive top-ups right now. Two options: **deploy it fresh (new wallet)**, or Senpi support with request id …"

That is a diagnosis of the bug you have just fixed, and the remedy it prescribes costs real money: a fresh deploy is a new wallet-creation fee, and #648 is in this same batch precisely because an agent burned **$3 of a $24 account** on that reflex. It also leaves the old strategy running, so the user ends up with two.

With the worker fixed, a second identical FAILED is no longer evidence that this wallet is structurally bad — it is evidence something new is wrong, which is a support question, not a redeploy. Suggested shape: keep the stop, drop the redeploy option, and route to support with the request id. If you want to keep a line about pre-hardening wallets, date it and mark it as history rather than as a live remedy.

That is a small edit to one rule and one table row. Everything else in the PR stands, my approval at `8a73d11` still holds for the rest, and I am happy to re-approve after the rebase.

betashop:
Thanks. Agreed on all of it, and pushed as `411809a`, rebased onto current main (548c9a1).

**The remedy you flagged is gone.** You were right: `SKILL.md` rule 4 and the say-table row still prescribed a fresh deploy after a second FAILED, and the needle test pinned that text. Rule 4 now reads: a second FAILED means something is wrong that a retry won't fix; apply rule 3 again, say where the money is, and hand the top-up request id to Senpi support. It also says **never offer a fresh deploy as the fix for a failed top-up**, because that costs another wallet-creation fee and leaves the old strategy running. The table row now says so too. I didn't keep a dated history line, since incidents stay out of the skill text.

**Guard.** `test_a_second_failed_top_up_goes_to_support_never_to_a_fresh_deploy` pins the new rule and fails on any of the old remedy's phrasings. Against the previous `SKILL.md` it fails; restored, all 14 tests pass.

**Kept, as you suggested:** the find-the-money-first rule and "never say 'still in the funding wallet' without naming the balance".

**CI on the rebased branch:** every `run:` step passes; full suites 1,191 passed / 9 skipped. The PR description is updated to match: the precheck gates on `withdrawable`, and a second FAILED goes to support.

On the merge order you raised on #642: this branch and #642 each add one directory to the same `tests.yml` pytest line. Whichever merges first, the other needs that one-line rebase.

PR #648, opened 2026-09-16
ops 3.19.0: diagnose and advise from the runtime's own numbers — runCount counts emits, stops in ROE and price, exposure an edit buys, runtime state before a top-up

betashop:
## What
`senpi-strategy-ops` 3.19.0 — four rules that make the agent diagnose and advise from the runtime's own numbers, each riding the bullet it belongs to.
- **`runCount` counts signals emitted, not ticks.** A fresh heartbeat with `runCount: 0` is a live scanner that found nothing — never a fault, never a reason to close and recreate (each new wallet costs a real creation fee); the fix for "it isn't trading" is the gates or the budget, applied in place. Sub-minute scanner intervals on a small book are refused with the fee arithmetic, not a timing claim.
- **Say every stop twice — as ROE and as price.** Price distance is ROE ÷ leverage (a 15% ROE stop at 5× sits 3% from entry); quote both from the recipe, never a computed table.
- **Exposure an edit buys.** `max_entries_per_day × marginPct` is the share of equity a day may commit (4 × 20% = 80%, where the venue starts warning); above ~60% say so and get a fresh yes, and never present the warning that follows as a market event.
- **Read the runtime's own state first.** A `Runtime paused: Max Entries/Day` line or a gate not `OPEN` means nothing opens today at any balance; free margin for a perp book is the perps `withdrawable`, never spot-side USDC.

## Why
Two sessions this week. In one, an agent read a health check showing a two-minute-old heartbeat and `runCount: 0` as a runtime fault and rebuilt the strategy twice by delete-and-recreate — three wallet-creation fees on a $24 account, two of them for nothing — while the runtime log showed a hundred overnight ticks with nothing passing the gates; the same agent used the in-place update correctly six times later that morning. In the other, a stop-distance table printed ~20% at 5× for a 15% ROE stop (it is 3%) and sold a leverage upgrade; a `marginPct` edit from 15 to 20 with a four-entries-a-day cap built the 80% margin warning that followed; and sixty seconds after the runtime had paused itself at 4/4, the advice offered was a top-up "for a fifth position". The liveness reference already carries the `runCount` rule; the skill body did not, and nothing said the other three.

## Size — read this before "inside the budget"
The skill body stays at **331 lines, exactly the ceiling** in `test_skill_surface.py` (`BODY_BUDGET["senpi-strategy-ops"] = 331`), because every rule rides an existing bullet. But the resident text grew **31,202 → 32,775 characters (+1,573, roughly +393 tokens)**. The guard counts lines; the cost it exists to control is tokens. So "inside the budget" is true by the only metric that is checked, and not by the one that matters. A character-based guard would make the number honest again — a one-line change, but it resets every skill's budget, so it belongs in its own PR.

## Verification
- `senpi-strategy-ops/tests`: 419 passed, 1 skipped. The new test is a `TestCase` above the `__main__` guard, so it runs under both `pytest` and `python3 test_skill_serr124.py`; removing any one rule fails it under both. Every needle is absent from the pre-change skill.
- The CI skills suite: 1091 passed, 9 skipped; vendor parity OK.
- README row synced to 3.19.0.


shnoodles:
**Review — four good rules, each traced to a real session. One structural note the body should own, and a test placement nit.**

**Intent ✓.** The `runCount` rule is the valuable one: an agent read a two-minute-old heartbeat with `runCount: 0` as a fault and delete-and-recreated twice — **three wallet-creation fees on a $24 account, two for nothing** — while the log showed a hundred overnight ticks with nothing passing the gates. Saying "counts signals EMITTED, not ticks" plus "never a reason to close and recreate (each new strategy wallet costs a real creation fee)" closes both the misreading and the expensive reflex.

`ROE ÷ leverage` is the other one I'd keep above all: a stop table off by the leverage factor **sold a leverage upgrade**, which is a user losing money to a formatting error. Stating it as an identity with a worked pair (15% ROE at 5× = 3% from entry) is the right shape — "quote both from the recipe, never a computed table" is what stops it recurring.

---

### The line budget did not see this change, and that is now a pattern

The body says the skill "stays inside its line budget". True, and precisely:

```
BODY_BUDGET = {"senpi-strategy-ops": 331, ...}
ops body on this branch: 331 lines     ← exactly at the ceiling, zero slack
```

But the resident text grew **31,202 → 32,775 chars (+1,573, ~+393 tokens)** — because all four rules ride existing bullets rather than adding lines. The guard counts **lines**; the cost is **tokens**.

This is the second time in two weeks (#611 did the same, +305 chars on one table cell). Neither PR did anything wrong — a budget that measures lines rewards exactly this, and riding the relevant bullet is genuinely better placement than a new section. But ops is now at its ceiling by the only metric that is checked while the thing the ceiling exists to control has grown ~400 tokens since the budget was set.

Not a change request for this PR. Worth one sentence in the body saying the growth is in characters, so the next reviewer is not reassured by "inside the budget" — and worth someone switching that guard to characters, which is a one-line change in `test_skill_surface.py` and would make the number honest again.

### Ponytail (diff-scoped)

The new test is appended **after** the `if __name__ == "__main__": unittest.main()` guard, as a bare module-level function with its own `import os` inside the body — while every other test in the file is a method on the `unittest.TestCase` above it. pytest collects it either way, so it passes; but `python3 test_skill_serr124.py` will not run it, which is the invocation the file's own `__main__` block exists to support. Move it into the class (it needs no fixtures) and drop the inline import.

Eight needles, of which `"Say every stop twice"` and `"ROE ÷ leverage"` are the two that must never silently vanish. `"Max Entries/Day"` already appears elsewhere in the file, so that one does not discriminate.

### Verified locally
Branch @ pr-648, 0 behind main: full CI **1,091 passed, 9 skipped, 0 failed, 64 subtests**. Version 3.18.0 → 3.19.0 consistent across title, README and SKILL.md. (Body says `senpi-strategy-ops/tests`: 419 passed — I see the aggregate only, no disagreement.)

betashop:
Both points taken.

**The size numbers reproduce exactly.** Body without frontmatter: **31,202 → 32,775 characters (+1,573)**, **331 → 331 lines** against `BODY_BUDGET["senpi-strategy-ops"] = 331`. The PR description no longer says "inside its line budget". It now has a section stating the characters grew while the line count held at the ceiling, so the next reviewer isn't reassured by the wrong metric. I agree the guard should count characters. That resets every skill's budget, so it belongs in its own PR, and I'm glad to open it.

**The test — fixed in f45ef8b.** Confirmed first: `python3 test_skill_serr124.py -v` ran it **0 times**. It's now a `TestCase` above the `__main__` guard, reading through the file's existing `SKILL` path, with no inline import. One divergence from your suggestion: I gave it its own class, `DiagnoseFromTheRuntimesOwnNumbers`, rather than filing it under `Serr124IsThisAccountsPoolNotARetry`. It isn't about SERR124, and a class name is documentation.

**`Max Entries/Day` — dropped.** Checked it: the string appears **once on `main`** already, so it couldn't detect the rule's removal. The other seven needles are each absent on `main` and present on the branch.

Proof: removing `ROE ÷ leverage` from the skill fails the test under **both** runners — `pytest` 1 failed, `python3 <file>` exit 1. Ops suite 419 passed, 1 skipped.

shnoodles (approved):
Approved at `f45ef8bf`. The test is now a TestCase above the `__main__` guard and actually runs under `python3 <file>` (verified: 4 tests, OK) — its own class name is the right call. Dropping the `Max Entries/Day` needle was correct, it could not discriminate. Thanks for putting the character growth in the body rather than leaning on "inside the line budget"; the character-budget change belongs in its own PR as you say.

PR #649, opened 2026-09-16
ci: run every strategies/<id>/tests suite — 37 directories that never ran

betashop:
## What
CI now runs every `strategies/<id>/tests` suite — 37 directories that have never run in CI.

A new script, `.github/scripts/run_strategy_package_tests.sh`, runs them one process per directory and gates on every exit code. A new workflow step calls it, placed before the existing full-suites step.

## Why
Raised by review on #643: its 20 new tests are real, but nothing keeps them true. `tests.yml` runs `strategies/tests` (the catalog-wide guards), never `strategies/<id>/tests`. The workflow's own comment explains the gap — a module-level `sys.exit()` in the chimp family aborts collection for the whole tree, so the tree was named narrowly and every per-strategy suite fell out.

Two things stop the obvious one-line fix of listing the directories:
- **They cannot share a pytest session.** Many use the same basenames (`test_scan.py`, `test_scoring.py`) with no `__init__.py`, and each puts its own `scanners/` on `sys.path`, so `pytest strategies/*/tests` stops with an `INTERNALERROR` in collection. One process per directory avoids both.
- **Eight files are script-style, not pytest-style.** All in the chimp family (chimp, gibbon, gorilla, orangutan): checks run at import and the file ends `sys.exit(1 if FAIL else 0)`. pytest cannot import them. The script runs each with `python3 <file>` and gates on its exit code.

## The traps, checked
- **A bare `python3 file` can be a silent green** (the workflow comment warns about exactly this). Not here: both chimp files end `sys.exit(1 if FAIL else 0)`. Proven by breaking chimp's `scoring.py` — the file exits 1.
- **Classification is by content, not by a list.** A file defining `def test_` or `class Test` is pytest-style; anything else is script-style. On `main`: 42 pytest-style files, 8 script-style, no directory mixes the two. A new strategy needs no CI edit.
- **A missing pytest fails loudly.** Without the preflight, a runner lacking pytest reports every pytest-style directory as a test failure. It now stops with exit 2 and names the interpreter.

## Verification
Run exactly as CI runs it (`bash .github/scripts/run_strategy_package_tests.sh`, with a pytest-capable `python3`):

| case | result |
|---|---|
| clean `main` | exit 0 — 37 directories green, ~5 s |
| break `strategies/aegis/main/scanners/scoring.py` | exit 1 — `FAILED: strategies/athena/tests (pytest exit 1)` |
| break `strategies/chimp/main/scanners/scoring.py` | exit 1 — `FAILED: strategies/chimp/tests/test_engine.py (exit 1)` |
| no pytest on `python3` | exit 2 — names the interpreter, runs nothing |

**One finding from the aegis mutation, for a follow-up:** raising the high-conviction multiplier from 1.5× to 9.9× was caught by athena's drift guard, not by aegis's own suite. `test_margin_pct_for`'s only high-conviction case sits at a regime where the 15% cap clips both values to the same answer, so the 1.5× branch is never exercised uncapped. The runner is doing its job; aegis's test has a blind spot.


shnoodles:
**Review — it does what it says. I tested the gate in both directions.**

This is the PR that closes the finding I raised on #643 (20 new tests landing in a tree CI never ran), and the classification approach — `def test_` / `class Test` → pytest per directory, anything else → `python3 <file>` — is the right call given the `strategies/*` tree genuinely contains both styles.

**Verified, not assumed:**

| check | result |
|---|---|
| green on `main` | **exit 0**, `strategy package suites: 37 directories` |
| all 37 actually run | yes — per-directory `::group::` blocks, script-style ones reporting e.g. `26 passed, 0 failed` |
| red on a broken **pytest-style** suite | appended a failing test to `strategies/aegis/tests` → **exit 1** |
| red on a broken **script-style** suite | forced `sys.exit(1)` in `strategies/chimp/tests/test_package_shape.py` → **exit 1** |

Both styles gate. That was the thing worth proving, since a runner that loops over directories and swallows a non-zero exit is worse than no runner at all — it reports coverage it does not have.

**Your correction to my #643 comment stands too:** the chimp family was never failing, it is script-style and pytest simply cannot import it. I had called it a blocker; "invisible to CI" is the accurate description, and all 37 were green on `main` the whole time.

Two small notes, neither blocking:

- The runner is `bash`, and my first attempt to exercise it ran it under `python3` (my error). Worth a `#!/usr/bin/env bash` shebang plus the executable bit if it does not have them, so `./` invocation cannot pick the wrong interpreter.
- Once this lands, #642's `test_scan_noop.py` in `strategies/signals/tests/` is covered by this runner, and #643's 20 aegis/athena tests by the same. Those two PRs' verification stories become real rather than local — worth saying so in their bodies when they rebase.

Nothing else from me.

betashop:
Checked both notes.

- **Shebang and mode.** The runner already starts with `#!/usr/bin/env bash` and is committed as `100755` (`git ls-tree` on this branch), so `./.github/scripts/run_strategy_package_tests.sh` runs under bash. No change needed.
- **Coverage for #642 and #643.** This holds. The runner loops over `strategies/*/tests`, so `strategies/signals/tests` from #642 and the aegis and athena suites from #643 run as soon as this is on main. Both descriptions now say so.

**Merge order.** `git merge-tree` shows #647 conflicting with this PR and with #642 on `.github/workflows/tests.yml`, while #642 and this PR merge clean with each other. #647 goes first, then this PR and #642 rebase.

betashop:
Update to my note above: #642 no longer carries `strategies/signals` (the host moved to v2), so once this merges the runner picks up #643's aegis and athena suites and nothing from #642.

shnoodles (approved):
Approved at `e33e8cf7`. I tested the gate in both directions rather than the happy path: green on main (exit 0, 37 directories), exit 1 on a broken pytest-style suite, exit 1 on a broken script-style suite. That was the thing worth proving. The workflow comment now names all 8 files across 4 packages. This closes the gap I raised on #643 and makes #642's and #643's verification stories real.

PR #650, opened 2026-09-16
improve-trades 1.10.0: a closed trade's side comes from its label, never the size's sign

betashop:
## What
`senpi-improve-trades` 1.10.0 fixes two things:
- **Side from the label only.** A closed trade's side now comes only from its side label: `type` ("Close Long" / "Close Short"), or `dir` / `side` / `direction` / `positionSide` containing "long" / "short". A row without a label is unknown.
- **No verdict for an unknown side.** The if-held comparison now gives an unknown side no dollar figure and no verdict. Before, it treated an unknown side as a long.

## Why
The discovery reader never read `type`, which is where `discovery_get_trader_history` puts the side. It fell back to the sign of `szi`, and at close `szi` is positive on both sides. So every closed short read as a long, and its if-held dollar figure and verdict had the wrong sign. This is the same defect #644 fixes in senpi-portfolio. The on-chain fills path already reads `Close Long` / `Close Short` correctly, so it is unchanged.

The feed shape, from production telemetry of `discovery_get_trader_history` results:

| window (UTC) | result |
|---|---|
| Sep 15 17:00 to Sep 16 17:00 | 649,058 closed-row `szi` values, 0 negative |
| Sep 16 14:00 to 17:00 | 95,409 closed rows, every one typed `Close Short` or `Close Long` |

The MCP guide documented the wrong shape: a signed `szi`, and `type` as an order type. Senpi-ai/senpi-hyperliquid-mcp#216 corrects it.

## Tests
- **Fixtures.** The seven closed rows in the two review fixtures now use the real shape, with the side in `type` and an unsigned `szi`. They previously carried signed sizes and no `type`, the shape the guide described.
- **New test.** Labels decide the side. Four unlabelled rows come back unknown: a signed size, an unsigned size, a row whose PnL and price move disagree in sign, and `side: "sell"`. `_if_held` returns no figure and an `unknown` verdict for an unknown side.
- **Old reader against the new fixtures.** Main's `review.py` fails 6 tests, including `test_short_sign_is_direction_adjusted` with `assert 'long' == 'short'`.
- **Old counterfactual.** Restoring the unknown-as-long line fails the new test.
- **Suites.** `senpi-improve-trades/tests`: 72 passed, up from 71. The CI pytest command gives 1,090 passed, 9 skipped, exit 0. Vendor parity and the min-budget golden both exit 0.
- **Versions.** The README row and SKILL.md are synced to 1.10.0.
- **Docstring (d1d3d24).** The docstring no longer says fees can flip realized PnL. On a closed row, `realizedPnl` is gross: it equals `szi × (exit − entry) × side`, and fees sit in `totalFees`. Nothing else is read because the label is on every row. No behaviour change.

## Worth knowing
- **CI never runs these tests.** `senpi-improve-trades/tests` is not in the CI pytest command, and neither is `senpi-smart-money/tests`. Adding them edits the same line as #642 and #647, so it is a follow-up once those merge.
- **README conflict with #645.** Its version row is on the line next to this one. Whichever merges second needs a one-line rebase.


shnoodles (approved):
Approved. This is the fix for the live bug I reproduced while rechecking #644 — verified here against a real-shaped row: closed SHORT now reads `short` (it returned `long` on main), closed LONG reads `long`, no label returns `None`. `_direction` is now label-only, matching portfolio's shape.

You also caught a downstream consequence I missed: `_if_held` previously did `-raw_move if direction == "short" else raw_move`, so an **unknown** side was silently treated as a long and got a dollar figure and a verdict. It now returns no figure and no verdict. That was the more dangerous half — a wrong side is visible, an invented counterfactual is not.

72 improve-trades tests, full CI 1,090 passed / 0 failed. Version 1.9.0 → 1.10.0 consistent.

PR #651, opened 2026-09-16
phalanx 1.3.3 / athena 1.1.3: read the 4h board from the nested response the MCP returns

betashop:
## What
`phalanx` 1.3.3 and `athena` 1.1.3. The 4h leaderboard read that feeds Phalanx's divergence booster now reads the response shape the MCP actually returns. Athena's Phalanx leg is byte-identical to Phalanx, as the Athena drift test requires, so both change together.

## Why
`_crowd_lean` accepted the board only as a flat list under `data.markets`. The MCP's `leaderboard_get_markets` handler calls `createResponse(tool, 'read', { markets })`, where `markets` is the board client's own `{markets: [...], source_trader_count, window, timestamp}`. So live rows arrive at `data.markets.markets`:

| check | result |
|---|---|
| production `leaderboard_get_markets` responses, last 24 h (Sep 16) | 46 of 46 nested, 0 flat |
| `_crowd_lean` from `main` on the nested response | `{}` |
| `_crowd_lean` from `main` on a flat response | `{'ETH': 88.99}` |

Because the read returned `{}` on every real tick, `crowd_lr = crowd_lean.get(asset_key, 50.0)` was always 50.0. `scoring.is_divergent` counts 50 as SHORT, since only `> 50` is LONG. So the booster did not stay off. It fired on every cohort-LONG candidate and never on a SHORT one:

| cohort long % | `is_divergent(x, 50.0)` on `main` before this PR | multiplier |
|---|---|---|
| 80, 60 | True | 1.5 |
| 40, 20 | False | 1.0 |

Every long also carried the reason "divergence: cohort LONG vs 4h board SHORT", which was false. Conviction feeds `margin_pct_for`, which gives base × 1.5 at 80 or above and × 1.25 at 65 or above, so longs were also sized up. The Barnacle template already unwraps this response twice, which confirms the shape.

## Behaviour change
Once a runtime runs this version, the 1.5× goes only where the proven cohort really disagrees with the 4h board. Behaviour changes both ways. A long loses the boost, and the larger size that came with it, whenever the board is also net long, so some longs rank lower or don't qualify. A short now gains the boost when the board is net long. Entries, ranking and sizing can all change. New deploys pick this up immediately, because `deploy.py` fetches the current package from `main`. Running Phalanx and Athena runtimes keep the old code until they're updated.

## Verification
- New `strategies/phalanx/tests/test_crowd_lean.py` covers the real nested shape and the flat shape. With `main`'s `scan.py`, the nested test fails. The fix passes both.
- `strategies/phalanx/tests`: 42 passed. `strategies/athena/tests`: 3 passed, drift test included.
- CI pytest command: 1,090 passed, 9 skipped, exit 0. Vendor parity OK. Min-budget golden OK, 109 templates.
- `validate_strategy.py`: ✓ phalanx v1.3.3, ✓ athena v1.1.3.
- Catalogs regenerated with `gen_catalog.py --updated 2026-09-16`. Only the two versions and the date changed.

Merge note: #643 also bumps `strategies/athena/strategy.yaml`, from 1.1.2 to 1.2.0. It needs a one-line rebase after this lands.


betashop:
**Correction to the description, now edited.** Before this fix the divergence booster did not "never fire". The empty board read made `crowd_lr` default to 50.0, which `is_divergent` counts as SHORT. So every cohort-LONG candidate got the 1.5× conviction boost, plus the larger margin that comes with conviction 65 or above, and no SHORT candidate ever did. Repro on the parent commit:

```
python3 -c "import sys; sys.path.insert(0,'strategies/phalanx/main/scanners'); import scoring; print([scoring.is_divergent(x, {}.get('ETH', 50.0)) for x in (80,60,40,20)])"
[True, True, False, False]
```

So the fix changes live behaviour in both directions. Longs lose the boost when the 4h board is also net long. Shorts gain it when the board is net long. Found by the parallel scanner audit and verified independently.

PR #652, opened 2026-09-16
improve-trades 1.11.0: book_vs_market reads the live leaderboard_get_markets envelope

betashop:
`book_vs_market` (the "what did I miss / compare me to the market" read) comes back empty on every live call. `leaderboard_get_markets` nests its payload under the same key it uses, so after `_ok()` strips `data`, `_extract_movers` receives `{"markets": {...}}` instead of a list and returns `[]`. This PR steps into the inner payload when `markets` is a dict. A flat payload still reads as before.

## Evidence

| Check | Result |
|---|---|
| Response shape | The MCP handler wraps its client's payload, which is already `{markets: [rows], source_trader_count, window, timestamp}`, under `markets` again. Live result: `{"success": true, "data": {"markets": {"markets": [rows], "source_trader_count": N, "window": "4h", "timestamp": T}}}` |
| Production tool-call results, last 24 h | `leaderboard_get_markets`: 47 of 47 nested (`"data":{"markets":{"markets":[`), 0 flat, 0 failed. `leaderboard_get_momentum_events`: 27 of 27 nested (`"data":{"events":{"events":[`) |
| `book_vs_market` on `main`, live envelope | 0 movers, 0 gaps, `window: null` |
| `book_vs_market` on `main`, flat payload (the shape the test fixture records) | 4 movers, 4 gaps, `window: 4h` |
| This branch, both shapes | 4 movers, 4 gaps, `window: 4h` |

## Repro (repo root, offline)

```python
import json, sys
sys.path.insert(0, "senpi-improve-trades/scripts")
import review
bare = json.load(open("senpi-improve-trades/tests/fixtures/review_fixture.json"))["leaderboard_get_markets"]
live = {"success": True, "data": {"markets": bare}}   # the shape the MCP returns
for label, resp in (("bare payload", bare), ("live envelope", live)):
    bvm = review.book_vs_market(review._FixtureClient({"leaderboard_get_markets": resp}), [], [], {}, True)
    print(f"{label:14} movers={len(bvm['top_movers'])} gaps={len(bvm['gaps'])} window={bvm['window']}")
```

On `main`, the `live envelope` line prints `movers=0 gaps=0 window=None`. On this branch both lines print `movers=4 gaps=4 window=4h`.

## Audit of every reader (counts)

`git grep -l 'leaderboard_get_markets\|leaderboard_get_momentum_events' -- 'strategies/*.py' 'senpi-*/scripts/*.py'`, with tests excluded, lists **72 files**. I ran each of their 62 reader functions against both the live nested shape and the flat shape, using a fake MCP client, and compared the outputs.

| Class | Count | Where |
|---|---|---|
| Works on the live shape (identical output for both shapes) | 57 functions in 56 `scan.py` files | every scanner board reader, plus the meerkat and rotator event readers. The phalanx and athena `_crowd_lean` copies count here since #651 |
| Passthrough, no row parse | 4 | `pulse.py` `fetch_smart_money`; `smartmoney.py` `fetch_near_term` and `_dry`; `review.py` `_dry`. The response goes to the agent as-is, and all rows and events are present in both shapes |
| Broken on the live shape | 1 | `review.py` `book_vs_market` (this PR) |
| No read (tool name only in a docstring or comment) | 13 | `scoring.py` in bald-eagle, beaver, grizzly, hummingbird, hyena, jaguar, marlin, meerkat, osprey, scorpion, sheep, vulture, wolverine |

Before #651, `_crowd_lean` in phalanx and athena was broken as well.

## Live behavior change — Jason decides whether and when to merge

**This skill:** once installed agents auto-upgrade to 1.11.0, `review.py market` and `review.py all` start returning top movers, gaps and `smart_money_pct` where they returned nothing. Agents will surface "what did I miss" gaps they previously reported as empty, and users may act on them by changing or adding strategies. The skill places no trades, and its no-performance-chasing guardrails still apply.

**The same defect in trading templates (#651, merged):** Phalanx and Athena conviction is affected, so entries, ranking and sizing change for every account running them once it upgrades. On `main` before #651, the crowd read was `{}`, so `crowd_lr` defaulted to 50.0, and `scoring.is_divergent` counts 50 as SHORT. As a result, every cohort-LONG candidate got the 1.5x divergence conviction, which can also lift it into a larger margin tier through `margin_pct_for`. No cohort-SHORT candidate got it. You can reproduce this from the repo root. The command below prints `[True, True, False, False]`, which is what `is_divergent` returns for cohort long ratios of 80, 60, 40 and 20 when the crowd read is empty.

```bash
python3 -c "import sys; sys.path.insert(0,'strategies/phalanx/main/scanners'); import scoring; print([scoring.is_divergent(x, {}.get('ETH', 50.0)) for x in (80,60,40,20)])"
```

With a real read, longs lose the boost whenever the 4h board is also net long, and shorts gain it whenever the board is net long.

## Version

- `senpi-improve-trades` goes from 1.9.0 to **1.11.0**, a MINOR bump. Installed agents auto-upgrade within major 1. The README row is synced.
- 1.10.0 is skipped on purpose because #650 already bumps this skill to 1.10.0. If both PRs used 1.10.0, git would merge the identical version edit without a conflict. Whichever fix landed second would then ship with no new version, and installed agents would never pull it. With different versions, whichever PR lands second conflicts on that line, and whoever resolves it should take the next unused minor.

## Tests

- `python3 -m pytest senpi-improve-trades/tests -q`: 72 passed. With `main`'s `review.py` restored, the new `test_book_vs_market_reads_the_live_envelope` fails (`assert ([])`) and the other 63 pass.
- CI command from `.github/workflows/tests.yml`: vendor parity OK; min-budget golden OK (109 templates); pytest 1162 passed, 9 skipped. That is 1090 from the existing list plus this suite's 72.
- Run locally on Python 3.9; CI uses 3.11.

## Second commit: CI runs this suite

`senpi-improve-trades/tests` was not in the CI pytest list, so the regression test above would only ever run by hand. The commit adds one path to that list, and the suite collects cleanly alongside the others. Drop the commit if this suite should stay out of CI.

## Ponytail pass

- The fix unwraps once at the call site rather than teaching `_extract_movers` two shapes, so one place restores both the rows and `window`.
- The test reuses the recorded fixture wrapped in the live envelope, so no new fixture file.
- I cut a separate `window` assert, because comparing against the flat-payload result already covers it.


shnoodles (approved):
Approved on content, with one merge-order condition below.

The envelope fix is right and matches the same nesting bug #653 documents in the phalanx engine — `data.markets = {markets: [...]}`. The test is the good kind: it wraps the **recorded fixture** in the live envelope and asserts the result is identical to the bare payload, so it proves equivalence rather than pinning a shape. Adding `senpi-improve-trades/tests` to CI continues the right pattern.

**Needs a rebase before it can merge.** It is branched from main (1.9.0 → 1.11.0) while #650 sets 1.10.0, and I verified the collision: merge #650, then #652 → **CONFLICT**. After rebasing onto #650 the 1.11.0 number is correct.

Also note **five open PRs now edit `.github/workflows/tests.yml`** — #642, #645, #647, #649 and this one. Tested pairwise: 647→649 conflicts, 647→652 conflicts, 649→652 is clean. Whichever lands first, the rest need a rebase; worth picking an order deliberately rather than discovering it.

betashop:
Merge order, checked with `git merge-tree` (each PR merged onto main, then the other on top; both orders tried) against main at the time of this comment.

**Your three pairs hold.** #647→#649 conflicts, #647→#652 conflicts and #649→#652 is clean. #650 then #652 also conflicts (`README.md`, `senpi-improve-trades/SKILL.md`).

**Every conflicting pair conflicts in both orders:**
- `.github/workflows/tests.yml`: #642↔#647, #642↔#652, #647↔#649, #647↔#652
- `README.md`: #645↔#650, #645↔#652, #650↔#652

**Clean in both orders:** #642↔#645, #642↔#649, #642↔#650, #645↔#647, #645↔#649, #647↔#650, #649↔#650, #649↔#652.

No three of the approved five (#645, #647, #649, #650, #652) are mutually clean, so any order needs at least three rebases. One order that hits that minimum is #649 → #645 → #647 (rebase: tests.yml) → #650 (rebase: README) → #652 (rebase onto #650: README, improve-trades SKILL.md, tests.yml). #642 is not approved yet and conflicts with #647 and #652, so it takes one rebase after those land.

PR #653, opened 2026-09-16
phalanx / athena: the board-read comment says what the bug actually did (comment only)

betashop:
## What
A comment-only correction in `strategies/phalanx/main/scanners/scan.py` and `strategies/athena/phalanx/scanners/scan.py`, which stay byte-identical as the Athena drift test requires. There is no behaviour change and no version bump.

## Why
The comment added in #651 says the divergence booster "never fired". It did fire, on every long. With the board read empty, `crowd_lr` defaulted to 50.0, which `scoring.is_divergent` counts as SHORT, so every cohort-LONG candidate scored as divergent and no SHORT did. #651's description was corrected at the time, but this comment wasn't. Agents read it when they explain the update to users, and one has already repeated the wrong version.

## Why no version bump
The bytes that run are unchanged. A bump would send every Athena and Phalanx user an "update available" for a comment. It would also re-conflict #643, which moves Athena to 1.2.0.

## Verification
`strategies/phalanx/tests` 42 passed, and `strategies/athena/tests` 3 passed, drift test included. The CI pytest command gives 1,090 passed, 9 skipped, exit 0. Min-budget golden is OK.


shnoodles (approved):
Approved. Comment-only, and the correction matters: "the crowd read defaulted to 50 (counted as SHORT), so every long scored as divergent and no short ever did" says what the bug actually produced, where the old wording only said the booster never fired. That is the difference between a reader knowing the signal was absent and knowing it was systematically wrong in one direction.

Verified the two legs are still byte-identical after the edit (`cmp` on `strategies/phalanx/main/scanners/scan.py` vs `strategies/athena/phalanx/scanners/scan.py`) — which is the thing a two-file comment change can quietly break.

PR #654, opened 2026-09-16
why 1.3.0: "you're just ChatGPT" and "prove it makes money" get a real answer

betashop:
## What
`senpi-why` 1.3.0:
- **Triggers.** The description now also covers "how are you different from ChatGPT / Claude / any AI", "why not just ask ChatGPT", and "where's the proof you actually make people money", in any language.
- **"Senpi is your AI quant".** The one-line answer and a new ChatGPT section lead with it. Each capability names the skill that shows it. It reads the whole market (senpi-market-pulse). It finds who is making money on Hyperliquid right now and spots the divergence (senpi-smart-money, and senpi-signals where installed). It turns that into actionable ideas, codes the strategy (senpi-strategy-author), and runs it with the risk managed. A chatbot can't read live Hyperliquid data, won't code and run a strategy, and won't manage risk. The agent then offers one of those reads live, using the skills every box has, and copy trading is never the answer.
- **"Show me it actually makes people money".** A second section answers with functionality only. It never discusses Senpi users' profits or returns: no leaderboard of Senpi strategies, no top strategy's returns, and no user's own PnL offered as proof. Showing which Hyperliquid traders are profitable right now stays allowed, because that is market data. One clause says Senpi doesn't promise profits, then comes what it does and the live read.
- **README row** synced to 1.3.0.

## Why
An agent was asked, in Ukrainian, what makes it different from ChatGPT and where one could see it actually making people money. The positioning skill never loaded, because its triggers named other apps and bots but not chatbots or proof requests. The agent answered from `discovery_get_top_strategies` instead: a table of other users' copy-trading returns led by the best one. It gave a thin list of what ChatGPT can't do, and closed by suggesting the user copy the top strategy as "a proven result". That is survivorship, someone else's money, and an implied promise, and it frames Senpi as copy trading.

## Verification
Prose only. Frontmatter parses, and the description is 742 characters. The CI pytest command gives 1,090 passed, 9 skipped, exit 0. Vendor parity and min-budget golden are OK.


shnoodles (approved):
Approved. I looked at this one specifically because "prove it makes money" is a question where a positioning skill could easily ship a claim — and it does the opposite. *"Answer with what Senpi does, never with profits"*, no leaderboard of Senpi strategies, no user's PnL as proof, and one plain clause that Senpi does not promise outcomes. The distinction that showing which Hyperliquid traders are profitable is market data rather than a claim about Senpi is exactly the right line to draw.

"Show it rather than claim it — offer to run one of those reads now" is the part that makes the section work.

One light note, not a blocker: *"What Cursor and Claude are for code, Senpi is for trading"* names two third-party products in user-facing copy. It reads as a simile rather than a comparison claim, but it is a marketing decision sitting in a skill file, so it is worth Jason seeing it once.

Full CI 1,090 passed / 0 failed.

PR #655, opened 2026-09-16
otter 1.2.2 / pangolin 1.2.2 / raptor 1.1.4 / owl 1.3.2: the 4h board read keeps each token's dominant side, not the last row

betashop:
## What was wrong

`leaderboard_get_markets` returns one row per (token, dex, direction), and the live feed orders rows by `pct_of_top_traders_gain` descending across the whole board. For a token with both a LONG row and a SHORT row, the minority side is therefore always listed last. Four board readers built their per-token map by overwriting on every row, so the map held the **minority** side:

| Template | Reader | Consumer | Fix |
|---|---|---|---|
| otter | `fetch_sm_map` | +2 `SM_ALIGNED` when the board side matches the OI flow and pct ≥ 5 | keep the `is_dominant_direction` row (the #612 pattern) |
| pangolin | `_get_sm_map` | `SM_*` points (+3 to −2) by side and pct, `SM_MOMENTUM_FADING` (the row's 15m contribution) | same |
| raptor | `fetch_sm_map` | `sm_gate`: the board side must match the whale's side, pct ≥ 2, ≥ 10 traders; `SM_STRONG`/`SM_ALIGNED` and `15M_SPIKE`/`15M_STALE` points | same |
| owl | `fetch_sm_positioning_map` | `score_crowding` tilt `\|long% − 50\|` > 20 → +3, > 12 → +1, `sm_confirms_funding` +1 | long share over both rows (below) |

Owl had two more defects. It scaled `pct_of_top_traders_gain` as a 0–1 long fraction (`pct * 100` for long rows, `(1 - pct) * 100` for short rows), producing values like −2463 and 410. Its xyz ban also checked for an `XYZ:` name prefix that the board never sends, because xyz rows arrive as a bare token plus `dex: "xyz"`.

**Owl's intended scale.** `score_crowding` treats `sm_long_pct` as a coin's long share on a 0–100 scale with 50 as neutral. The v5 scanner read exactly that as `longPct` / `pctOfGainsLong` (default 50). The live field is instead each side's share of the **whole board's** 4h gains. So the share is now `long / (long + short) * 100` over the coin's two rows, floored on the leading side's `trader_count`: the G2/G4 ratio pattern already used by lynx, dragonfly, barnacle and osprey. `trader_count` counts both sides; scoring does not read it.

## Evidence: live feed

Source: prod telemetry, every `leaderboard_get_markets` result recorded from 2026-09-02 to 2026-09-16 (agent tool calls to the same MCP tool the scanners use).
- Telemetry truncates each result near 8.7 KB (about the top 11 rows), so pair counts cover the visible top of each board only.
- Token names are masked, so a token's two rows were matched on `(dex, max_leverage, day_notional_volume, token_price_change_pct_4h)`.

| Check | Count |
|---|---|
| Results in the nested shape `data.markets.markets` | 601 / 601 successful (the other 3 of 604 were `UNAVAILABLE` errors; 0 flat) |
| Results whose rows are in `pct_of_top_traders_gain` descending order | 601 / 601 successful |
| Complete rows carrying `is_dominant_direction` | all. No result lacks it on more than one row, its last (truncated) row. |
| `dex` values | `""` 5,678 rows, `"xyz"` 1,503 rows |
| Two-sided tokens with both rows visible | 70 (in 66 results) |
| … dominant row listed before the minority row | **70 / 70** |
| … `is_dominant_direction` marks the higher `pct_of_top_traders_gain` | 70 / 70 |
| … dominant side has *fewer* traders than the minority side | 6 / 70 |
| … minority row clears the 10-trader floor | 70 / 70 |
| … minority row clears raptor's pct and trader floors (≥ 2 %, ≥ 10 traders) | 54 / 70 |
| … minority row clears otter's ≥ 5 % bonus threshold | 10 / 70 |

## Repro (executed)

Setup: main @ 4a69c9ee vs this branch. The fixture rows from `fixtures/leaderboard_get_markets.json` were re-nested in the live envelope, sorted pct-desc, plus a synthetic XRP pair shaped like a live one (dominant side has fewer traders). Each cell below is (direction, pct, traders), or owl's long share.

| Template | Token | Board: dominant · minority | main | this PR |
|---|---|---|---|---|
| otter / pangolin | BTC | LONG 5.52 / 112 · SHORT 0.91 / 17 | SHORT 0.91 / 17 | LONG 5.52 / 112 |
| otter / pangolin | ETH | SHORT 8.2 / 114 · LONG 1.1 / 21 | LONG 1.1 / 21 | SHORT 8.2 / 114 |
| otter / pangolin | SOL | LONG 3.6 / 69 · SHORT 1.3 / 26 | SHORT 1.3 / 26 | LONG 3.6 / 69 |
| raptor | HYPE | SHORT 25.63 / 212 · LONG 0.42 / 9 | LONG 0.42 / 9 | SHORT 25.63 / 212 |
| raptor | XRP | SHORT 3.95 / 44 · LONG 2.12 / 189 | whale LONG **confirmed**, whale SHORT refused | whale LONG refused, whale SHORT confirmed |
| owl | BTC | 5.52 L · 0.91 S | 9.0 (reads short-tilted) | 85.85 |
| owl | ETH | 1.1 L · 8.2 S | 110.0 | 11.83 |
| owl | HYPE | 0.42 L · 25.63 S | −2463.0 | 1.61 |
| owl | SOL | 3.6 L · 1.3 S | −30.0 | 73.47 |
| owl | NVDA / GOLD / SP500 (xyz) | — | 410.0 / −120.0 / 140.0 (leaked in) | absent |

Downstream, on one `scan()` tick with the same board:
- **pangolin:** the BTC funding fade scored 7 on main, with the SM reason `SM_CONFIRMS 0.9% (17t)` taken from the minority row. With this PR it scores 6, with `SM_SLIGHT_CROWD 5.5%` + `SM_MOMENTUM_FADING -6.84` from the dominant row.
- **owl:** `smCovered` goes 7 → 4.
- **owl, BTC with positive funding:** `score_crowding` said `sm_tilted_SHORT_9%` on main. It now says `sm_tilted_LONG_86%` + `sm_confirms_funding`.

## Tests and counts

- New tests in `strategies/tests/test_board_last_row_overwrite.py` use the live nested shape (`_live_board`): one per template. `test_board_momentum_rename_g4.py` now runs owl through the 4-case ratio probe (thin leading side refused, the ratio's own tilt kept), and its docstring moves owl into the ratio group.
- **On main's four `scan.py` files:** 5 failed, 21 passed (the 4 new tests + `test_board_read_refuses_a_thin_side[owl]`). **With this PR:** 26 passed.
- The exact `.github/workflows/tests.yml` command, each step gated on its exit code (local venv, Python 3.14; CI pins 3.11):
  - vendor parity: OK (rc 0)
  - min-budget golden: OK, 109 templates (rc 0)
  - pytest: **1094 passed**, 9 skipped (rc 0). Main: 1090 passed.
- **Per-package suites:** otter, pangolin, raptor and owl have no `tests/` directory, so there is no package pytest to run. Instead, a one-tick `scan()` smoke ran per package in its own process on main and on this branch; all exited 0. The smoke is shallow for two of them: otter stops at bootstrapping and raptor at no-traders. Raptor's full `scan()` path is covered by `test_raptor_direction_from_side.py`.
- **Catalogs:** regenerated with `gen_catalog.py --updated 2026-09-16 --branch main`. The diff is the four version lines, and both copies are identical. The caribou/hydra min_budget warnings are pre-existing on main.
- **Versions:** otter 1.2.2, pangolin 1.2.2, raptor 1.1.4, owl 1.3.2 (patch, as in #612/#651). No open PR bumps these four. The packages have no README or SKILL.md, so there is no version text to sync.
- **Ponytail pass:** the only cut was a history note in owl's docstring (it is in this body instead). otter, pangolin and raptor are the one-line #612 filter.

## ⚠️ This changes live trading behavior

The direction and tilt these templates read from the 4h board feed entries and ranking for **every account running otter, pangolin, raptor or owl**, once the package is deployed or upgraded to these versions:

- **otter:** the +2 bonus follows each token's dominant side. It can now fire where the dominant side matches the flow, and it no longer fires for a minority side ≥ 5 %.
- **pangolin:** the `SM_*` points and `SM_MOMENTUM_FADING` come from the dominant row, so candidate scores and the emitted pick can change.
- **raptor:** a whale is confirmed only when the board's dominant side matches, and the pct and 15m points come from that row. In the live sample, 54 / 70 minority rows cleared the pct and trader floors by themselves, so main could follow a whale on the minority side and refuse one on the dominant side.
- **owl:** the tilt points and `sm_confirms_funding` now come from each coin's long share. On main they came from one mis-scaled row, the last one listed, so which coins cross the crowding floor, and when, can change. xyz rows no longer enter the map.

One dependency: otter, pangolin and raptor rely on `is_dominant_direction`, as condor, lemon and kestrel have since #612 (present on every complete live row above). If the API dropped it, otter and pangolin would lose the board points and raptor would log "no smart-money data" and emit nothing.

Opened for review only: not merged. Jason decides whether and when to merge.


shnoodles (approved):
Approved. The bug is real and the fix is the right shape: the board returns LONG and SHORT as separate rows per token, and keeping the last row instead of the dominant side is exactly the class of error #653 documents in phalanx.

**`is_dominant_direction` verified against production**, since it appears nowhere in the MCP's TypeScript: it is present in **92 of 96** live `leaderboard_get_markets` results over the last 2 days. So the field is real and near-universal on the live payload.

The `long_pct` docstring rewrite is the valuable part — stating that it is long / (long + short) × 100 over the coin's two rows, not the raw board share, is what stops the next porter re-introducing this.

Verified: validators ✓ otter 1.2.2 / pangolin 1.2.2 / raptor 1.1.4 / owl 1.3.2, min-budget golden ✓ (109), catalogs regenerate to no diff, full CI 1,094 passed / 0 failed.

PR #656, opened 2026-09-17
author 3.10.0 + runtime 4.1.3: candle times are epoch ms, time.time() is seconds

betashop:
## What
This PR adds time units to the two places where authors learn the candle schema. It also pins the closed-trade unit that `risk-gates.md` already states.

| doc | change |
|---|---|
| `senpi-strategy-author/references/creating-a-strategy.md` (candle schema callout) | States that `t` (candle open) and `T` (candle close) are epoch **milliseconds** and `time.time()` is **seconds**. Gives the closed-candle check `_f(c["t"]) / 1000 + candle_seconds <= now`, where `candle_seconds` is the candle interval, not the scanner's `interval_seconds`, and `now = time.time()` comes from `scan.py`. Warns that comparing `t` to `time.time()` directly means no candle ever closes. |
| `senpi-trading-runtime/references/scan-contract.md` (market-data cast table) | States that `t`/`T` are epoch milliseconds on every runtime version, because the cast changes their type, not their unit. I checked this against the runtime scaffold's candle cast, which calls `float()` and never rescales. Gives the same closed-candle check. |
| `senpi-strategy-author/references/risk-gates.md` | No change. It already says `discovery_get_trader_history` `openTime`/`closeTime` are Unix epoch seconds, which agrees with the tool. The new test pins that sentence. |

I searched the rest of the repo's docs for a unit on either field and found no other statements.

**Versions**
- `senpi-strategy-author` 3.9.0 → **3.10.0** (minor)
- `senpi-trading-runtime` 4.1.2 → **4.1.3** (patch, the runtime-skill exception)

Both README rows change in the same commit. When this PR opened, no open PR or pushed branch claimed author 3.10.0 or runtime 4.1.3. The numbers assume no other bump to either skill merges first. If one does, this PR takes the next number.

## Why
Candle times and trade times use different units. Candle `t`/`T` are milliseconds. `time.time()` and trader-history times are seconds. Neither doc gave the candle unit.

- A scanner that compared candle `t` with `time.time()` never saw a closed candle. It never emitted, and nothing raised an error.
- A guide that gave the wrong unit for trader-history `closeTime` led readers to divide seconds by 1000. That silently turns off anything built on those times, such as a loss kill-switch or a cooldown.

The guide itself is fixed in Senpi-ai/senpi-hyperliquid-mcp#218. This PR puts the units where scanners get written.

## Verification
**New needle tests**
- `senpi-strategy-author/tests/test_reference_time_units.py` has 2 tests.
- `senpi-trading-runtime/tests/test_reference_candle_time_units.py` has 1 test.
- Both basenames are new across the full-suite invocation, so no import-file collisions.

**Mutation proof**
- On the `main` docs, the two candle tests fail. The risk-gates test passes because that text is unchanged.
- On the new docs, I broke each pinned statement one at a time. All 9 mutations fail their own assertion:
  - `creating-a-strategy.md`: drop the ms/seconds headline; say `t`/`T` are seconds; remove `/ 1000` from the check; delete the never-closes warning.
  - `risk-gates.md`: drop the trader-history field names; flip "seconds" to "milliseconds".
  - `scan-contract.md`: say `t`/`T` are seconds; remove the "type, not unit" clause; remove `/ 1000` from the check.
- With the docs restored, 3 of 3 pass.

**CI:** every `run:` step of `.github/workflows/tests.yml` except `pip install`. I ran them in a venv with pytest 9.1.1 and pyyaml 6.0.3 on Python 3.14, since the workflow's 3.11 is not installed locally, and gated on each exit code.

| step | `main` @ 548c9a12 | this branch |
|---|---|---|
| min_budget vendor parity | exit 0, `VENDOR PARITY OK` | exit 0, `VENDOR PARITY OK` |
| min_budget golden | exit 0, 109 templates match | exit 0, 109 templates match |
| account-status suite | exit 0, 16 passed | exit 0, 16 passed |
| strategy package suites | exit 0, 37 dirs: 302 passed, 8 script files exit 0 | exit 0, the same |
| full suites | exit 0, 1177 passed, 9 skipped | exit 0, **1180 passed**, 9 skipped |

No failures on `main` or on this branch.


shnoodles (approved):
**Approved.** Both units verified against live data:

- **Candles** (`market_get_asset_data`): `"t": 1789551000000`, `"T": 1789551899999` — **13 digits, milliseconds**.
- **Trader history** (`openTime`/`closeTime`): **10 digits, seconds** — confirmed in mcp #218.

So the product really does use both units, in two surfaces an author touches in the same scanner, and neither errors when confused. That is what makes this worth a rule rather than a fix.

The closed-candle formula is the valuable part — `_f(c["t"]) / 1000 + candle_seconds <= now`, with the explicit warning that `candle_seconds` is the **candle interval** (4h = 14400) and not the scanner's `interval_seconds`. Those are two separate ways to get this wrong and both are named.

Docstring naming both directions of the confusion in one place is the right call, since the failure is the mismatch rather than either unit. Pairs correctly with #218.

Full CI **1,108 passed / 0 failed**. Branch is 14 behind main — rebase before merge.

PR #657, opened 2026-09-17
senpi-signals 2.1.0 + market-pulse 1.4.0: one closing question, the signals brief in every pulse

betashop:
Stacked on #642 (senpi-signals 2.0.0). Retarget to `main` once #642 merges.

## What
Getting senpi-signals ready to go live for every user.

**senpi-signals 2.1.0**
- **Every run ends with one question.** Set up a trade on one read, or start a strategy: the user's own Athena, or one built around the reads.
  - **Trade (senpi-trade):** the agent shows margin, share of account, leverage, the stop and the liquidation price first. The stop must sit before liquidation, and nothing is placed without a yes to that exact order.
  - **Athena first (senpi-strategy-ops):** the lead strategy offer. Its Phalanx sleeve follows the same proven cohort the smart-money reads come from (≥ $1M lifetime realized). Ops runs its walkthrough and deploys it under the user's name; the minimum budget is read from the catalog.
  - **The read's own template, or the user's own:** each read type names the template built on that kind of read as the quick start: smart money vs the crowd → Athena or Phalanx; funding extreme → Pangolin or Camel; momentum event → Meerkat; laggard behind BTC → Mantis. Building from scratch (senpi-strategy-author) is its peer, with the guardrails: a DSL stop on every position, 3x or less, few trades, budget and creation fee stated before funding.
  - **Templates are starting points the user makes their own.** No promised results, never "proven", deploy only on a yes.
  - This replaces the opt-in "how could I play it?" follow-up, which conflicted with the rule to add no closing notes after the feed.
- **Legend:** the top band reads "strongest" instead of "act on it", because the feed is an observation.
- **`sweep.py --brief N`:** the same sweep, printed as a title plus the top N trade reads, one line each (`score.render_brief`). No legend, no news feed, and nothing about the engine. A quiet market prints one plain line. It's for another skill to close with.
- **Self-contained:** the sweep no longer imports from another skill's folder. `scripts/smartmoney.py` and `scripts/mcp_client.py` are byte-identical copies of senpi-smart-money's, so the skill runs with only itself installed. `tests/test_vendored_parity.py` fails if a copy drifts.
- **Triggers:** added for the chat-chip wording ("scan Senpi Signals", "scan for market anomalies").

**senpi-market-pulse 1.4.0**
- **Brief:** every run, full or narrow, closes with the Senpi Signals brief (`sweep.py --brief 3`, about 10 s).
- **One question:** "Want the full Senpi Signals sweep? I can also check how your positions sit in this market, or build a strategy for it." When the smart-money layer is live, it adds the mirror clause.
- **Missing skill:** a box without senpi-signals skips the step silently.
- **Replaces** the verbatim three-CTA list; the positions and strategy routes are kept. The strategy route offers the user's own Athena first, with a strategy built from the pulse's thesis as its peer.
- **The 4h board is no longer called "smart money."** From a live run on M401059 (Sep 17) where the pulse's "Smart Money" section had ZEC and HYPE long while the signals brief under it had smart money short both: they are different populations — `leaderboard_get_markets` (who is winning in the last 4h, survivorship included) versus the >= $1M lifetime-realized cohort — and the SKILL told the agent to narrate the board as "the >$1M cohort". It is now **4h leaders** everywhere the reader sees it (the JSON key is unchanged), with the distinction stated. Same rename the 8 board-reading templates got in #636-#640.
- **A missing move prints `-`, never `0.00%`.** The same run showed NIFTY, VIX and DXY at exactly 0.00%; the engine returns `null` for a move it could not read, so a flat number invents an observation.
- **The question no longer offers to mirror "one of these smart-money traders."** Four hours of gains is not a track record, and the offer read as a recommendation. Mirroring stays available when the user asks, routed through senpi-trader-research on the trader's record.
- **`render_brief` says when nothing is strong.** Two yellow reads (63.4, 57.9) under "top reads right now" is the best of a weak lot; below the orange band the header now reads "nothing strong right now, context only" and still prints the reads.

## Why
Signals is going live for every user, through a chat chip and inside every market pulse. The end of a run is where a reader decides what to do. Now one consistent, consented next step is offered, and the trade and strategy paths carry the risk rules. The self-contained engine removes a path where the sweep exits before reading anything if another skill is missing or installed elsewhere.

## Verification
- `senpi-signals/tests` + `senpi-market-pulse/tests`: 69 passed.
- **20 new tests**, each proven to fail against the previous code or text:
  - brief output and ordering
  - quiet market
  - failed-source line
  - vendored parity
  - no cross-skill reach
  - legend
  - closing question
  - read-only rule
  - triggers
  - market-pulse brief command, closing and skip clause
  - Athena named first, and every template the closing names exists in `strategies/catalog.json`
  - no gains claims in the closing ("proven template", "guaranteed", "profitable", …)
  - market-pulse's strategy route offers Athena before a scratch build
  - the 4h-leader layer is never called smart money, and the old ">$1M cohort" example is gone
  - a missing move is not printed as flat
  - the question does not offer to mirror a 4h winner
  - the brief's header names the band when nothing clears the orange line
- **Full CI steps** on the branch: vendor parity ✓, min-budget golden ✓, account-status ✓, strategy package suites ✓, and the full suites at 1,237 passed / 9 skipped.

## Go-live order
1. Merge #642, then this PR (retarget to `main` first).
2. Add senpi-signals to the runtime skills manifest (senpi-trading-runtime `src/skills-manager/skills-manifest.json`, `maxMajor: 2`). This happens only after both are on `main`, or the coordinator 404s.
3. Verify a sweep on a live agent.
4. Then the Quick Actions chip migration.


senpi-signals/scripts/mcp_client.py
```diff
@@ -0,0 +1,158 @@
...
+        status, _raw, _ct, sid = _post(self.url, body, self._headers(), timeout)
+        if status >= 400:
+            raise MCPError(f"initialize HTTP {status}")
+        self._sid = sid
+        # streamable-http requires notifications/initialized after init
+        note = {"jsonrpc": "2.0", "method": "notifications/initialized", "params": {}}
+        _post(self.url, note, self._headers(sid), timeout)
+        self._initialized = True
+
+    def mcp_call(self, tool, timeout=12, **arguments):
```
0xsarvesh, line 151: **S1 · reliability — zero retries, anywhere in the skill**

```
$ grep -rn "retry|retries|backoff|Thread|asyncio|concurrent" scripts/*.py
→ (none)
```

`mcp_call` is a single attempt. One transient failure and that source is dark for the whole run — and because `gather()` short-circuits when the universe read fails (sweep.py, flagged on #642), a single 503 on `market_list_instruments` produces a zero-information run that still **exits 0** and prints "nothing notable stands out right now."

Every call this client makes is a read, so they are all trivially safe to retry. I'd treat this as the cheapest single reliability win in the PR.

**Fix:** 2 retries with jittered backoff on connection errors and 5xx; leave 4xx alone.

senpi-signals/scripts/mcp_client.py
```diff
@@ -0,0 +1,158 @@
...
+class MCPError(Exception):
+    pass
+
+
+def _post(url, body, headers, timeout):
+    """POST a JSON body on a fresh connection; return (status, raw_bytes, content_type, session_id)."""
+    parts = urlsplit(url)
+    scheme = parts.scheme
+    host = parts.hostname or ""
+    port = parts.port or (443 if scheme == "https" else 80)
```
0xsarvesh, line 34: **S2 · reliability/performance — a fresh TLS connection per call, and no concurrency**

`_post` opens and closes a new connection every call, so a sweep pays **eight full TLS handshakes**. Combined with strictly sequential execution in `sweep.gather()`, worst case is roughly:

```
24s (initialize) + 12 + 20 + 3×20 + 3×12  ≈  152s
```

The three `discovery_get_trader_state` calls are independent 50-wallet batches and trivially parallelizable — that alone is ~40–60s of the tail, free.

This matters more in this PR than in #642: senpi-signals' own SKILL.md records that the 1.x cron *"timed out on every firing."* The cron was removed but the latency cause wasn't addressed, and **#657 now puts that tail inside every market pulse** — `--brief 3` costs the identical 8 reads as the full sweep to print three lines.

**Fix:** keep one connection for the process, thread-pool the trader-state batches, and add an overall wall-clock budget that degrades to whatever landed rather than hanging the pulse.

senpi-signals/scripts/smartmoney.py
```diff
@@ -0,0 +1,575 @@
...
+    thousands of ranks deeper — page by offset until both are sampled or the page drops below the
+    crowd floor."""
+    smart, crowd, seen = [], [], set()
+    pages = 0
+    for page in range(MAX_PAGES):
+        try:
+            resp = client.mcp_call("discovery_get_top_traders", time_frame="ALL_TIME",
+                                   sort_by="PROFIT_AND_LOSS_REALIZED", open_position_filter=False,
+                                   limit=PAGE_SIZE, offset=page * PAGE_SIZE, timeout=20)
+        except Exception as e:  # noqa
```
0xsarvesh, line 145: **S3 · reliability — the cohort depends on an undeclared sort order**

The docstring states *"The ranking is DESC by realized"*, but the call passes `sort_by` with no sort **direction**. If that server-side default ever flips, `smart` comes back empty and the skill silently degrades to OI/funding/price — reported as `NO DATA`, indistinguishable from a dead service.

Worth an explicit guard on the first page (assert the first row's realized PnL >= the last row's) so a contract change surfaces as an error rather than as a quietly worse product.

Related, on the caller side: `sweep.cohort()` sets `sm.MAX_PAGES = COHORT_PAGES`, mutating this module's global. Harmless today, a footgun the moment anything else imports `smartmoney` in the same process.

senpi-market-pulse/SKILL.md
```diff
@@ -135,47 +138,63 @@ Top-down, always this shape:
...
+the engine returns `null` when it could not read a move (a closed market, a row that failed), and printing that as
+flat invents an observation the data never made. If `smart_money` is present, add a short **4h leaders** note
+(e.g. "in the last 4h, 22% of the winners' gains sit in ZEC longs, across 228 traders") — it's high-signal.
+**Never call it smart money.** That layer is `leaderboard_get_markets`: who is winning *right now*, survivorship
+included. senpi-signals' "smart money" is the >= $1M lifetime-realized cohort, and the two are regularly on
+opposite sides of the same name in the same answer — so the words have to say which population each one is.
+
+## Mandatory closing: Senpi Signals in brief, then one question
+
+Every market-pulse run — a full read or a narrow ask — ends the same way, after the bottom line (or
```
0xsarvesh, line 155: **S3 · maintainability — this reintroduces the coupling the rest of the PR just removed**

Vendoring `smartmoney.py`/`mcp_client.py` with a byte-parity test is the right call, and it exists specifically so senpi-signals never reaches into another skill's folder. `cd ../senpi-signals` then re-asserts, in the prompt layer, that both skills are siblings in one root.

The graceful-skip clause below covers **absence** (and is properly pinned by `test_a_box_without_senpi_signals_skips_the_step_silently`), but not a different layout — there, the `cd` succeeds or fails in ways the agent isn't told how to handle.

Separately, worth being explicit about the cost being added here: **every market pulse now triggers a full 8-read sweep** to print three lines, with the ~150s worst-case tail noted on `mcp_client.py`. If the brief is meant to be cheap, it needs its own lighter path or a short-TTL cache — right now it is the most expensive closing section in the skill set.

This is also what makes the unlocked shared `out_dir` (flagged on #642, `sweep.py:383`) reachable: before this PR there was one caller, now there are two on the same box writing the same `current.json`/`signals.md`.

senpi-signals/tests/test_signals_skill_text.py
```diff
@@ -47,3 +48,78 @@ def test_compare_detectors_are_v2_and_never_claimed():
...
+
+
+def test_every_run_ends_with_one_question_to_trade_or_build():
+    """The feed is observation; acting is a separate, consented step. One question closes every run,
+    it routes to senpi-trade or to a strategy, and nothing is placed or funded without a yes."""
+    skill = _flat(SKILL)
+    assert "## How every run ends — one question" in skill
+    assert ("Want to act on any of these? I can set up a trade on one of them — you see the size and the stop "
+            "before anything is placed — or start a strategy: your own Athena, the smart-money hedge fund, "
+            "or one built around these reads.") in skill
```
0xsarvesh, line 60: **S3 · test quality — prose pinning makes SKILL.md unrefactorable**

This file asserts ~30 exact strings. Some are genuinely valuable and should stay — the guardrails here (`"Place nothing until the user says yes to that exact order."`, `"The stop must sit before liquidation."`) and especially `test_the_strategy_offer_leads_with_athena_and_every_template_it_names_is_real`, which cross-checks every named template id against `strategies/catalog.json`. That one validates a real integration and would catch a genuine bug.

But assertions like `"two runs a minute apart give the same feed"` and `"## v2 — compare over periods (not in 2.0)"` pin turns of phrase, so any editorial pass turns CI red for reasons unrelated to behaviour. The doc's redundancy becomes load-bearing for the test suite.

Concretely: I compacted SKILL.md by 22% in #671 and had to preserve several sentences **verbatim** purely to keep this file green — not because the behaviour needed them.

**Suggestion:** keep the safety and catalog assertions as exact strings; convert the phrasing ones to intent checks.

0xsarvesh (commented):
## Review — senpi-signals 2.1.0 + market-pulse 1.4.0

Reviewed against its base branch as asked. Ran the suite on the stacked tree: **69 passed, 0 failed**. The 2.1 changes themselves are small and good — `render_brief` is clean, and vendoring `smartmoney.py`/`mcp_client.py` behind `test_vendored_parity.py` is the right way to make the skill standalone.

**My main concern with this PR is not what it changes, it's what it activates.**

#642 has one caller of the sweep. #657 adds a second — every market pulse, fleet-wide — and that turns two latent problems into reachable ones:

- **The unlocked shared `out_dir`** (`sweep.py:383`, flagged on #642) becomes a live race the moment a pulse and a chip tap overlap. I reproduced it: one run led its feed with an asset it never gathered.
- **The ~150s sequential worst case** with no retries, no connection reuse and no deadline now sits inside the pulse path, for a three-line closing section that costs the same 8 MCP reads as the full sweep.

Neither is caused by this PR, but both are *scaled* by it, so they're worth resolving before it lands rather than after.

| # | Sev | Where | Finding |
|---|---|---|---|
| 1 | **S1** | `mcp_client.py:151` | Zero retries anywhere; a single 503 yields a zero-information run that exits 0 |
| 2 | **S2** | `mcp_client.py:29` | Fresh TLS connection per call, fully sequential, no deadline — ~150s worst case, now inside every pulse |
| 3 | **S3** | `smartmoney.py:138` | Cohort depends on an undeclared sort direction; a contract change degrades silently |
| 4 | **S3** | `market-pulse/SKILL.md:152` | `cd ../senpi-signals` reintroduces the cross-skill coupling vendoring just removed; also the real cost of the brief |
| 5 | **S3** | `test_signals_skill_text.py:58` | ~30 pinned prose strings make SKILL.md unrefactorable |

**S1 = fix before merge · S2 = fix before it's load-bearing · S3 = should fix**

The S1s I found in the engine itself (missing-price rendering as "price already ran — late"; the universe SPOF; the ⛔ engine banner reaching users verbatim) are inline on **#642**, since that's where those lines are introduced.

### On SKILL.md

Body is 3,034 words, description 212 — mid-pack for this repo (portfolio is 11,486), so not an outlier by volume, but misallocated. The closing-question section (505w, 17%) is larger than the circularity trap + what-one-reading-shows + two-lenses combined, and the same fact is stated up to six times (*one reading / no compare* ×6, *observation not advice* ×4, *say no to a cron* ×3).

The structural issue is that it's written as a changelog against a version nobody has — "2.0 / not in 2.0 / v2" framing exists to stop the agent claiming history, which is one rule, not a running commentary. It also carries two internal contradictions: rule 7 ("always state direction, never omit it") vs. `funding_extreme`'s deliberate `direction=None`, and the engine-talk convention vs. `score.py`'s coverage banner, which violates it in the user-facing path.

I've put a concrete version up as **#671** (stacked on this branch): body 3,034 → 2,358, description 212 → 172, **all 69 tests passing unchanged** because every pinned string is preserved. Text only, no behavioural constraint dropped. Take it, cherry-pick it, or close it — no attachment to it landing.

Leaving this as a comment rather than a blocking review. Happy to re-review once the S1s here and on #642 are addressed.

senpi-signals/tests/test_vendored_parity.py
```diff
@@ -0,0 +1,28 @@
...
+
+Run: python3 -m pytest senpi-signals/tests -q
+"""
+from pathlib import Path
+
+import pytest
+
+ROOT = Path(__file__).resolve().parents[2]
+SIGNALS = ROOT / "senpi-signals" / "scripts"
+SMART_MONEY = ROOT / "senpi-smart-money" / "scripts"
```
0xsarvesh, line 16: **S3 · maintainability — this proves source-tree parity, not deployed parity**

`ROOT` is the monorepo, so this guarantees the copies are identical **in CI**. The docstring's promise — *"a fix to either lands in both"* — holds in the repo and not on a box.

The two skills version and install independently: senpi-smart-money is at `1.3.0`, senpi-signals at `2.1.0`. The moment senpi-smart-money ships a cohort-engine fix as 1.4.0, an agent runs the corrected engine under `senpi-smart-money` and the stale vendored copy under `senpi-signals` — **two different answers to "where is smart money positioned" on the same box, from the same data, with nothing anywhere signalling drift.**

That's the standard cost of vendoring and may well be the right trade; the point is that nothing currently surfaces it, and the cohort engine is what the whole skill's credibility rests on. Options, cheapest first:

- Stamp the source version into the vendored header and report it in `coverage`.
- Fail CI when senpi-smart-money's version bumps without a matching senpi-signals bump.
- Extract the engine into a shared package both skills depend on.

More a decision to make explicitly than a bug to fix.

0xsarvesh (commented):
### Re-review against the current head (7a17b77) — three S1s verified fixed

Came back to add a couple of follow-ups and found the branch had moved under me, so I re-ran every repro against the current head rather than reporting from stale notes. **78 tests pass** (up from 69).

**Fixed, and I re-verified each one:**

| Finding | Fix | Verification |
|---|---|---|
| **S1** missing 4h price rendered as "price already ran — late" | `EARLINESS_PRICE_UNREAD = 0.35` + a dedicated `"price not read this run"` branch in `trade_read` | Unknown price now sits below a confirmed read and above a contradicted one, with its own sentence. The `not d` case also gets `"no side resolved"` rather than falling into a price band — that's the latent rule-7 gap too. |
| **S1** universe SPOF → outage reads as a calm market | `READ_ATTEMPTS`/`RETRY_SLEEP_S` retry loop + `universe_is_dark()` guarding **both** `feed_text` and `brief_text` | Outage now prints *"could not read the market this run … This is an outage, not a quiet market."* — in the pulse brief as well, which was the path I was most worried about. |
| **S1** shared unlocked `out_dir` → cross-run contamination | per-run `.current.<pid>.<uuid>.json` + atomic `os.replace` publish | **My repro no longer reproduces:** `ONLYINB appears in the feed run A PRESENTED TO ITS USER: False` (was `True`). |

The retry addressing the SPOF and the atomic publish addressing the race are both cleaner than what I suggested — nice.

**Still open from the earlier passes** (all inline on #642, where the lines live; none are in this PR's diff so I can't anchor them here):

- **S2** `score.py:901` — the ⛔ coverage banner still reaches `--print-feed` output verbatim, naming `senpi-smart-money` to the end user, against this skill's own *"never about the engine"* convention. Re-confirmed on this head: the cohort-dead feed still renders it. One-line fix — gate on `consumer == "social"` like the posting reminder four lines above.
- **S2** `crowd_source` is still never read by any renderer, so "vs the crowd" reads identically whether it came from the 4h board or from the sign of the funding rate.
- **S3/S4** `momentum_event` still has no staleness gate while `whale_move` does; `_traders_of` still can't distinguish an unparseable payload from an empty one; tied cohorts still resolve to LONG.
- **S2** `mcp_client.py` is unchanged, so reads are still fully sequential with a fresh TLS connection per call. The retry landed a layer up in `sweep.py`, which covers the reliability half — the ~150s worst case and the three parallelizable `discovery_get_trader_state` batches are still there, and that tail now sits inside every pulse.

One new finding inline on the vendored-parity test, which is more a decision to make than a bug.

For what it's worth: turning three S1s around this fast, with the dark-universe guard extended to the brief without being asked, is a better response than the review deserved.

shnoodles (approved):
**Approved.** Both of Sarvesh's S1 findings are fixed on this branch, and the fixes are the right shape.

**The shared-directory race is genuinely closed.** `run()` now hands `score.py` files only that run can see — `.current.<pid>.<uuid8>.json` / `.signals.<pid>.<uuid8>.md` — ranks from those, and publishes both under the stable names with `os.replace` (atomic), cleaning up in `finally`. `feed_text()` prefers the in-memory `result.feed_md`, so the normal path never round-trips through the shared file at all; the disk read survives only as a documented fallback for a hand-built `rep`.

And the regression test is **his exact scenario**, which is the best possible answer to a reviewer who proved a bug:

```
ONLYINB in B's universe            ✓
ONLYINB never in A's universe      ✓
ONLYINB not in A's ranked result   ✓
ONLYINB not in A's presented feed  ✓
```

That was the finding that mattered most, because this PR is what made the race reachable — it adds market-pulse's `--brief 3` as a second concurrent caller of the same default dir. Closing it here rather than deferring is the correct call.

**The universe SPOF is handled at the right level.** `universe_is_dark()` + `OUTAGE_LINE`, with the reasoning stated where someone will read it: *"A total outage would therefore be reported to the user as a calm market. It is an error, not a feed."* That is the precise trap — the skill teaches that a quiet read is a correct answer, so a dead universe was indistinguishable from a calm market. Special-casing the one read everything hangs off, while letting the optional lenses keep degrading with a not-measured line, is the right asymmetry.

**Vendoring guarded.** `smartmoney.py` (+575) and `mcp_client.py` (+158) are copies, and `test_vendored_parity.py` is what keeps them honest — same pattern as the min-budget vendor parity check.

Stacking is correct: based on `feat/senpi-signals-scanner`, so it merges after #642.

0xsarvesh (commented):
## Re-review — stack verified on `29616c6`

Re-ran the suites and my repros against the rebased head. **80 passed.** Everything I raised on #642 carries through correctly here, and I verified the two things that only matter on this PR:

- **The dark-universe guard reaches the brief.** `universe_is_dark` gates `feed_text`, `brief_text` and the exit code (3 on both paths). Confirmed the pulse closing now prints *"could not read the market this run … This is an outage, not a quiet market."* instead of *"nothing notable stands out right now."* That was my biggest single worry about this PR, since it makes the brief close every pulse fleet-wide — it's properly closed.
- **The weak-lot header.** `render_brief` now says *"nothing strong right now, context only"* when the top score is under 65, rather than captioning two 🟡 lines as "top reads". Didn't ask for it; it's the right call — a header that promises strength over a weak lot is the same class of overstatement as the ones the golden rules police.

### Still open on this PR, unchanged and not blocking

- **S2** `mcp_client.py` is unchanged, so reads remain fully sequential with a fresh TLS connection per call. The retry landed in `sweep.py`, which covers reliability — but the ~150s tail is now ~200s with `READ_ATTEMPTS = 2`, there's still no wall-clock deadline, and **this PR is what puts that tail inside every market pulse**. The three `discovery_get_trader_state` batches are independent and parallelizable; that alone is most of it.
- **S2** the retry doesn't cover the four `discovery_*` reads (detailed on #642, `sweep.py:209`) — relevant here because the cohort lens going dark is what the brief's degraded path depends on.
- **S3** `test_vendored_parity.py` proves source-tree parity, not deployed parity (inline above). Now that the vendored engine is what a live agent actually runs, this is the one I'd most like a decision on rather than a fix.

### Verdict

I'm satisfied with this PR. The SKILL.md side is resolved (#671 merged, duplication caught, occurrence-counting test added — good catch on my rebase artifact). Nothing outstanding here is user-visible or a correctness bug.

My one ask before step 1 is on #642, not here: **finding 4, the ⛔ banner** — one line, and it's the last place the output tells a user something untrue. Everything else I'm happy to see as follow-ups.

0xsarvesh (commented):
## Round 3 — verified on `0f85955` (carries `a67b68d`)

**82 passed.** Confirmed this branch already contains the banner gate and the all-8-read retry from #642, so nothing is pending a rebase.

Re-verified the two things specific to this PR:

- **Dark-universe guard reaches the brief** — `universe_is_dark` gates `feed_text`, `brief_text` and the exit code. The pulse closing prints the outage wording, not "nothing notable stands out". That was my main worry about this PR and it's properly closed.
- **Weak-lot header** — `render_brief` says "nothing strong right now, context only" under 65 rather than captioning 🟡 lines as "top reads". Unprompted and correct.

### Verdict: LGTM

Every S1/S2 is closed across the stack. Follow-ups recorded on #642, plus the three that are specific to this PR and unchanged — none blocking:

- **657-2** `mcp_client.py` still opens a fresh TLS connection per call and reads are fully sequential. Retries made reliability right; the ~200s tail is still there with no deadline, and this PR is what puts it in every pulse. The three `discovery_get_trader_state` batches are independent and parallelizable.
- **657-3** `build_cohorts` still assumes a DESC sort it doesn't request.
- **657-6** vendored parity is proven in CI, not on a deployed agent — the decision I'd most like made explicitly, since the two skills version independently and the cohort engine is what the credibility rests on.

Happy for these to land after step 1.

0xsarvesh (approved):
**Approving.** `0f85955` — carries `a67b68d`, so the banner gate and all-8-read retry are in. 82 passed locally, CI green.

Verified the two things specific to this PR: the dark-universe guard reaches `brief_text` (so a failed sweep no longer closes a market pulse with *"nothing notable stands out"*), and `render_brief` captions a weak lot honestly rather than calling two 🟡 lines "top reads".

Disclosure: this branch carries `1e3ccaf0`, my own SKILL.md compaction from #671. Jason reviewed it independently and caught a rebase artifact I introduced, so it isn't self-merged unreviewed — flagging it for the record.

Open follow-ups specific to this PR (`mcp_client` connection reuse + no wall-clock deadline, the undeclared DESC sort, and parity being CI-only rather than deployed) are in my earlier comments. None blocking.

One thing to watch on rollout rather than in code: this PR makes a full 8-read sweep close **every** market pulse, fleet-wide. Worth confirming the MCP/RPC headroom before step 2 widens it to every agent — raising that separately.

PR #662, opened 2026-09-17
portfolio 1.26.0: closed trades carry a UTC time, and the record can arrive late

betashop:
## What
`senpi-portfolio` 1.26.0 — the closed block dates every close in UTC and prints the newest close it holds, and the SKILL says what to do when a close has not reached the record yet.
- **`closed.recent[].closed_at_utc`** (`YYYY-MM-DDTHH:MM:SSZ`), read from epoch seconds or milliseconds (1e12 or more is milliseconds, the same cut `senpi-improve-trades` uses). A value that is not a 10-digit second is `null`, and one `meta.warnings` line counts those rows; no date is made up. The raw `closed_time` stays beside it.
- **`closed.closed_record_newest_utc`**: the newest close the record holds. `strategy_groups[].totals` carries the newest across a strategy's wallets.
- **SKILL.md** (source-of-truth callout, a pointer from the closed-trades rule, the field reference): quote a close's UTC date and time from `closed_at_utc`, never a raw epoch. The closed record can arrive hours after a close, so a close the user saw may not be in it yet: say it has not reached trade history, show the live positions and account value, and never present older closes as today's or invent the missing P&L. "In profit" on an open position is unrealized; say so, and don't forecast the close.

## Why
`discovery_get_trader_history` sends `closeTime` in epoch seconds, and the engine passed it through as a bare number. Converting that by eye is how an older close gets reported as today's. The record is also built by an ingest pipeline that can trail a close by hours, and a fresh read does not skip that, so the newest close in the record and the newest close on the wallet can differ.

## Verification
- `senpi-portfolio/tests/test_closed_stats.py`: seconds, milliseconds and a numeric string date the same instant; unreadable times (text, zero, a flag, an 11-digit number, a missing field) are `null` with a warning; the newest close wins whatever the row order or unit; a failed or empty read prints `null`; a strategy group reports the newest across its wallets; one needle test per new SKILL rule.
- Mutation check, 17 breaks, each fails the suite and passes once restored: seconds read as milliseconds, milliseconds read as seconds, no range check, no warning, first row taken as newest, group newest as the oldest or dropped, the freshness key missing on a failed read, and each of the 9 rule phrases removed.
- CI, every `run:` step of `tests.yml` run locally: vendor parity, min-budget golden, account-status suite, strategy package suites, full suites (1184 passed, 9 skipped). All exit 0.
- Version 1.25.0 → 1.26.0, README row synced. The open PRs that touch the `senpi-portfolio` version bump it to versions older than main, so 1.26.0 is unused.


shnoodles (approved):
**Approved.** I tested `_utc_iso` against the values it will actually see rather than the fixtures:

| input | result |
|---|---|
| `1789391168` (real `closeTime`, 10-digit seconds) | `2026-09-14T13:06:08Z` ✓ |
| `1789551000000` (13-digit ms) | `2026-09-16T09:30:00Z` ✓ |
| `0` / negative / text / `None` | `None` — no date invented ✓ |

The 1e12 cut is the right discriminator and it is consistent with what #218 and #656 establish this week: trader-history is epoch **seconds**, candles are **milliseconds**. Three PRs now agree on the units, which is the first time that has been true.

The second rule is the one I would keep above all: **the closed record can arrive hours after a close.** Exposing `closed_record_newest_utc` means the agent can say *"the record only goes up to X"* instead of presenting older closes as today's — that is the actual failure, and a bare epoch read by eye is exactly how it happens. "Show the live state from this read instead — open `positions[]` and `account_value`" gives it something true to say rather than only a prohibition.

Keeping the raw `closed_time` beside `closed_at_utc` is right too: the formatted field is what to quote, the raw one stays for anything that needs to compute.

`senpi-portfolio/tests` **99 passed**, full CI **1,112 passed / 0 failed**. Branch is 14 behind main — rebase before merge.

PR #663, opened 2026-09-17
runtime 4.1.4 + ops 3.20.0: an edit is live only after it is applied

betashop:
## The rule

An edit to a running strategy is not live until it is applied.

- `openclaw senpi update` without `--apply` only plans. Only the first line of its output says so (`Dry run for <runtime_id> — nothing has been applied.`), so never pipe `openclaw senpi` output through `tail` or `head`.
- An edit is live only once `--apply` exits `0` (first line `Updated <runtime_id>.`, no warning that the external scanners are not running) and the running strategy shows it: `openclaw senpi events -r <runtime_id> --name runtime.updated` lists the apply, and a re-run plan reports `The recipe is unchanged.`
- Until then, tell the user the edit is **saved, not applied**.
- Re-running `create` / `senpi deploy` on a running strategy applies nothing: it keeps the runtime that is running.

## Where it lands

- **senpi-trading-runtime 4.1.4** (patch)
  - `SKILL.md`: the rule, next to the `update` commands.
  - `references/scan-contract.md`: a short section after the `ctx.state` rules. Every key of every state record is saved as JSON. A whole tick (state and signals) is discarded on an exception, a timeout, a non-list return, a value `json.dumps` rejects, or `[]` after a failed `ctx.senpi_mcp` call. `scan.py` is imported once per scanner process. `senpi validate` runs the copy on disk and never inspects the running scanner.
- **senpi-strategy-ops 3.20.0** (minor)
  - `SKILL.md`: four resident lines. The body budget in `test_skill_surface.py` goes from 331 to 335, with a comment saying why.
  - `references/editing-a-live-strategy.md`: a step 4 with the two read-only checks, plus what counts as a successful apply.
- README rows are synced. Runtime 4.1.4 follows #656, which takes 4.1.3.

Each claim was checked against the runtime release source. The state, discard and load-once behaviour was also run against the release scaffold.

## Tests

- `senpi-trading-runtime/tests/test_edit_is_live_only_after_apply.py` and `senpi-strategy-ops/tests/test_skill_edit_not_live_until_applied.py` hold 37 needles. Whitespace is collapsed before matching, so rewrapping a line cannot break a needle.
- Every needle is missing from `main` and appears exactly once on this branch.
- Every needle was mutation-checked. Deleting its phrase makes its test fail; restoring the phrase makes it pass.
- All five `run:` steps of `.github/workflows/tests.yml` pass locally with exit 0. The full-suite step reports 1181 passed and 9 skipped.


shnoodles (approved):
**Approved.** This is 37 phrases of CLI behaviour, so I checked the quoted strings against the runtime source rather than the PR body. They are exact:

| quoted | source |
|---|---|
| `Dry run for <runtime_id> — nothing has been applied.` | `src/runtime/update-report.ts:228` |
| `The recipe is unchanged.` | `update-report.ts:302` |
| `runtime.updated` | `src/runtime/lifecycle-events.ts:198` |

The runtime even has a comment at `update-report.ts:224` explaining that phrasing — *"'nothing has been applied', not 'nothing has changed': the reader has just edited their…"* — which is the same distinction this PR teaches. Nice to see both sides agree on why the wording matters.

**The `tail`/`head` rule is the sharpest thing here.** A dry-run marker that appears only in the first line, in a CLI whose output people habitually pipe, is a trap that produces exactly the reported failure: four users believing an edit was live. Naming the pipe explicitly is worth more than the rule it protects.

Also right: a running scanner keeps its old code until step 3 restarts it, so the step-4 recipe plan cannot see a code-only edit and `runtime.updated` is the proof instead. That is the distinction that makes "is my edit live?" answerable at all, and it lines up with #399's `recipeHash` proposal.

**The budget raise is done properly.** `BODY_BUDGET["senpi-strategy-ops"]` goes 331 → 335 with a three-line justification in the file's own convention, which is what that mechanism exists for — it makes the growth visible rather than forcing a cut. Body is exactly 335.

`senpi-strategy-ops/tests` **421 passed**, full CI **1,109 passed / 0 failed**. Branch is 14 behind main — rebase before merge.

PR #670, opened 2026-09-17
account-status 1.4.0: the plan catalog and the free-credit ladder are publishable, the balance still is not

shnoodles:
## Why

`senpi-account-status` could explain the credits meter but could not answer **"what does Senpi cost."** Its rule *"never quote a plan size"* was written when nothing in the skill carried the plans — so a catalog question got a shrug at a meter the skill also cannot read. Two different things were collapsed into one prohibition.

## What changed

**Split the two.** The plan catalog and the free-credit ladder are published facts and now live in a new **`## Plans & free credits`** section to quote from. The user's own figures — what is left, what they were charged, where their bar sits — still come from their subscription page, never from the skill.

The new section carries:

- Starter $25 / Pro $50 / Advanced $100 / Quant $200, with credits, bonus rate and points rate; **every plan includes every model**; trading fees separate, from 0.05%.
- Cycle rules: no rollover, no top-ups, no overage; **upgrade instant** (cycle restarts, unused credits credited against the price); **downgrade/cancel at the next renewal**; payment web-only.
- Free trial: activates on agent deploy, one-time, 14 days; strategies pause on expiry but **open positions are NOT auto-closed**. **The trial figure stays unpublished** — it has changed more than once and their page is the only source of truth.
- The ladder: **$315 across 12 milestones**, with **"Launch your first strategy" (+$10)** replacing the retired **"Fund your wallet" (+$5)** — plus not-retroactive, milestone credits expire at the next renewal, volume progress is not live, the bonus is never USDC, and **the card wins** any disagreement.

**The meter section keeps its no-figure discipline.** The existing test that forbids a dollar amount there is unchanged and still passes; the section now says what each plan includes is published *below*, while how much is **left** still belongs to the user's page. Two rows of the Say / Never-say table were re-scoped accordingly, and routing sends catalog questions ("what plans are there", "how do I get free credits", "where are my milestone credits") to the new section.

Also borrowed from the support KB: the in-app credit indicator surfaces **only when credits run low**, so a missing indicator is expected rather than a lost balance — a recurring ticket.

## Test

New `PlansAndFreeCreditsArePublished` class locks the catalog, the ladder, the card-wins rule and the still-unpublished trial figure in place; the existing credits-meter tests are untouched.

```
python3 -m pytest senpi-account-status/tests -q
20 passed
```

Version 1.3.0 → 1.4.0, README row updated to match (a test asserts they agree).

## Note on timing

The "Launch your first strategy" milestone goes live **2026-09-18**.

Sister PR: Senpi-ai/senpi-hyperliquid-mcp#220 puts the same catalog in the MCP overview guide.


betashop (approved):
Approved — good split: publishable catalog + ladder go into the skill, user-specific figures still owned by the subscription page. README/SKILL 1.3.0→1.4.0 in sync, tests carry the new invariants.

PR #673, opened 2026-09-18
quant-desk 1.4.0: the desk remembers whose book is whose, and knows when a wallet is not indexed

betashop:
Builds the address book, the whose-book resolution and the not-indexed capture together — they are one feature, and items 1, 2 and 6 of the MVP plan all sit on it.

## What this does and does not fix

```python
whose = "other" if a.other else "mine"     # before
```

A bare address was treated as the reader's own book, so reading a whale told the reader **they**
were leaking $6,770 a year and recommended **they** fix it.

**Be clear about the scope: on a FIRST read of a whale, that is unchanged.** The default is still
the reader's own book — deliberately, because the flagship path is a Hyperliquid trader pasting
their own address, and putting a claim question in front of that sits on the one moment the product
exists for. So the first read of a stranger's wallet still depends on the agent passing `--other`,
exactly as before.

What changes is the **second** read. The book remembers, so a whale read once as someone else's
stays someone else's on every bare re-run after that. That is a real narrowing of the window, not a
closing of it, and the remaining exposure is the first read of each new address.

## What changed

**Whose book it is is a lookup.** `resolve_whose()` — an explicit flag wins, then what the book
already knows, then the default. **An address the book has not seen is the reader's own**; an
address already read as someone else's stays someone else's. The book adds memory, not suspicion.

**Two tiers of "theirs", because they are not the same claim:**

| | how we know |
|---|---|
| `verified` | a wallet senpi issued |
| `claimed` | the reader typed it and said it was theirs — **unverifiable from a chat message** |
| `analyzed` | someone else's book they read |

A relationship never downgrades: an `--other` run on your own wallet must not make the desk forget it is yours.

**Not-indexed is detected, not guessed.** `indexed=False` means the public endpoints showed closed round trips in the window and senpi's index returned none **for the same window** — a contradiction between two sources. A quiet wallet stays `null`. Previously the desk fell through to public fills silently, and public fills miss TWAP slices, so a whale got a confident desk built on a fraction of their volume with nothing saying so.

It also makes the promise honest: *"I've flagged it to the team"* is only true because the address lands in the book. `--addresses` lists the not-yet-indexed set, which is also the tracking artifact. The count is read from `references/coverage.json` at run time, and the SKILL drops the precise figure once `as_of` is stale.

**The lateral offer.** Both follow-up banks go deeper on the *same* book, so nothing ever told a reader the desk reads any wallet on Hyperliquid — or, after an analyst run, that it reads theirs. "Find me traders worth reading" routes through the cohort, the leaderboard or `senpi-trader-research`; it never invents an address.

## Test

14 new tests. Six invariants mutation-proven — each reverted, each fails:

| mutation | caught by |
|---|---|
| a known-`analyzed` address allowed to resolve back to `mine` | `test_an_address_already_read_as_someone_elses_stays_someone_elses` |
| relationship allowed to downgrade | `test_a_relationship_never_downgrades` |
| junk addresses allowed into the book | `test_addresses_are_case_insensitive_and_junk_is_refused` |
| indexed detection removed | `test_indexed_is_a_contradiction_between_two_sources_not_a_guess` |
| `coverage.json` drifts from the SKILL's quoted figure | `test_the_skill_answers_a_not_indexed_wallet…` |

```
quant-desk: 66 passed
full CI:    1281 passed, 9 skipped
```

1.3.0 → 1.4.0, README row and `render.VERSION` updated together — that header stamp is how a stale install is spotted.

## Not in here

Backend storage (V1 in the plan — the book is local per box, which is per user). The renderer line-break fix, which is the other half of making the desk readable. Verified-wallet marking is implemented (`mark_verified`) but not yet wired to a `strategy_list` read.

---

**Revision (2026-09-19).** The three paragraphs above marked scope, `resolve_whose` and the mutation
row are rewritten. The original text claimed an unknown address resolved to *someone else's* and
headlined the $6,770 case as fixed; the code defaults to the reader's own book and the case is only
fixed on a re-read. Jason made that call mid-build and I did not carry it back into the description.
Caught in review by Ignas — the code was never wrong, the description was. A test still *named*
`test_an_unknown_address_is_not_the_readers` has also been renamed, since it asserted the old
default in its name and docstring while testing a different layer.


shnoodles:
Reviewed against `origin/main` in a clean worktree at `34800c5`. The code is good. **Not approving yet** — the PR description describes the opposite of what ships on the headline behaviour, and that is the thing a reviewer signs off on.

## The module itself holds up

I went through `addresses.py` line by line and tried to break it. The rank ladder only ever moves up, `mark_verified` can only upgrade, `save()` is mkstemp + `os.replace` so a half-written book can't be read, `load()` swallows a corrupt book instead of taking the desk down, and `norm()` lowercases before matching `^0x[0-9a-f]{40}$` so junk is refused and case is handled. Coverage claims reproduce exactly: **1281 passed, 9 skipped** on the `full suites` step, quant-desk 68.

The `indexed` tri-state is the real thing you say it is, and the guard that matters is there:

```python
elif public_closed and not meta.get("senpi_history_failed"):
    indexed = False
```

A read senpi *couldn't answer* stays `None` rather than becoming "not indexed". That distinction is the whole value of the field and it's correct.

## The description says the opposite of the code

Body:

> `resolve_whose()` — an explicit flag wins, otherwise the address book decides, and **an address the book does not know is someone else's**.

Code:

```python
if addr_book.relationship(book, addr) == addr_book.ANALYZED:
    return "other"
return "mine"                      # ← unknown address
```

Executed, on a book that has never seen the address:

```
book knows it?  None
resolve_whose(unknown address) -> 'mine'
```

And the mutation table cites `test_an_unknown_address_resolves_to_analyst_mode` — that test does not exist. What exists is `test_an_unseen_address_is_the_readers_own_book`, asserting `== "mine"`, with `test_an_address_already_read_as_someone_elses_stays_someone_elses` commenting `# before we know anything` on a WHALE address resolving to `"mine"`.

So the code is internally consistent and deliberately tested — this isn't a bug, it's a product decision you appear to have made partway through and not carried back into the description.

**Why it needs resolving rather than a description patch alone.** The PR opens with the harm:

> Reading a whale told the reader **they** were leaking $6,770 a year and recommended **they** fix it.

On a first read of a whale, that is unchanged. The book has never seen the address, so it resolves `"mine"` and the desk speaks in the second person about the whale's book — identical in effect to the `whose = "other" if a.other else "mine"` line you quote as the bug. The fix only bites on the **second** read. And the fragility you name as the reason the old approach couldn't hold —

> It worked only while the agent reliably turned the word "analyst" in a sentence into `--other`

— still fully applies to every first read, which is the only read most whales get.

The docstring makes a fair counter-case (don't put a question in front of the flagship moment), and that may well be the right call. But it's the opposite trade from the one the description argues for, so pick one deliberately:

- **Keep `mine`** — then rewrite the description: drop the "an address the book does not know is someone else's" claim, fix the mutation-table row, and say plainly that the first-read case still rests on the agent passing `--other`. The $6,770 story stops being the headline, because this doesn't fix it.
- **Switch to `other`** — then the description is already right, and the tests at lines 101 and 114 need to flip with it.

I don't think it's my call which, but shipping the description as written would put a safety property in the commit record that the code doesn't have.

## One small thing

`record(book, addr, relationship=None, …)` shadows the module-level `relationship(book, addr)`. Harmless today since `record` never calls it — but the next edit inside that function that wants the lookup gets a confusing `TypeError`. Rename the parameter to `rel` (which is what the call site at `desk.py:417` already calls it).

## Ponytail — diff only

Nothing to cut. Stdlib only, no dependency added, the book is a dict in a JSON file rather than anything cleverer, and `awaiting_index()` earns its place as the artifact behind the "I've flagged it to the team" promise. Good restraint on leaving backend storage out.

Flip the description or flip the default and I'll approve on sight — the rest of this is ready.

betashop:
Taking option 1 — keep `mine`, fix the description. You were right on all three counts, and the cause is that Jason changed the call mid-build ("the default behavior should be to analyze your own book, 2nd path is someone else's"); I changed the code and the tests and never carried it back into the body. The code was never wrong, the description was.

**Rewritten:**
- The `resolve_whose` paragraph now says an unseen address is the reader's own, and that the book adds memory rather than suspicion.
- The mutation row cites `test_an_address_already_read_as_someone_elses_stays_someone_elses`, which exists.
- The $6,770 story is no longer headlined as fixed. It now says plainly that a **first** read of a whale is unchanged and still rests on the agent passing `--other`, and that what this closes is the **second** read onwards — a narrowing of the window, not a closing of it.

**One thing you didn't have to find, in `d8bfef1`.** A test was still *named* `test_an_unknown_address_is_not_the_readers`, with a docstring calling that "the default the whole feature exists for". It passes, because it exercises the book layer (`ab.is_mine` on an unrecorded address) rather than the desk's resolution — but the name and docstring assert the opposite of what ships, which is exactly how the next reader concludes the default is `other`. Renamed to what it guarantees, and the docstring now points at `resolve_whose` as where the product default lives.

quant-desk: 68 passed.

PR #674, opened 2026-09-18
senpi-signals 2.2.0 + smart-money 1.3.1: the timeout a read asks for reaches the socket, the sweep has a deadline, and an empty cohort is not a token verdict

danielmbirochi:
## The mechanism

`senpi-signals/scripts/sweep.py`:

```python
class Client:
    def mcp_call(self, tool, timeout=12, **kw):
        ...
        resp = self.call_tool(tool, kw)      # timeout bound, never forwarded
```

`timeout` bound to the named parameter and went nowhere — only `kw` was forwarded. `main()` wired
`call_tool` as `lambda name, args: client.mcp_call(name, **args)` onto
`MCPClient.mcp_call(self, tool, timeout=12, **arguments)`, so **every** read in the sweep ran on the
transport's own 12s default. `smartmoney.py` asks for `timeout=20` on both cohort reads
(`discovery_get_top_traders`, `discovery_get_trader_state`); none of it landed.

The MCP server's upstream budget is **20s** (axios, `SENPI_API_TIMEOUT_MS`). A 12s client budget
abandons every read that takes 12–20s *before the server can return anything* — and the server then
keeps working, and retrying, against the already-overloaded backend.

## Why it mattered this week

The ClickHouse cluster behind `common-hyperliquid-traders` degraded at 01:00 UTC 2026-09-18:

| read | latency |
|---|---|
| `GetTopPerpTraders` | p90 3.0s → **10.5s** |
| `GetTraderStates` | p95 **26.2s** |

12s sits mid-tail of that distribution, which is the ~50% of runs where the smart-money lens
(proven-cohort divergence + flow detectors) goes dark.

## The fix

**1. The budget reaches the socket, and it outlives the server.**

- `Client.mcp_call` forwards the caller's budget; the two `COHORT_TOOLS` reads are floored at
  `COHORT_TIMEOUT_S = 30`. The floor is needed because the engine only asks for 20s, which *equals*
  the server's own budget — a client that gives up at the same instant still always loses the race.
  30s = 20s server budget + the streamable-HTTP `initialize` round trip + the ~1000-row / 50-book
  transfer, at the round number the runtime scanner scaffold already uses for a per-call timeout.
- The market/leaderboard reads keep a short `READ_TIMEOUT_S = 15`; their intent is unchanged. Those
  calls sit behind the MCP's 10s leaderboard and Hyperliquid budgets, so 15s still leaves the client
  outliving the server it waits on.
- **The budget never enters the tool-arguments payload.** That dict is sent verbatim to the MCP
  server, so a `timeout` key in it would arrive as a tool argument. `_accepts_timeout()` asks
  `call_tool` by signature (`inspect.signature`, falling back to the 2-arg call when the callable
  can't be inspected); a `call_tool(name, args)` that knows nothing about budgets — the runtime's,
  and the tests' — is still called with two arguments. Signature introspection rather than a
  `TypeError` fallback: a `TypeError` raised *inside* the callee would otherwise be misread as
  "doesn't take a timeout" and silently re-issue the read.
- `main()`'s two duplicate lambdas are now one `_adapter(client)` that carries the budget through.
- **Considered and deliberately left alone:** the documented decision in `Client.mcp_call` that a
  `success: false` envelope is *not* retried there. It still returns as-is, so `_read` degrades that
  source exactly as before and the vendored engine keeps seeing the response shape it expects.

**2. `meta.cohorts_unavailable` no longer diagnoses itself.**

It asserted "discovery_* needs a USER-scoped `SENPI_AUTH_TOKEN`; an app-scoped token returns nothing
here" whenever both cohorts came back empty — *whatever the cause*. During this incident the cause
was an upstream timeout, and that sentence sent the investigation down a token-scope dead end for
hours. An empty successful read and a failed read are different facts:

- read failed → `no cohort data — the discovery_get_top_traders read failed, so the cohort was never
  read (this is not a token problem): top_traders page 0 failed: timed out`
- read succeeded, no rows → the token line, which is where it is actually true.

The failure is already recorded in `warnings`, so it is quoted rather than guessed at. `smartmoney.py`
is a byte-identical vendored copy in two skills; both are updated and
`senpi-signals/tests/test_vendored_parity.py` stays green. The smart-money SKILL.md text that told
the agent to blame the token was updated to match.

## Tests

- `test_the_budget_a_read_asks_for_reaches_the_socket` — records the timeout each tool's socket call
  receives through the real `main()` path; asserts the cohort reads get **> 20** and the market reads
  stay at 15. **Fails on `main`** (all three reads get 12).
- `test_a_call_tool_that_can_carry_a_budget_is_given_one` — **fails on `main`**.
- `test_the_budget_is_never_sent_to_the_server_as_a_tool_argument` — 2-arg `call_tool` back-compat
  plus the payload-purity guard.
- `test_a_failed_cohort_read_is_never_reported_as_a_token_problem` and the extended
  `test_cohorts_unavailable_flag_on_empty` — both **fail on `main`**.

`senpi-smart-money/tests` had never run in CI, so the engine half of a two-copy fix was ungated; it is
added to the `full suites` step. Every CI step verified green on a pristine copy of this branch:
1296 passed / 9 skipped in the full suites, plus vendor parity, golden minimums, account-status,
the 43 strategy-package directories and quant-desk.

Versions bumped (skill delivery gates on frontmatter version): `senpi-signals` 2.1.0 → 2.2.0,
`senpi-smart-money` 1.3.0 → 1.3.1, with the README version table. No catalog regeneration — these are
skills, not strategy packages, and `gen_catalog.py` reads only `strategies/*/strategy.yaml`.

## 3. The run has a deadline, and the cohort reads last *(added after review)*

The budgets above bound **one** read. Nothing bounded the run, and at 15s/30s that is up to 360s of
patience for a caller that has none: `--brief` is given ~60s (market-pulse runs it on every pulse)
and `--print-feed` ~120s. Unbounded patience on a degraded upstream does not return a thin feed — it
gets the process killed mid-read and returns nothing. That trade is the wrong way round.

- **A wall-clock deadline for the whole gather:** `BRIEF_DEADLINE_S = 45`, `FEED_DEADLINE_S = 100`,
  selected by the path `main()` is on. `gather`/`run` take it (and an injected `clock`) as keywords;
  `deadline_s=None` is the old unbounded behaviour.
- **A read is not STARTED unless what is left can pay for its own timeout.** `Client.can_start(budget)`
  projects `elapsed + budget` against the deadline. Checking elapsed time alone would let a 30s cohort
  read start at 44s of a 45s deadline and land at 74s — the exact overrun the deadline exists to
  prevent. The projection caps a run **at** the deadline, not at the deadline plus one hung read.
- **The four market/leaderboard reads now run before the cohort.** The cohort is one lens and the only
  one that can spend the whole budget alone (one ranking read plus a batch per 50 wallets, each the
  slowest call in the sweep). Reading it first is what made a slow upstream cost four cheap detectors
  to save one expensive one. Last, it is the lens that goes dark — and `score.py` renders everything
  that did get fed, with the existing not-measured line naming what did not.
- **A skipped read is never reported as a failed read.** Nothing was asked of the server, so there is
  no upstream fault to chase: `_read` writes `NO DATA: <tool> not started: 36s of the sweep's 45s
  deadline is spent and this read needs up to 15s`, and `cohort()` refuses the whole lens before its
  first read with `… Not a failed read and not a token problem.` Said in `sweep.py` rather than let
  through the vendored engine, which would have recorded it under `top_traders page 0 failed:` —
  `smartmoney.py` stays byte-identical in both copies.
- `current.json` carries `reads_skipped`; the `[sweep]` summary names the deadline when it bit.

Five tests, on an injected clock so the deadline is exercised without waiting for it: the projection
refuses an unaffordable read untried (`skipped == 1`, `failed == 0`); `deadline_s=None` stays
unbounded; a slow upstream yields the four market lenses plus a dark cohort and a rendered brief, with
`clock.t <= 45`; a skipped read's coverage prose is not a failure's; and `--brief` gets the shorter
deadline while `--print-feed` gets the longer one.

Scope note: this does **not** close the `HyperliquidClient.callInfo` gap in the MCP (3 × 10s ≈ 31s),
which can still outlive a 15s market-read budget. That needs the retry capped server-side, not the
client budget raised — it rides with Senpi-ai/senpi-hyperliquid-mcp#225's `totalBudgetMs` seam.

## This is not the cure

The root cause is the ClickHouse degradation behind `common-hyperliquid-traders`. This PR only buys
the sweep enough patience to collect what the server does eventually return; it does nothing about the
26.2s p95 itself, and a read slower than 30s still goes dark (honestly, and now with the real reason
quoted).

### Follow-up, not in this PR

`senpi-smart-money`'s own CLI path passes `timeout=20` straight to `MCPClient` — a client budget
exactly equal to the server's, so that path still always abandons first on a slow upstream. Fixing it
means editing the vendored engine's constants in both copies; called out here rather than bundled.


betashop:
## Verified the mechanism independently, and it is real

I read it on `main` rather than from the description:

```python
def mcp_call(self, tool, timeout=12, **kw):
    ...
    resp = self.call_tool(tool, kw)      # only kw is forwarded
```

`smartmoney.py` asks for `timeout=20` at `:144` and `:190` (both cohort reads) and 8/10 elsewhere. None of it reached the socket. Every read ran on the transport's 12s default, against a server with a 20s upstream budget — so the client abandoned reads the server was still working, in the exact 12–20s band where a degraded ClickHouse puts them.

Worth saying plainly: **I rewrote that method last night** to move the retry into it, and I did not notice the parameter was inert while I was standing in it. Good catch.

**`cohorts_unavailable` is the better half of this PR.** That string blamed token scope whatever the cause, and it cost real hours today — mine included. When I investigated this morning I had to explicitly disown my own 401 as evidence in the Slack thread, precisely because that sentence had already primed everyone toward a token dead end. Separating "read failed" from "read succeeded, no rows" is the same distinction this repo has now got wrong three times in two days, and quoting the recorded warning instead of guessing is right.

**Adding `senpi-smart-money/tests` to CI** — that suite had never run, and it owns the engine `senpi-signals` vendors. That is the kind of gap that lets a two-copy fix ship half-applied.

Verified on a clean checkout: vendored copies byte-identical, full suites **1296 passed / 9 skipped**, matching your number.

---

## One thing I would not merge without addressing

**The reads got more patient. The caller above them did not.**

| | worst case |
|---|---|
| before — 12s everywhere | **192s** |
| after — 15s market / 30s cohort | **360s** |

(4 market reads + 1 `top_traders` + 3 `trader_state` batches, × `READ_ATTEMPTS = 2`.)

And there is no wall-clock bound anywhere in `sweep.py` — I grepped for one.

Now the budgets the agent actually gives it, from today's telemetry:

```
--brief 3     timeout 60
--print-feed  timeout 120
```

`--brief 3` is the path **market-pulse runs on every single pulse**, and it gets 60 seconds. With your own measured `GetTraderStates` p95 of 26.2s, three sequential batches is ~78s before the market reads. That is over budget on the typical slow case, not the worst one.

So on a degraded upstream this trades a **thin feed** for **no feed**: today the sweep gives up early and prints "not measured this run" at exit 0; after this it keeps waiting and the agent kills it mid-flight. The user goes from a shorter answer to an error.

The fix is small and belongs with this change rather than after it: **an overall deadline that renders what it has.** Track elapsed from the start, stop *starting* new reads past ~45s on the brief path, and let `score.py` render the detectors that did get fed. The coverage line already exists to say what was missed — it just needs to be reached instead of being killed before it prints.

Sarvesh flagged the missing wall-clock bound on #642 and I left it as a follow-up because it was not load-bearing at 12s. At 30s it is.

Everything else here I would take as-is. Happy to write the deadline myself if you would rather keep moving on the ClickHouse side — say the word and I will push it onto this branch.

danielmbirochi:
Taken, and pushed onto this branch — `0e13f1c8`. You were right that it belongs here rather than after: the 12s ceiling was the only thing bounding the run, and this PR removed it.

## What landed

**A wall-clock deadline on the whole gather.** `BRIEF_DEADLINE_S = 45` / `FEED_DEADLINE_S = 100`, picked by the path `main()` is on, threaded through `run`/`gather` as a keyword (`deadline_s=None` is the old unbounded behaviour, which is what a caller with its own bound gets).

**A read is not started unless what is left can pay for its own timeout**, not merely "past 45s, stop". `Client.can_start(budget)` projects `elapsed + budget`. An elapsed-only check lets a 30s cohort read start at 44s and land at 74s — which is your 60s overrun with extra steps. The projection caps the run **at** the deadline instead of at the deadline plus one hung read, which is what makes the 45s number safe inside a 60s budget rather than merely smaller than it.

**The cohort reads last.** This is the half of your comment I'd flag as more than plumbing, so I want it seen: your own arithmetic says "~78s **before** the market reads". That ordering is what turns a slow upstream into a bad trade — the one lens that can spend the entire budget alone (1 ranking read + a batch per 50 wallets, each the slowest call in the sweep) was running before the four cheap ones. A deadline on top of that order would have cut the 4h board, the momentum events and the cross-asset flows to protect the cohort. Reversed, the cohort is the lens that goes dark, `score.py` renders the detectors that got fed, and `not_measured()` names the one that didn't. No dependency between them — `cohort()` only writes `smart_*` onto metrics; `momentum`/`cross_asset` read `price_change_pct`, which `board()` sets, so those three keep their relative order.

**A skipped read is not reported as a failed read.** That one is yours by precedent. Nothing was asked of the server, so there is no upstream fault to chase:

```
NO DATA: market_get_cross_asset_flows not started: 36s of the sweep's 45s deadline is spent
         and this read needs up to 15s
NO DATA: the proven-cohort read was not started — 32s of the sweep's 45s deadline is spent and
         a cohort read needs up to 30s. Not a failed read and not a token problem.
```

The cohort line is written in `sweep.py`, not let through the vendored engine — the engine would have recorded it as `top_traders page 0 failed:`, i.e. a third wrong cause on the line that has already cost you a morning. `smartmoney.py` stays byte-identical in both copies.

## Tests

Five, on an injected clock so the deadline is exercised without waiting for it: the projection refuses an unaffordable read untried (`skipped == 1`, `failed == 0`); `deadline_s=None` stays unbounded; an 8s-per-read upstream yields all four market lenses ok, a dark cohort, a rendered brief ending in `_Not measured this run: proven-trader positions._`, and `clock.t <= 45`; the skipped-read prose is not a failure's; and `--brief` gets 45 while `--print-feed` gets 100.

Local: `senpi-signals` + `senpi-smart-money` 79 passed, full CI suites green. `senpi-signals` 2.1.1 → **2.2.0** (behaviour, not a fix), README table and `references/detectors.md` (`reads_skipped`, the read order) updated with it.

## What it does not close

`HyperliquidClient.callInfo` in the MCP is still 3 × 10s ≈ 31s, which can outlive the 15s market-read budget — and it is the fallback that runs precisely when the traders read has already failed. That needs the retry capped server-side; the `totalBudgetMs` seam in Senpi-ai/senpi-hyperliquid-mcp#225 is where it goes. Not in this PR.

And the deadline is a bound, not a cure: at 45s a sweep against the current ClickHouse tail will lose the cohort lens on a fair share of runs. It loses it visibly and with a correct reason, inside a feed that still renders. The cure is upstream.

betashop (approved):
## Approved — the deadline does exactly what it needed to

I re-simulated it rather than reading the diff, using the `clock` seam you added (which is the right seam to have added). Virtual time, reads that each take a fixed cost:

| scenario | reads | elapsed | rendered | outcome |
|---|---|---|---|---|
| healthy, 1s reads, brief | 6 | **6s** | 1681 ch | full feed, nothing skipped |
| degraded, 9s reads, brief | 4 | **36s** | 1383 ch | thin feed, cohort dark, 1 skipped |
| degraded, 9s reads, full | 6 | 54s | 1681 ch | full feed |

The middle row is the whole objection answered: 36s against the agent's 60s budget, and it **renders** rather than being killed mid-flight. On the healthy path the deadline costs nothing, which matters as much — a guard that taxes the good case gets removed later.

Three things I want to note because they are better than what I asked for:

**`can_start` projects instead of checking elapsed.** `elapsed() + budget <= deadline_s` is the difference between a real cap and a suggestion — a 30s cohort read starting at 44s of a 45s deadline lands at 74s. Checking elapsed alone would have passed review and not worked.

**`SweepDeadline` is a distinct type, and `_read` maps it to `NO DATA: … not started` rather than `failed:`.** A read nobody made is not a read that failed, and the coverage line now says which. That is the same distinction that cost us hours this morning on `cohorts_unavailable`, applied before anyone had to learn it twice.

**`if last is not None: raise last`** — running out of time for the *retry* of a genuinely failed read re-raises the real failure instead of the deadline. The upstream fault stays the reported fact. Easy to get backwards.

Verified independently: full CI **1301 passed / 9 skipped**, vendored copies byte-identical, `senpi-smart-money/tests` in the `full suites` step.

## One thing before you merge — a version collision, not a code problem

This PR now claims `senpi-signals 2.2.0`. So does [#675](https://github.com/Senpi-ai/senpi-skills/pull/675), which was opened on that number earlier today. Skill delivery gates on the frontmatter version, so two different trees shipping as 2.2.0 is a real problem, not a cosmetic one.

**Keep 2.2.0 — yours lands first and it is the incident fix.** I will take 2.3.0 on #675 when I rebase onto this. Nothing for you to change; flagging it so you do not discover it in a conflict.

Ship it.

PR #675, opened 2026-09-18
senpi-signals 2.2.0: a proven wallet opening size, read from one sweep

betashop:
## Whale reads were v2 for a reason that does not apply to opens

Every whale read needed an earlier reading of ours to diff against — that is why `whale_move` sits in `HISTORY_DETECTORS` and why rule 5 banned the lot.

An **open** does not need one. `discovery_get_trader_state` returns each position's own age, so *"$12.4M short, opened 18 minutes ago"* is a fact about the position, not a comparison against something we stored. The engine simply never asked for it:

```python
resp = client.mcp_call("discovery_get_trader_state", trader_addresses=batch,
                       include_position_age=True, timeout=20)   # <- one parameter
```

Same call, same batch, **no extra latency and no state** — which matters given the timeout work in #674.

## What ships

A wallet from the ≥ $1M lifetime-realized cohort, holding ≥ $1M of notional, opened inside the last 4 hours. Named, sized, dated, with its lifetime gains alongside:

> A proven wallet `0x8f3a…21c4`, $37.2M in lifetime gains, opened $12.4M SHORT opened 18 minutes ago on HYPE — size, on a record, with a time on it.

## What does not — adds and flips stay v2

A bigger position tells you nothing without the old size; a flip needs the old side. Rule 5 now separates the three instead of banning them together, because a blanket ban would have the agent suppressing a detector the feed actually carries.

## Two rules the detector holds itself to

- **An undated position is dropped, never called fresh.** An undated whale position is almost always an old one, and "opened just now" is precisely the claim that must never be guessed.
- **An age is never rounded up into a bigger claim.** 59 minutes stays minutes.

`whale_open` is deliberately **not** in `HISTORY_DETECTORS`, and a test pins that — landing there would silently stop `--print-feed` carrying it, which is how this would quietly stop working.

## Test

10 new tests. The fixture holds a $12.4M short opened 18 minutes ago; three variants each prove a different reason to stay silent:

| variant | expected |
|---|---|
| $12.4M, 18 min | fires, reaches both feeds, carries the wallet |
| same, no `durationInSeconds` | **silent** — undated is not fresh |
| same, 5 hours old | **silent** — a holding |
| $500k, 18 min | **silent** — a position, not a statement |

```
senpi-signals + senpi-smart-money: 76 passed
full CI:                          1287 passed, 9 skipped
```

senpi-signals 2.1.0 → 2.2.0, senpi-smart-money 1.3.0 → 1.4.0 (vendored engine, byte-identical in both copies, parity test green), README rows with them.

## Conflicts with #674

Both branches change `sweep.py` and both copies of `smartmoney.py`, and both bump the same two versions. **#674 should land first** — it is a live-incident fix and mine is additive. I will rebase onto it and take its `2.1.1`/`1.3.1` as the base for `2.2.0`/`1.4.0`.

Worth noting the two interact in the right direction: #674 makes the cohort read patient enough to actually return, and this PR is the thing that most rewards it — a whale open only exists if `discovery_get_trader_state` came back.


shnoodles:
Reviewed against `origin/main` in a clean worktree at `ae6c0d8`. **Not approving yet** — one blocking question about `startTime`, plus a grammar bug in the flagship sentence. Everything else holds up, including the parts I expected to be wrong.

## Verified, and it's the right call

I went looking for the usual failure — a plausible-but-invented field name — and did not find one. The MCP schema confirms `include_position_age` exists and that it adds `startTime` (Unix **seconds**) and `durationInSeconds`; 137,060 production responses in the last 7 days confirm both, alongside `positionValue`. `szi` is genuinely signed on *open* positions (`"-0.0008"` vs `"37.4"`), so `"long" if szi > 0 else "short"` is right here — it is only *closed* rows where the sign is dropped and the side moves into `type`. Identifiers copied from the source, not from memory.

CI claim is honest: I reproduced **1287 passed, 9 skipped** on the exact `full suites` step. Vendored parity is byte-identical (`md5 7ee0ae78…` both copies). The skill-surface budget still passes with the longer rule 5.

## Blocking: `startTime` resets on positions that never moved

The detector's whole claim is "opened N minutes ago". On the only path with production history, that field resets on untouched positions.

One wallet's SOL position, 2026-09-18, two reads 14 seconds apart:

| time | `szi` | `durationInSeconds` |
|---|---|---|
| 13:57:58 | 0.41 | 4592 |
| 13:58:12 | 0.41 | **17** |

Same size, same side, no close, no flip — `startTime` jumped forward 4,590s and the position went from "76 minutes old" to "just now". Four more resets on that same wallet+coin in 12h (6165→11, 5389→17, 6270→11). Across the window, **19 of 170** (address, coin) pairs observed 3+ times show the signature: `szi` constant, `startTime` moved.

Two things make this land badly rather than academically:

- **The error is one-directional and lands exactly on your trigger.** A reset always makes an old position look newer, never older. `durationInSeconds` of `11` or `17` is indistinguishable from a genuine open, and `WHALE_OPEN_MAX_AGE_S` waves it straight through.
- **The tests cannot see it.** `_whale_tool` fakes the transport and `FRESH_WHALE` hand-sets `durationInSeconds: 1080`, so all 10 tests pass whatever the real field does.

To be fair to the field: it is *not* junk in general. The distribution is sane — 62.4% of positions at 4–24h, 25.2% at 1–7 days, max 17.3 days, only 8.8% under 4h. It is a real position age most of the time. The problem is the minority case, and the minority case is your entire output.

**What I could not determine, and why it's your call to close:** every one of those observations used `latest: true`. The sweep does not — it takes the cached path, and in 7 days that path has **24 calls against 137,060**. So the reset may be an artifact of the fresh-fetch re-derivation and may not touch you at all. I can't tell from telemetry, and I don't want to fail the PR on a mechanism I only saw on the other path.

Settle it with a real-API check, not a fixture: read the cohort twice ~20 minutes apart with `include_position_age=True` and no `latest`, and assert that for every position whose `szi` is unchanged, `startTime` is unchanged too. If it holds, this is cleared and I'll approve. Reproduction for the signature:

```sql
SELECT addr, coin, uniqExact(szi) AS sizes, uniqExact(startTime) AS starts
FROM (SELECT JSONExtractString(res,'data','trader','address') AS addr,
             JSONExtractString(p,'coin') AS coin, JSONExtractString(p,'szi') AS szi,
             JSONExtractInt(p,'startTime') AS startTime
      FROM (SELECT SpanAttributes['gen_ai.tool.call.result'] AS res FROM telemetry.otel_traces
            WHERE Timestamp >= now() - INTERVAL 12 HOUR
              AND SpanName LIKE '%discovery_get_trader_state%'
              AND SpanAttributes['gen_ai.tool.call.result'] LIKE '%durationInSeconds%')
      ARRAY JOIN JSONExtractArrayRaw(res,'data','trader','openPositions') AS p)
GROUP BY addr, coin HAVING sizes = 1 AND starts > 1
```

While you're in there: prefer `startTime` over `durationInSeconds`. On a cached read `durationInSeconds` was computed at cache-fill time and is stale by the cache age — in the same "looks fresher than it is" direction — whereas `startTime` is absolute and stays correct. It also deletes four speculative branches:

```python
start = _num(p.get("startTime"))
if not start:
    return None                       # undated: say nothing
age = time.time() - start
```

`duration_in_seconds` / `start_time` / the `< 1e12` millisecond guard are guarding against shapes this API does not emit — the schema documents seconds and 137k responses agree. (`startTime: 0` does occur; `not start` still catches it.)

## The flagship sentence says "opened" twice

```
A proven wallet 0x0000…0001, $5.0M in lifetime gains, opened $12.4M SHORT opened 18 minutes ago on BTC
A wallet (0x0000…0001) up $5.0M lifetime just opened $12.4M SHORT opened 18 minutes ago on BTC.
```

That's the real rendered output of your own fixture, in both feeds. `numbers` ends with `opened {_ago(age)}` and both `frame()` and `trade_read()` prepend "opened" again. It's in the PR description too, verbatim — which is how it got past review: the tests assert `"12.4M" in …` and `"18 minutes ago" in …`, substring checks that pass either way.

One word, in `score.py`:

```python
"numbers": [f"{_usd_short(notional)} {side.upper()} {_ago(age)}"],
```

I ran it — both renderers read correctly after, 65 signals tests still green.

## `_ago` rounds into the bigger claim at 2.5h

`test_an_age_is_never_rounded_up_into_a_bigger_claim` tests 30 / 1080 / 3540 / 3600 / 14400 — none of which exercise the rounding it is named after. `f"{s/3600:.0f}"` rounds to nearest:

```
 8999s (2.50h) -> '2 hours ago'      # 30 min understated → reads fresher
12600s (3.50h) -> '4 hours ago'      # rounds the other way
```

Understating age is the "bigger claim" direction for a freshness detector. `math.ceil(s / 3600)` always reports older, never fresher. Separately, `_ago(90)` returns `'1 minutes ago'`.

## Ponytail — diff only

Two small ones beyond the fallbacks above:

- `"opened": True` (`score.py:469`) is never read. The only consumer, `score.py:825`, is inside the `whale_move` branch, which the `whale_open` branch returns before reaching. Cut it.
- `WHALE_OPENS_PER_ASSET` / `WHALE_OPEN_SCALE_USD` are defined ~250 lines below their use. Works, but put them with the other module constants.

No reuse missed — I checked for an existing relative-time helper and there isn't one, so `_ago` earns its place.

## On the merge order

Agreed, and #674 first is right. One thing to carry through the rebase: #674 adds `senpi-smart-money/tests` to the `full suites` step. On this branch it is still absent (`grep -c` → 0), so your change to the vendored engine is ungated here — the same gap #674 closes. Make sure the rebase keeps that step rather than resolving it away.

Happy to re-review as soon as the cached-path check is in — the shape of this is good, and reading an open out of a call we already make is the right instinct.

betashop:
Every point taken. Pushed — full CI **1288 passed / 9 skipped**, vendored copies still byte-identical.

## The four I could fix

**"opened" twice.** You are right, and it is in my PR description verbatim, which is how it got past me. The tests asserted `"12.4M" in …` and `"18 minutes ago" in …` — substring checks that pass either way. That is the exact weak assertion I have spent this week calling out in other people's tests. There is now one that counts the word in both renderers.

**`_ago` rounded to nearest.** Also right, and the test was named for a rule it did not exercise — 30 / 1080 / 3540 / 3600 / 14400 contains no value that rounds. `math.ceil` now, so it can only ever report older, plus a sweep across the range asserting the spoken age is never less than the real one. `_ago(90)` said "1 minutes ago" too.

**`startTime` over `durationInSeconds`.** Took it, and for your reason rather than as a preference: a duration computed at response-build time is stale by the cache age on a cached read, and stale in the direction that makes an old position look newly opened. The four speculative branches are gone — `duration_in_seconds`, `start_time` and the `< 1e12` millisecond guard were me defending against shapes I had not checked the schema for. The fixture now sets a real `startTime` so the tests exercise the field the detector reads.

**Ponytail:** `"opened": True` cut, constants moved up.

## The blocking one

I cannot settle it from telemetry either, and I am not going to argue the mechanism away — your evidence is 19 of 170 pairs and the error lands precisely on the trigger.

`scripts/check_position_age.py` reads the cohort twice on the **cached** path the sweep actually uses, with the real engine and no `latest`, and asserts that every position whose `szi` is unchanged kept its `startTime`. Exits 1 with the offending pairs if any moved. It needs a box with a user token, so it is a script rather than a test — running it on a live agent now and will post the output here.

If it fails, this detector does not ship on `startTime`, and whale opens need either a backend fix to that field or the history we were trying to avoid. I would rather find that out from the check than from a user being told a three-day-old position opened eleven seconds ago.

## On the rebase

Noted on `senpi-smart-money/tests` — my change to the vendored engine is ungated on this branch, and #674 is what closes that. I will rebase onto #674 once it lands and make sure the `full suites` step keeps the added directory rather than resolving it away.

Thank you for checking the field names against the schema and 137k responses rather than taking them from the diff. That is the part I would not have caught myself.

shnoodles:
Checked `fdf676a7` and `63bd5893`. The four fixes are all good — `startTime` over `durationInSeconds` for the right reason, `math.ceil`, the doubled "opened" with a test that counts the word, `"opened": True` cut. CI reproduces at 1291/9.

On the entry-price corroboration: the mechanism is sound and deriving mark from `positionValue / |szi|` out of the same row is the right instinct — no extra call, no state. **But I ran it against the reset I documented and it passes.**

The same wallet's SOL position, across the boundary:

| time | `szi` | `durationInSeconds` | `entryPx` | mark | drift | allowed | guard |
|---|---|---|---|---|---|---|---|
| 13:57:58 | 0.41 | 4592 | 108.34 | 108.38 | 0.037% | 20.63% | passes |
| 13:58:12 | 0.41 | **17** | 108.34 | 108.36 | **0.019%** | 1.571% | **passes** |

`entryPx` does *not* get recomputed on a reset — it stayed at 108.34, which is what you'd want. The guard still waves it through, because the position happened to be sitting a fifth of a percent from its entry. Seventy-six minutes of age became seventeen seconds, and the second witness agreed.

That is exactly the case you called out — "an old position sitting near its entry still passes" — so this isn't a correction, it's the measurement. Worth knowing the size of it: the guard bites hardest at low claimed ages, which is where resets land, but it only catches resets on positions that have *drifted*. A whale holding near entry is both common and precisely the kind of position that sits untouched long enough to accumulate a reset.

So it lowers the rate, it doesn't close the hole, and `check_position_age.py` is still what decides this. Splitting it into `--snapshot` / `--compare` with the two-minute floor is the right shape — a compare inside one cache window would prove nothing.

Holding the approval for that output. If the cached path is stable, this ships and the entry-price check is a good belt-and-braces. If it isn't, the guard as tuned won't rescue it, and I'd rather we both know that before it's load-bearing.

Everything else here is ready to go.

betashop:
Merged as `3df1bc3b` on Jason's call — senpi-signals **2.3.0**, smart-money **1.4.0**, rebased onto #674 (which took 2.2.0). Full CI 1311 passed / 9 skipped, vendored copies byte-identical.

<a href="https://github.com/shnoodles">@shnoodles</a> — flagging that this went in ahead of your re-review, so you are not surprised by it landing.

**Everything you raised in code is in**: `startTime` over `durationInSeconds` with the speculative shapes dropped, `math.ceil` so an age can only ever read older, one "opened" in the sentence, the dead key and the constants.

**The blocking question is still open, and I have not pretended otherwise.** What went in instead is a second witness from the same read: the entry price. `positionValue` is |szi| x mark, so the mark falls out of the row, and a position claiming to be seventeen seconds old while sitting nine percent from its own entry is refused — the allowance scales with the claimed age (1.5% base for spread and a large entry, +0.25%/min). That is exactly the shape of the resets you found, which produce ages of 11-17s on positions that have been open long enough to drift.

It is corroboration, not proof: an old position sitting near its entry still passes. So `scripts/check_position_age.py` still matters and still needs running — it is two turns now (`--snapshot`, then `--compare` twenty minutes later), because a 1200s sleep inside one call is not something an agent can hold open in a chat session. If it comes back FAIL, this detector does not rest on `startTime` and I will take it out rather than leave it.

Your review changed the shape of this for the better — the entry-price cross-check exists because you would not take the field on trust, and I would have shipped it without one.

PR #679, opened 2026-09-19
ops 3.21.0, author 3.11.0: cwd-proof invocations, subset closes, a tick call budget

betashop:
Three findings from a day of M401059's own agent sessions (2026-09-18/19), each
verified against the code before being written up.

**1. The docs taught a cwd-dependent invocation, and it cost two closes.**
Asked to close two strategies at once, the agent wrote
`cd <skill> && python3 scripts/close.py <a> &` and then more `python3
scripts/close.py <b>` lines. The `cd` lived in the backgrounded subshell, so
every later line ran from the original cwd and died on `python3: can't open
file 'scripts/close.py': [Errno 2] No such file or directory`, repeatedly. The
strategies stayed open, and the error names a missing script rather than a lost
directory — so it reads as a broken tool, not a shell mistake.

The agent copied the form the docs use. All 13 script invocations in ops
SKILL.md were relative, in two different flavours (`senpi-strategy-ops/scripts/…`
from the skills root, `scripts/…` from inside the skill), neither of which
survives a lost cwd. They are now absolute — `~/.openclaw/skills/…`, the form
quant-desk/SKILL.md already uses — so the cwd cannot matter. README's two
invocations with it.

There is also no subset flag: `close.py <id>` takes one package and `--all` is
all-or-nothing, so closing three strategies is three runs. That is now stated
where close usage is read, with the backgrounding trap and the observed error in
references/lifecycle.md.

**2. No per-tick call budget anywhere in the author skill.** Two scanners
authored 5 minutes apart came in at 398 and 390 calls against a 240s budget and
both failed `E_VALIDATE_TICK_TIMEOUT`. The validator's message is good and
names the call count, but the skill never states the constraint up front, so the
agent found the ceiling by failing validation twice — each attempt a multi-minute
live run. Production enforces `timeout_seconds` exactly as validation does, so
this is not a slow test: an over-budget scan times out on EVERY tick and the
strategy never trades. The invariant now says so, with the fan-out patterns that
cause it (one call per asset over a wide universe) and the fixes (fetch once per
unique asset, hoist board-wide reads, batch wallet reads) in
references/creating-a-strategy.md.

**3. `validate_universe.py` was named bare in 4 of 6 mentions.** The agent
looked for it in senpi-strategy-author/scripts/ — where it is not — and got "No
such file". It lives in senpi-strategy-ops/scripts/. Every mention now carries
the path.

Both skill bodies were already exactly at their line budget (ops 335, author
425), so neither addition buys space with a budget bump: the mechanism went to
references/ and what stayed in the body was folded onto existing lines. Bodies
are 335/335 and 425/425 — unchanged.

Two things I checked and did NOT change. A `[senpi-runtime] register(api)
invoked` line broke a `JSON.parse` at 00:32, but that was the agent's own inline
`node -e` one-liner; the shipped CLI paths (_cli.py `_extract_json`,
portfolio.py) already carry the guard for stdout pollution, and the vendored
mcp_client.py parses an HTTP body, not stdout, so it is not exposed. And
read_senpi_guide looked like 3 failures out of 3 on a SERR keyword sweep — all
three succeeded; the guide's own text documents SERR codes.

Full CI: 1311 passed, 9 skipped, 66 subtests. Doc-only; no behaviour change.


shnoodles:
Reviewed at `9725ec4b`. Doc-only, CI **1299 passed / 9 skipped**, skill-surface budget still green. **Not approving — the cwd-proofing uses a path that does not exist on an agent box.**

## The routing rule is exactly right, and it is the real fix for the #682 incident

```
A STRATEGY IS A PACKAGE, NEVER A SKILL: "install/reinstall <name>" is
`deploy.py create <id>` here — never `openclaw skills install`, never clawhub, and
never a GitHub user or repo named after the strategy.
```

That names all three wrong turns the agent actually took on 2026-09-19, in order. And putting it in the frontmatter `description` rather than the body is the detail that makes it work — the description is what the agent sees when *choosing* a skill, so it lands before the ClawHub detour instead of after it. In that session the agent only recovered by reading this SKILL.md at 13:27:01, three minutes and a web search in. This is what stops that.

Same for the per-tick call budget: production enforcing `timeout_seconds` identically to validation means an over-budget scan times out *every* tick and never trades, which is the silent-forever failure, and it's now written down with the mechanism rather than as a limit.

## Blocking: `~` is `/root`, and the skills are not there

Every rewritten line becomes `python3 ~/.openclaw/skills/senpi-strategy-ops/scripts/deploy.py …`. On a real box:

```
$ echo $HOME
/root

$ cp -r /tmp/sq/quant-desk ~/.openclaw/skills/quant-desk
cp: cannot create directory '/root/.openclaw/skills/quant-desk': No such file or directory
```

The skills OpenClaw actually loads are at `/data/.openclaw/skills/` — `ls` there lists all fourteen senpi skills, and the working invocation from the 2026-09-19 session was:

```
cd /data/.openclaw/skills/senpi-strategy-ops && python3 scripts/deploy.py validate athena
```

So all 16 rewritten lines would fail with `python3: can't open file '/root/.openclaw/skills/…': [Errno 2]`. That is worse than what's on `main` today, because the current relative form at least works from the skills root, and it fails in a way that looks like a broken install rather than a bad path.

(`quant-desk/SKILL.md` already carries one `~/.openclaw/skills` line on `main` — same bug, pre-existing, worth fixing in passing.)

I'd rather not guess the canonical form for you since you know the runtime. The two that are evidenced to work are the absolute `/data/.openclaw/skills/…` and the `cd <skill dir> && python3 scripts/…` shape the agent already uses. If OpenClaw exposes the skills root as an env var, that beats both.

## The rewrite is incomplete regardless

The PR says all 13 invocations; 16 lines changed and **9 relative ones remain** in the same four files:

```
senpi-strategy-author/SKILL.md            python3 senpi-strategy-author/scripts/validate_strategy.py …
senpi-strategy-author/SKILL.md            python3 senpi-strategy-ops/scripts/validate_universe.py …
senpi-strategy-author/SKILL.md            python3 senpi-strategy-ops/scripts/deploy.py validate …
creating-a-strategy.md                    python3 senpi-strategy-author/scripts/validate_strategy.py …
creating-a-strategy.md                    python3 senpi-strategy-ops/scripts/deploy.py create …
lifecycle.md                              python3 scripts/deploy.py validate spider
lifecycle.md                              python3 scripts/deploy.py create spider --budget 300
```

(The two remaining `scripts/close.py <a> &` lines in `lifecycle.md` are the illustration of the failing shape, so those are correct as-is.)

Three of the misses are in `senpi-strategy-author/SKILL.md` — the skill this PR bumps to 3.11.0 for precisely this. Worth a grep gate in the skill-surface test so the next one can't drift: no `python3 ` followed by a non-absolute `scripts/` path outside a fenced "this is the broken shape" example.

## Small

`validate_universe.py` is qualified as `senpi-strategy-ops/scripts/validate_universe.py` in the author SKILL.md — correct, that one lives in ops, and it was genuinely ambiguous before.

Fix the path and I'll approve — the four findings are all real and the routing rule in particular should land today, since it's the one that closes the incident #682 was written for.

shnoodles:
Plan is right, and I checked that `/data/...` is safe to hardcode rather than just observed once. Over 7 days of agent `exec` calls touching `.openclaw/skills`:

| path | calls | distinct users |
|---|---|---|
| `/data/.openclaw/skills` | **4,278** | **287** |
| `~/.openclaw/skills` (unexpanded) | 4 | 3 |
| other | 5 | 3 |

287 users on one layout, and the only appearances of the tilde form are the handful we already know fail. So `/data/.openclaw/skills/...` is the evidenced path, not a guess — go ahead.

On the grep gate: worth pointing it at the bug that actually propagated, not just the relative-path class. The failure here was a **wrong absolute path copied between skills**, which a "must be absolute" rule would have waved straight through. Two assertions rather than one:

- no `python3 ` followed by a non-absolute `scripts/` path (the original ask), and
- no `~/` or `$HOME` in any documented invocation — since `$HOME` is `/root` and nothing ships there.

The second is the one that would have caught the quant-desk line before it became install instructions.

Re "I propagated it instead of catching it" — the reason it was hard to catch is that the working form was in a session trace and the broken form was in a sibling skill, and the sibling is the thing you naturally copy from. That is a repo problem, not an attention problem; a gate in the skill-surface test is the right fix and it is why the gate matters more than this particular correction.

Nothing else outstanding from my side. Push it and I will re-review — the routing rule is the piece I most want live, since it is what actually closes the 2026-09-19 incident.

shnoodles (approved):

PR #682, opened 2026-09-19
ops: reinstalling a closed strategy resolves the fork name to its template

betashop:
Taking Ignas's option 1 as reviewed — he built and measured it rather than
specifying it, so this is his patch with the announcement and tests added.

Three things the first cut got wrong, all found in review:

**It broke both callers.** `deploy.py:171` and `close.py:225` each rebuild
`<dest_root>/<id>` from the id they passed. Writing the template's own name meant
a download that fully succeeded was invisible to them: `deploy.py` reported
"fetched tree has no strategy.yaml" — false, and pointing away from the cause —
then swallowed it in the `except`. That is a worse diagnosis than not resolving at
all, and the carefully-written refusal never printed because nothing raised. The
files now land under the requested id and the substitution happens only in the
content.

**It guessed the split point.** `split("-", 1)[1]` assumes a username has no
hyphen, so `john-doe-athena` failed. It now matches the SUFFIX against the
templates the repo actually carries, longest first — which also handles `athena-x`,
the first hyphenated template id in the repo and one that did not exist yesterday.

**It substituted silently.** The fork is the user's Make It Yours copy; handing
back the pristine template under that name without a word means any tuning they
did is quietly not there. It now says so on stderr, and says nothing when the id
was an exact match.

The deeper miss was in the tests, not the code: every one of them called
`fetch_package` directly, so neither caller was ever in the loop and all eight
passed against a version that broke both. `test_fetch_through_callers.py` drives
`deploy._fetch_fresh` and mirrors close.py's reconstruction. Mutation-proven —
write the template's name instead of the caller's and it goes red.

Still true and unchanged: this is NOT the fix for the incident that prompted it.
The agent never reached `fetch_package` — it took a ClawHub route and 404'd there.
That fix is the routing rule in #679, which should land regardless.

MID removed from the body and the docstrings.

Ops suite: 434 passed, 1 skipped, 66 subtests. Full CI: 1551 passed, 9 skipped.


shnoodles:
Reviewed at `1c180f8d`, clean worktree, level with `main`. Ops suite reproduces exactly: **425 passed, 1 skipped, 66 subtests**.

**The fix is good. It is not the fix for the incident it cites.** I pulled the session rather than reasoning from the diff, and `fetch_package` was never reached.

## What the trace actually shows

The agent never treated `ignasteamsenpi-athena` as a strategy package at all — it treated it as a **ClawHub skill**:

```
13:21:03  rm -rf strategies/ignasteamsenpi-athena strategies/athena strategies/athena.bak-…
13:24:44  openclaw skills list | grep -i athena
13:25:01  openclaw skills search athena          → "athena-protocol" (someone else's)
13:25:20  openclaw skills install ignasteamsenpi-athena
          → ClawHub /api/v1/skills/ignasteamsenpi-athena failed (404): Skill not found
13:25:32  web_fetch clawhub.ai/skills/ignasteamsenpi-athena
13:26:06  web_search "site:github.com ignasteamsenpi athena"
13:26:19  web_fetch github.com/ignasteamsenpi          ← the GitHub-user hunt
```

The 404 that sent it to GitHub came from **ClawHub**, not from `_fetch.py`. There is no `fetch_package` call anywhere in the window — the first `deploy.py` invocation in the whole session is at 13:27:15, *after* recovery, and it uses the template id:

```
13:27:01  read /data/.openclaw/skills/senpi-strategy-ops/SKILL.md   ← the actual unstick
13:27:03  curl …/senpi-skills/main/strategies/catalog.json
13:27:15  deploy.py validate athena          ← template id, which already worked
13:33:12  deploy.py fork athena
13:33:47  deploy.py create /data/workspace/strategies/ignasteamsenpi-athena --budget 900
13:34:10  …--budget 897
```

So the improved refusal message in this PR would never have been printed: the code that raises it was never run. The agent didn't get a confusing error from us — it never asked us.

**The real mechanism is routing, not resolution.** Nothing connected `<username>-<template>` to "this is a Senpi strategy, `senpi-strategy-ops` owns it". The agent's default assumption for an unknown hyphenated package id is ClawHub, and it burned ~2 minutes and a web search on that before the user said "it's senpi strategy". The knowledge that fixed it already exists — it's in `senpi-strategy-ops/SKILL.md`, which the agent read at 13:27:01 and immediately got right. What's missing is anything that routes it there.

That's a one-line rule in the ops SKILL.md (or `AGENTS.md`), something like: *a package id shaped `<username>-<template>` is a Senpi strategy fork — resolve it through `senpi-strategy-ops`, never ClawHub or GitHub.* Cheaper than this diff and it addresses the observed failure.

Keep this PR — the latent bug is real and worth closing. Just don't close the ticket with it, because the next agent in that position takes the same ClawHub detour.

## Two gaps in the fallback itself

**Hyphenated usernames fail.** `split("-", 1)[1]` assumes the username has no hyphen:

```
ignasteamsenpi-athena  -> OK  athena
john-doe-athena        -> FetchError  (split gives 'doe-athena')
a-b-c-athena           -> FetchError  (split gives 'b-c-athena')
```

All 111 template ids are single-token, so you don't have to guess the split point — match the suffix against the tree instead, which handles any username and can't pick the wrong package:

```python
base = next((t for t in templates_in(tree) if strategy_id.endswith(f"-{t}")), None)
```

`zec-fade` still fails (no `strategies/fade/`), so the property your test pins is preserved.

**The template silently replaces the fork.** `dest = Path(dest_root) / fetched_id` writes `athena/` and returns it, and `test_a_fork_name_resolves_to_the_template` pins `assert not (tmp_path / "ignasteamsenpi-athena").exists()`. A caller that asked for the fork gets the pristine template with nothing saying so — and the fork is the *Make It Yours* copy, so any tuning the user did is quietly not there. In this incident it didn't bite (the agent ran `fork` explicitly straight after), but a fetch that substitutes a different package should say which one it returned. Returning `fetched_id` alongside `dest`, or a one-line note the caller can relay, is enough.

## Small

The PR body and both test docstrings carry `M408027` — `senpi-skills` is the public repo. We stripped MIDs from five comments here a couple of weeks ago for the same reason. The date and the symptom carry the story fine without it.

Happy to approve once the fallback stops guessing the split point; the routing rule can land separately, but it's the one that fixes the reported failure.

shnoodles:
Followed up on whether this is shippable for the latent bug, setting the incident question aside. **It isn't yet — the fallback doesn't survive its own callers.**

`deploy.py:171` reconstructs the path from the id it passed:

```python
_fetch.fetch_package(sid, tmp, ref=ref)      # contract: writes <dest_root>/<id>
if not (tmp / sid / "strategy.yaml").is_file():
    raise _fetch.FetchError("fetched tree has no strategy.yaml")
```

The fix writes to `<dest_root>/athena`, so that check looks for `<tmp>/ignasteamsenpi-athena/strategy.yaml`, misses, and the `except` swallows it. Run through the real function, stubbed tree, nothing else changed:

```
this branch   _fetch_fresh('ignasteamsenpi-athena') -> None
              log: remote fetch of 'ignasteamsenpi-athena' unavailable (fetched tree has no strategy.yaml)

origin/main   _fetch_fresh('ignasteamsenpi-athena') -> None
              log: remote fetch of 'ignasteamsenpi-athena' unavailable
                   (strategy 'ignasteamsenpi-athena' not found under strategies/ …)
```

Same outcome, worse diagnosis. The download *succeeded* — the files are sitting in `<tmp>/athena/` — but the operator is told the tree had no `strategy.yaml`, which is false and points away from the real cause. And the carefully-written new refusal naming the `<username>-<template>` convention never prints on this path, because `fetch_package` no longer raises here; `deploy.py` raises its own generic message instead. The one place the message was meant to help is the one place it can't reach.

`close.py:225` has it identically:

```python
_fetch.fetch_package(sid, dest_root, ref=a.ref)
pkg = _pkg.load(dest_root / sid)          # same reconstruction, same miss
```

The unit tests pass because they call `fetch_package` directly and assert `dest == tmp_path / "athena"` — neither caller is in the loop.

## It's not a one-liner

I tried the obvious repair — use the returned dir instead of rebuilding the path:

```python
got = _fetch.fetch_package(sid, tmp, ref=ref)
if not (got / "strategy.yaml").is_file():
    ...
return got
```

That makes it work end to end (`ignasteamsenpi-athena -> athena`, `zec-fade` still refuses, with your good message this time) — **and breaks 5 tests in `test_pkg.py`**, including `test_ensure_pkg_fetches_into_durable_root_regardless_of_cwd`. `<dest_root>/<id>` is a pinned contract, not an incidental detail.

So it's a real decision, not a patch:

- **Keep the contract** — write the template's files under the *requested* id, so `ignasteamsenpi-athena/` exists and both callers work untouched. Smallest blast radius. Costs you the property your test currently pins (`assert not (tmp_path / "ignasteamsenpi-athena").exists()`), which would need to flip.
- **Change the contract deliberately** — return the resolved dir, update `deploy.py` and `close.py` to use it, and update the 5 tests. More honest about what happened, more surface.

I'd take the first: the caller asked for the fork, the fork name is what's deployed and what `status`/`close` will look for later, and it keeps the substitution visible in one place instead of rippling.

Either way the fallback also still needs the suffix match rather than `split("-", 1)` — `john-doe-athena` fails today.

To answer the question directly: good idea, real bug, not yet shippable. The fork path fails through both callers and the diagnosis it leaves behind is less accurate than what's on `main` today.

shnoodles:
**Please take option 1 — keep the `<dest_root>/<requested id>` contract.** I built and ran it rather than just specifying it, so this is verified rather than proposed.

It writes the template's files under the id the caller asked for. Both callers keep working untouched, and it picks up the hyphenated-username case on the way.

```diff
-    fetched_id = strategy_id
+    src_prefix = prefix
     if not files and "-" in strategy_id:
         # (your comment, unchanged)
-        base = strategy_id.split("-", 1)[1]
-        base_prefix = f"strategies/{base}/"
-        base_files = [t["path"] for t in tree
-                      if t.get("type") == "blob" and t.get("path", "").startswith(base_prefix)]
-        if base_files:
-            files, fetched_id = base_files, base
+        # Match the SUFFIX against the templates the repo actually carries, rather than splitting
+        # on the first "-": a Senpi username may contain hyphens ("john-doe-athena"), and guessing
+        # the split point gets those wrong. Longest match first so a template cannot be shadowed by
+        # a shorter one that happens to be a suffix of it.
+        templates = {q.split("/", 2)[1] for q in
+                     (t.get("path", "") for t in tree if t.get("type") == "blob")
+                     if q.startswith("strategies/") and q.count("/") >= 2}
+        base = next((t for t in sorted(templates, key=len, reverse=True)
+                     if strategy_id.endswith(f"-{t}")), None)
+        if base:
+            src_prefix = f"strategies/{base}/"
+            files = [t["path"] for t in tree
+                     if t.get("type") == "blob" and t.get("path", "").startswith(src_prefix)]
@@
-    dest = Path(dest_root) / fetched_id
+    # Always the id the CALLER asked for: deploy.py and close.py both rebuild this path from the id
+    # they passed, so writing anywhere else means the package they just downloaded is invisible.
+    dest = Path(dest_root) / strategy_id
     for path in files:
         status, content = _get("raw.githubusercontent.com", f"/{repo}/{ref}/{path}", "*/*", timeout)
         if status != 200:
             raise FetchError(f"raw fetch HTTP {status} for {path}")
-        out = _out_path(dest_root, path)
+        out = _out_path(dest_root, prefix + path[len(src_prefix):])
```

The last line is the one that does the real work — remapping the write path keeps `_out_path`'s traversal guard in force, since the prefix is ours and only the relative tail comes from the tree.

**Measured:**

```
fetch_package
  athena                 -> athena                 ['athena/strategy.yaml', 'athena/phalanx/runtime.yaml']
  ignasteamsenpi-athena  -> ignasteamsenpi-athena  ['ignasteamsenpi-athena/strategy.yaml', …]
  john-doe-athena        -> john-doe-athena        ['john-doe-athena/strategy.yaml', …]
  zec-fade               -> refused

deploy.py._fetch_fresh
  athena                 -> athena
  ignasteamsenpi-athena  -> ignasteamsenpi-athena
  john-doe-athena        -> john-doe-athena
  zec-fade               -> None
```

Ops suite: **424 passed, 1 skipped, 66 subtests** — the only failure is your own `test_a_fork_name_resolves_to_the_template`, on exactly the two lines that pin the other choice:

```python
assert dest == tmp_path / "athena"
assert not (tmp_path / "ignasteamsenpi-athena").exists()   # never the fork name
```

Those flip to the fork name. Worth keeping the test and renaming it to what it now guarantees — something like `test_a_fork_name_fetches_the_template_under_the_fork_name` — because the property that matters is *the caller gets a package at the path it asked for, built from the template*. The five `test_pkg.py` failures you'd hit going the other way don't occur here.

Two things I'd still like in the same push:

- **Say which template it resolved to.** Right now the substitution is silent, and the fork is the *Make It Yours* copy — a user reinstalling it gets the pristine template with no indication their tuning isn't there. A returned `(dest, base)` or one line the caller can relay is enough; `deploy.py` already has a `log` in hand.
- **Drop `M408027`** from the PR body and the two test docstrings — public repo. The date and the symptom carry it fine.

Separately from this PR, the routing rule is still what actually fixes the reported incident, since the agent never reached this code. Happy for that to land on its own.

With option 1 in, I'll approve.

shnoodles:
Good — nothing to add on the patch itself; it is measured as posted and the ops suite lands at 424/1/66 with only your own fork test needing its two assertions flipped.

**One correction to something I said, because it would send you after the wrong thing.** I wrote that `athena-x` makes `split("-", 1)` "wrong in a way it wasn't yesterday". That is not right, and I should have checked it before saying it:

```
ignasteamsenpi-athena-x → split("-", 1)[1] = "athena-x"   ✓ still resolves
john-doe-athena-x       → split("-", 1)[1] = "doe-athena-x"  ✗
```

A hyphenated *template* is fine under the split; it is still only a hyphenated *username* that breaks it, which was already true before athena-x existed. So the urgency I implied isn't there.

What athena-x does change is narrower and still worth having: the template set is no longer all single-token, so any code that assumes a template id has no hyphen is now wrong, and the suffix match stays correct as more of them land. That is a reason to prefer it, not a new break.

athena-x merged at 14:14 today, so forks of it can exist from now on — but under either implementation a simple username resolves fine, so there is no race to land this before someone forks it. Take the time to get the contract change right.

The caller breakage is the part I would keep as the lesson: the tests passing while both callers were broken is the same shape as the card that said "same gates" while a gate was gone. Worth a test that drives `deploy.py._fetch_fresh` end to end with a fork name, so the contract is pinned at the boundary that actually consumes it rather than one layer in.

shnoodles (approved):

PR #683, opened 2026-09-19
athena-x 1.0.0: Athena at conviction size, as a package instead of a hand-edit

betashop:
On 2026-09-19 Jason edited the live Athena in place to make it "bigger" — 5x
instead of 3x, 25%/20% margin per slot instead of 15%/12%, one more slot per
sleeve. It worked, and the reason it worked is worth separating from the reason
it appeared to: the redeploy also reseeded phalanx's baseline, which is what
actually restored entries (fixed properly in #681). Size and selection are two
different decisions and the evidence only speaks to one of them.

So this is the sizing decision made explicit, as a package a user can choose
rather than an edit they perform on a live book. Selection is Athena's, unchanged:
the scanners are byte-identical to the Phalanx and Aegis source templates, and a
test pins that runtime.yaml may differ ONLY in identity lines and an enumerated
list of sizing keys. A logic change cannot enter this package without failing CI.

~3.7x the gross exposure of Athena for identical entries (phalanx 135% -> 500%
notional, aegis 144% -> 500%).

Two deliberate departures from the live run it came from:

- **The daily entry cap stays.** That run removed `max_entries_per_day` — but it
  predates phalanx 1.4.0, which can now enter on a STANDING consensus rather than
  only a growing one. Uncapped entries at 5x with that path open is how a book
  fills in one burst; a scanner on this same box did exactly that in 138 seconds
  the night before. Raise it on evidence, not to match the original run.
- **`min_budget_floor: 100`.** Computed mechanically, Athena-X's minimum comes out
  LOWER than Athena's ($40 vs $100) because a bigger marginPct clears the $10
  exchange floor on a smaller wallet. True and useless: it would advertise the
  higher-risk package as the cheaper one.

Wallet env vars are namespaced (`ATHENA_X_*`) so both packages can be installed at
once without binding to the same wallet — a test covers that too.

`risk_level: aggressive`, `tier: advanced`, and the catalog copy says plainly that
this is the wrong first strategy: start on Athena, move here on evidence.

8 tests. Catalog regenerated in both locations (110 strategies) and the min_budget
golden updated. Full CI: 1315 passed, 9 skipped, 66 subtests, 46 strategy dirs.


shnoodles:
Reviewed at `509b5fe4` in a clean worktree. **Not approving — the PR body and the code disagree about the daily entry cap, and the body has the better argument.**

## Verified and genuinely well built

- **Scanners byte-identical** — all four files, `md5` matched against `strategies/athena/*/scanners/`. The central claim of the package holds.
- **Catalog regenerated properly.** I re-ran `gen_catalog.py` and got no diff; both copies are in sync.
- **CI:** 1303 passed, 9 skipped, 66 subtests.
- **`min_budget` floored to 100** — correct and a good catch. Advertising the 5x package at $40 while Athena sits at $100 would have put the riskier one on the shelf as the cheaper one.
- **The wallet-separation test** is the right paranoia for two packages that can be installed together.
- **Exposure:** measured 3.58x combined (phalanx 3.70x, aegis 3.47x), so "~3.7x" is the phalanx figure rather than the package's. Worth saying 3.6x, or saying "up to 3.7x on the alpha sleeve".

Worth stating plainly somewhere: at full slots **each sleeve commits 100% of its wallet as margin** (4 × 25%, 5 × 20%). Athena sits at 45% and 48%. That is the real change in character — not the leverage number, the absence of headroom.

## Blocking: the cap is gone, and the PR says it stays

PR body:

> **The daily entry cap stays.** That run removed `max_entries_per_day` — but it predates phalanx 1.4.0, which can now enter on a STANDING consensus rather than only a growing one. Uncapped entries at 5x with that path open is how a book fills in one burst; a scanner on this same box did exactly that in 138 seconds the night before.

Both runtimes:

```diff
-    max_entries_per_day: 12          # phalanx (aegis: 8)
-    bypass_max_entries_per_day_on_profit: false
+    # NO max_entries_per_day. Removed deliberately, matching the live "bigger" run…
```

and a test locks it in:

```python
def test_the_daily_entry_cap_is_deliberately_absent():
    assert "max_entries_per_day" not in rails, leg
```

So the body argues the case for keeping it, names the mechanism, cites an incident from the night before — and the code does the opposite and pins it. One of the two is a leftover, and which one is a product decision I can't make for you. But note the body's reasoning is load-bearing and still true: **#681 merged overnight**, so the standing-consensus path is live on `main` right now. That change exists because a growing-only gate produced 359 consecutive empty ticks; it necessarily fires far more often. Removing the entry cap on top of it, at 5x, with no margin headroom, is three permissiveness changes stacked in one release.

If the removal is what you want, the body needs rewriting and I'd want the 138-second burst addressed explicitly rather than left as an argument against the merged behaviour.

## The card tells users the gates are unchanged

`strategy.yaml` `thesis`, which `gen_catalog.py` puts in front of a user through discover:

> Every signal, gate and exit is Athena's, unchanged

And Athena's own card, bumped to 1.4.1 in this same PR:

> the identical strategy — same signals, same gates, same exits

`max_entries_per_day` lives in `risk.guard_rails`. It is a gate, and it was removed. Whatever you decide above, these two sentences cannot ship as they are — a user comparing the two packages is being told the only difference is size.

## `default_signal_validity_seconds` is not a sizing key

`600 → 1800` is enumerated in `SIZING_KEYS`, so the parity test waves it through. Tripling how long a signal stays actionable is a behavioural change: a cohort read can now fire half an hour after it was taken. That is exactly the "logic change entering through the sizing door" the test exists to prevent, and it is the third permissiveness change in the stack.

Either move it out of `SIZING_KEYS` and leave it at 600, or keep it and stop describing the diff as sizing-only.

## Smaller

- **`%%` reaches the user.** `strategy.yaml:18-19` — `25%%`, `15%%`, `20%%`, `12%%`, `5%%` — and it flows straight into the catalog card. Looks like a printf escape that never got consumed.
- **Version mismatch.** PR title and description say 1.0.0; `strategy.yaml` says `1.1.0`.
- **The fee advisory should be on the card.** `validate_strategy.py` says ≈1.50% of budget per day, ≈45%/month, and explicitly "say the number to the user before they fund it". With the cap removed the entries/day it assumes is a floor, not a ceiling. The card discloses drawdown scaling but not fee drag.
- **`min_budget: 100.0`** in the catalog vs `100` for Athena and `100` in the golden fixture. Cosmetic, but it is a generated field so the float is coming from somewhere.
- **I could not run `validate_universe.py`** — local SSL failure reaching the live instrument list, not a defect in this PR. Given `assets: []` and byte-identical scanners that derive the universe, I expect it is clean, but I have not verified it and someone should before merge.

## One cross-PR note

`athena-x` is the **first template id in the repo with a hyphen** — I checked all 111. That makes #682's `strategy_id.split("-", 1)[1]` wrong in a way it was not yesterday: a fork `<user>-athena-x` splits correctly, but the suffix-match version I asked for there handles it under any username. Worth landing #682's fix before this, or at least not after a user has forked `athena-x`.

Happy to re-review quickly once the cap question is settled — the packaging work itself is good, and making sizing an explicit choice instead of a hand-edit is the right call.

PR #686, opened 2026-09-19
athena-x 1.4.0, athena 1.5.1: make the cards true

betashop:
Ignas's review of #683 caught that the catalog card and the code disagree. He was
right, and by the time #685 landed the card was wrong in more places than he
listed. This is what a user browsing discover was being told:

- `25%%`, `15%%`, `20%%`, `12%%`, `5%%` — printf escapes I never consumed, shipped
  straight into `belief_plain` and `thesis` and out through gen_catalog.
- "Every signal, gate and exit is Athena's, unchanged" — false. The daily entry
  cap was removed in 1.1.0 and the daily loss halt in #685. Both live in
  `risk.guard_rails`. They are gates.
- "The daily loss limit stays at 5%, which at this size trips sooner" — describes
  a gate that no longer exists, and warns about the opposite of the real risk.
- "a single bad regime day reaches the 5% daily loss limit quickly" — same.
- "~3.7x the gross exposure" — 3.70x is the phalanx sleeve alone. Weighted 65/35
  the package is 3.62x.

Athena's own card carried the mirror of it: "`athena-x` is the identical strategy
— same signals, same gates, same exits". A user comparing the two was told the
only difference is size while two entry gates had been removed.

Both cards now say what is actually true, including the thing neither said and
Ignas flagged as the real change in character: **at full slots each Athena-X
sleeve commits 100% of its wallet as margin**, against 45% and 48% on Athena. The
absence of headroom matters more than the leverage number.

`default_signal_validity_seconds` reverted 1800 -> 600 and removed from
`SIZING_KEYS`. He is right that it is not a sizing key: tripling how long a signal
stays actionable lets a cohort read fire half an hour after it was taken, which is
exactly the logic change the parity test exists to keep out — and I had put it on
the allow-list myself, so the test waved through the one thing it was written to
catch.

Catalog regenerated (110). Full CI: 1315 passed, 9 skipped, 66 subtests.

Still open from that review and not in this PR: the fee advisory is not on the
card, `min_budget` serialises as `100.0` where Athena's is `100`, and nobody has
run `validate_universe.py` against athena-x.


shnoodles:
Checked the merged cards on `main`. The rewrite is a real improvement — `%%` gone, the stale "daily loss limit stays at 5%" gone, and the honest sentences are in:

> At full slots each sleeve commits 100% of its wallet as margin, against 45% and 48% on Athena; that absence of headroom is the real change in character, not the leverage number.

> Two entry gates are also gone: there is no daily loss halt and no cap on entries per day, so the book can fill in one burst and can keep adding while it is underwater.

> Nothing flattens the book.

That last one is the sentence I was going to ask for and did not have to. A user can now read the card and know that the 37% figure halts entries and closes nothing.

I also checked the drawdown arithmetic rather than take it:

```
phalanx   athena 10/1.35x = 7.41% move   athena-x 37/5.00x = 7.40% move
aegis     athena 10/1.44x = 6.94% move   athena-x 35/5.00x = 7.00% move
```

Same market move, as claimed. And `default_signal_validity_seconds` is back to 600 in both sleeves.

**One thing did not land: the tagline.**

```
tagline       'the same entries' … 'Roughly 3.7x'
belief_plain  'about 3.6x'
```

The tagline is the highest-visibility field — it is what discover shows first — and it still says **3.7x** where `belief_plain` says 3.6x, so the card now disagrees with itself. It also still says *"the same two sleeves, the same signals, **the same entries**"*, which `belief_plain` immediately contradicts with *"loosens two gates that ration entries"*. The entries are the one thing that are **not** the same any more.

`test_the_cards_do_not_promise_what_the_code_does_not_do` passes, so whatever it checks does not cover the tagline or cross-field consistency. Cheapest gate: assert every `3.\dx` figure in a card agrees with the computed `slots × margin_pct × leverage` ratio, and that the tagline does not claim sameness the body withdraws.

Also still `min_budget: 100.0` (float) where Athena has `100` and the golden fixture has `100` — cosmetic, but it is a generated field so something upstream is producing the float.

Everything else here is good, and the test suite you replaced it with is better than what I reviewed — pinning stop width and profit-lock rungs in PRICE terms rather than parameter equality is the right invariant for a sizing fork, and `test_the_entry_rationing_gates_are_gone` makes the removals deliberate rather than incidental.

PR #688, opened 2026-09-19
min_budget: an authored floor keeps the field's type

betashop:
Ignas, reviewing #683: `min_budget` serialises as `100.0` for athena-x where
Athena's is `100` and the golden fixture has `100`. Cosmetic, but it is a
generated field, so the float was coming from somewhere.

`_f()` returns a float, so a package that RAISES its minimum with an authored
`catalog.min_budget_floor` kept that float all the way into the catalog, while
every computed minimum is an int. Two rows that mean the same thing rendered with
different types, and only for packages that author a floor — which is a small
enough set that it read as a one-off rather than a rule.

Whole values round-trip to int now. Nothing about the VALUE changes: the floor
still only ever raises, and is still never shrunk by the rounding.

`min_budget.py` is vendored byte-identically in senpi-trading-runtime and
senpi-strategy-ops, so both copies move together — the vendor-parity test caught
me editing one.

Catalog regenerated (110): athena-x 100.0 -> 100, nothing else moved.
Golden unchanged, vendor parity OK. Runtime + ops suites: 1022 passed.


shnoodles (approved):
**Approved.** This is the float I flagged on #686, fixed at the source rather than in the catalog.

Verified at `76a06f4`:
- The two `min_budget.py` copies are still byte-identical (`md5` matched) — the fix went into both, which is the part that is easy to half-do.
- Re-ran `gen_catalog.py`: no diff, so the committed catalog matches what the generator now produces.
- No `float` `min_budget` rows remain across all 110 skills.
- 453 passed, 1 skipped.

Right call putting it in `strategy_min_budget` rather than normalising in the generator — `_f()` returning a float is the actual cause, and every consumer of the authored floor now gets the same type as a computed one.

One gap: nothing pins it. A reintroduced `float()` passes the suite. Cheapest guard, wherever the golden minimums are already asserted:

```python
assert all(isinstance(s["min_budget"], int) for s in catalog["skills"])
```

Not blocking — the golden fixture will catch a value change, just not a type change, and this bug was only ever a type change.

PR #689, opened 2026-09-19
chore: no real user ids in a public repo, and a guard so they stay out

betashop:
85 real Senpi user ids were committed here. 70 of them were one copy-pasted
comment, duplicated across 70 template runtime.yaml files as templates were forked
from one another — and that one paired a real first NAME with a real id inside a
single parenthetical, which is the part that matters. An id alone is opaque to an
outsider; an id next to a name is not.

Ignas asked me to strip a single id from one of my own PRs for exactly this
reason, and noted this had already been cleaned once before. So the cleanup is
only half the job — the guard is the other half.

**The findings are kept; only the who is removed.** These comments are what make a
rule evidence rather than opinion: "found live on this date, on this asset, and it
cost this much". Every rewrite keeps the date, the asset, the strategy and the
measurement:

    (Ignas, M<id>, xyz:PLTR on cougar)     -> (observed live on xyz:PLTR in cougar)
    Observed live on M<id> on 2026-09-18   -> Observed live on 2026-09-18
    M<id>'s catalog `dog` discarded 151    -> one live catalog `dog` discarded 151
    found live on M<id> (2026-08-13)       -> found live (2026-08-13)

Synthetic fixture ids are untouched and allow-listed — M123456 and the M0000xx
values in research_fixture.json are unmistakably not real.

**Nothing behavioural moved.** Only comments, docstrings and markdown prose: the
phalanx and aegis scanners are still byte-identical across athena and athena-x
(the drift tests pass), and regenerating the catalog produces no diff at all — so
no version bumps. A privacy fix that churned 70 template versions would have cost
the whole fleet a pointless upgrade for a comment.

The guard is repo-wide and runs in CI: no `M\d{6}` outside the synthetic allow-list,
plus a separate assertion for an id written next to a name — that one stays even
once the first is green, because the pairing is the worse failure. Mutation-proven:
restore the comment in any one of the 70 files and the suite goes red.

Full CI: 1555 passed, 9 skipped, 66 subtests, 46 strategy dirs.


shnoodles (approved):
**Approved** — the scrub is complete and the guard works. Three gaps below, none of which makes this worse than `main`, so they are follow-ups rather than blockers.

Verified at `ce36504`:
- **Working tree is clean.** Swept the whole repo for `M\d{6}`: only `M123456`, `M000001`, `M000009`, `M111111`, `M999999` remain, all unmistakably synthetic.
- **Nothing behavioural moved.** phalanx and aegis scanners still byte-identical across athena/athena-x, and `gen_catalog.py` reproduces the committed catalog with no diff — so the no-version-bump claim holds. Deciding not to churn 70 template versions for a comment was the right call.
- **The guard bites.** Reintroduced an id into `strategies/weaver/main/runtime.yaml` → `test_no_real_user_id_is_committed` FAILED. Removed it → green.
- **CI 1555 passed, 9 skipped, 66 subtests.** Matches exactly.

Keeping the findings and dropping only the who is the right instinct — `(observed live on xyz:PLTR in cougar)` is still evidence, and a rule with a date and a measurement behind it survives the next editor. And separating the name-pairing assertion from the id assertion is correct: the 70-file comment was worse than 70 bare ids.

## The ids are still in public history

This is the part the description does not cover. I picked one id this PR removes and looked for it on `origin/main`:

```
commits on origin/main containing it: 1
```

The repo is public, so `git log -S<id> -p` still returns it to anyone, and the name-pairing comment is recoverable the same way for as long as those commits exist. Scrubbing the tip is necessary and not sufficient. That is a separate decision — history rewrite on a public repo is disruptive and not mine to call — but it should be an explicit decision rather than an assumption that this PR closed it. Worth raising with whoever owns the repo, and worth a line in this PR saying it is knowingly out of scope.

(7 distinct people across the 85 occurrences, for whoever makes that call.)

## Two blind spots in the guard, both demonstrated

`.github` is in `SKIP_DIR`, so the guard does not scan it:

```
printf 'note M408027 here\n' > .github/PROBE.md   → 2 passed   ✗
```

And `SCAN_EXT` is seven extensions, so anything else is invisible:

```
printf 'M408027\n' > strategies/probe.jsonl       → 2 passed   ✗
```

Neither is hypothetical for this repo — `.github` holds workflow files that get edited by hand, and `.jsonl` / extensionless files are exactly where a pasted telemetry sample would land. Dropping `.github` from `SKIP_DIR` costs nothing, and inverting the extension rule (scan everything that decodes as UTF-8, skip the binaries you already skip) removes the whole class rather than the two cases I happened to try.

Given the stated history here — cleaned once, came back, 85 occurrences by the second time — the guard is the durable half of this PR, so it is worth closing those two before the third round.

PR #691, opened 2026-09-19
signals-hunter 1.0.0: the Senpi Signals feed, traded on a clock

betashop:
Jason asked to template the strategy he authored on his box overnight. Four
decisions were his: vendor the engine with a drift test, keep conviction sizing,
trade both directions, and skip a tick when the gather degrades.

**Why it had produced zero ticks.** I expected the gather to be too big. It is
not. The box version makes the SAME reads with the SAME limits as
senpi-signals/scripts/sweep.py — universe top-120, one cohort page, board 500,
momentum 50 — and the sweep runs that in about a minute. The difference is that
the sweep budgets its reads (15s each, 2 attempts, 100s for the whole gather) and
the hand-written scanner had a bare `call_tool` with neither. One hung read
blocked the tick forever: runtime alive, position tracker ticking every 10s, scan
silent for 2h49m with no error and no timeout line. Vendoring the engine fixes
that by construction — `timeout_seconds` here is 240s rather than 900s precisely
because the gather now returns on its own terms.

**The engine is vendored, not reimplemented.** sweep.py, score.py and
smartmoney.py are byte-identical to the skill's and a drift test enforces it. The
detector library IS the strategy; a hand-written approximation of it silently
stops being the thing whose signals the user was reading. A second test asserts
scan.py never redefines `detect_from_metrics`, `trade_score`, `credibility`,
`coverage` or `rank` — that is exactly how the box version diverged.

**What the scanner owns**, because a feed does not have to care:

- *History.* Most of the library diffs against an earlier reading — OI surges,
  conviction builds, funding flips, whale moves. A feed gets that from a state
  file; this keeps a 24-deep hourly ring in `ctx.state`, which covers the trend
  detector's ~12h slow arm. The first tick after a deploy can only fire the
  point-in-time detectors and SAYS so rather than looking quiet.
- *A degraded gather is not a quiet market.* If the proven-cohort lens goes dark
  the score is OI/funding/price — a different number under the same name. The
  tick is skipped and the log names the missing lens.
- *A direction it cannot name is not a LONG.* The box version defaulted unscored
  direction to LONG. This skips it and counts them in the WAITING line.

**Sizing is conviction, on Jason's call.** 5 slots x 20% (25% above score 80) at
5x = the whole wallet committed when full, 5.00x notional. I recommended the
conservative profile for an unproven strategy and he chose conviction; the card
says so in plain words rather than burying it. Every DSL threshold is scaled for
5x so the stops sit where they would at 3x in PRICE (25 ROE = 5.00% of price),
and the drawdown budget is 37% = a 7.4% adverse market move — divided by the
NOTIONAL multiple, not leverage. No daily loss halt, consistent with the other
conviction packages.

`min_budget` floored to 100: the mechanical minimum is $15, which puts each of
five slots under the $10 exchange floor before the book is half full.

15 tests — 2 parity, 13 behavioural. Three mutation-proven: default an unnamed
direction to LONG, trade through a dark cohort, or flatten the sizing tiers, and
the suite goes red. `sub_style: signal_feed` registered in the glossary.

Catalog regenerated in both locations (111 strategies).
Full CI: 1557 passed, 9 skipped, 66 subtests, 47 strategy dirs.

NOT yet proven live: this has never completed a tick anywhere. It needs one clean
hourly tick on a funded box before it should be recommended to anyone.


shnoodles:
Reviewed at `b25ef2d`, clean worktree, level with `main`. **Not approving** — one money-risk that is new to this package, and one mechanical failure the repo's own validator already predicts.

Crediting the good parts first, because they are the hard parts: the engine really is byte-identical (I `md5`'d all three against `senpi-signals/scripts/`), the vendored `sweep.py` carries #674's `COHORT_TIMEOUT_S = 30` / `_accepts_timeout` fix, `gatherDeadlineSeconds: 100` inside `timeout_seconds: 240` leaves honest headroom, the catalog regenerates with no diff, there is no `@senpi/runtime`, and 13 tests pass. Diagnosing the 2h49m silent tick as an unbudgeted `call_tool` rather than an oversized gather is right, and vendoring to fix it by construction is the correct instinct. So is saying plainly that this has never completed a tick anywhere.

## `whale_open` can now open a position

The vendored `score.py` includes `whale_open`, and `scan.py` has no detector allow-list — `detect_from_metrics` returns every detector and the only filters are the credibility floor, a nameable direction, not-already-held, the dedupe TTL and `minScore`.

I scored one, using the same `$12.4M / 18 minutes` shape as the signals tests:

```
credibility 1.0   trade_score 75.1
clears minScore 65?        True
reaches highScore 80 tier? False  -> margin 20% at 5x
on a $100 budget: $20.00 margin -> $100.00 notional, one slot of five
```

`whale_open` carries `NON_OBVIOUS 1.0` and `EDGE 0.85`, which is top-tier, so it ranks well and will be picked.

The problem is what that detector rests on. On #675 I showed `startTime` jumping forward on a position whose `szi` and `entryPx` had not changed — 4592s of age becoming 17s — and then showed that the entry-price corroboration added in `63bd5893` **passes that exact reset**, because the position was sitting 0.019% from its entry. `check_position_age.py` was written to settle whether the cached path resets at all, and as far as I can see it has never reported. #675 merged on Jason's call with that explicitly open.

That was a defensible trade for a feed: the cost of a false "a whale just opened $12.4M" is a misleading sentence a user can discount. This package turns the same signal into a 5x entry sized at a fifth of the wallet, on an hourly clock, with no human between the detector and the order. The failure mode is specific rather than theoretical — an old position reads as newly opened, and the direction it reports is the direction we take.

Two ways out, both cheap:

- **Exclude `whale_open` from the tradeable set** until the cached-path check reports, and say so in one line where the other lens decisions are made. Every other detector in the library is a statement about a level or a change we measured ourselves; this is the only one whose freshness is asserted by a third-party field we have caught being wrong.
- **Or run `check_position_age.py`** and land this behind it. It is two turns and it answers the question for both packages at once.

I would not block the package on this — just that one detector.

## The free-margin gap will bite every full tick

`validate_strategy.py` flags it and I think it is right:

> a 5-slot scanner with no free-margin gate — every signal in one tick is sized off the same balance read, so the second and later opens land `position_open_failed` (insufficient margin), tick after tick

With `marginPct: 20` × 5 slots you are at exactly 100% of the wallet with zero headroom; with `highMarginPct: 25` on any signal scoring ≥80 the intended total goes **over** 100%. `topN: 6` allows six emits against five slots. So a tick that emits well is the tick that fails, and it fails on the later — lower-ranked but still qualifying — signals, which is a silent quality bias, not just noise in the log.

The fix the advisory names is the right one: read `withdrawable` in `scan.py` and emit only what it funds. That also makes the `familyCap: 2` and `topN: 6` numbers mean what they look like.

## `validate_strategy.py` hard-fails this package, and the validator is the thing that is wrong

```
✗ signals-hunter:
    - instance main: missing sibling main/scanners/scoring.py ('import scoring' will fail)
```

Nothing imports `scoring` — I checked, and `import scan` succeeds. The check at `validate_strategy.py:328` requires `scoring.py` unconditionally, while `CLAUDE.md` states it is optional:

> `scoring.py`  optional pure math (no I/O) so the edge unit-tests

signals-hunter is the **only** package in the repo without one, which is why this has never surfaced. `deploy.py` does not call `validate_strategy.py`, so it will not block a deploy — but it is the preflight `senpi-strategy-author/SKILL.md` tells an author to run, and an agent that runs it gets a hard ✗ on a package that is structurally fine.

Cheapest correct fix is the validator: make the `scoring.py` error conditional on something actually importing it. Adding an empty `scoring.py` to silence it would be the wrong way round — it would encode a requirement the architecture doc explicitly denies.

## Smaller

- The fee advisory (≈0.50%/day, ≈15%/month at 5 entries/day) is worth a line on the card, same as it was for athena-x. The card is honest about sizing; this is the other number a user needs before funding.
- `min_budget` floored to 100 against a mechanical $15 is correct and well reasoned — five slots under the $10 exchange floor before the book is half full is exactly the trap.

Land the free-margin gate and park `whale_open`, and I will approve — the vendoring architecture here is good and I would rather see it running than reworked.

betashop:
All four done in `bfe9e7b`. CI green.

**`whale_open` parked.** Your framing settled it: every other detector states a level or a change we measured ourselves — this is the only one whose freshness is asserted by a field we have caught being wrong, and the feed/trader asymmetry is the whole point. A discountable sentence becomes a 5x entry in the direction the field got wrong, hourly, with nobody in between. It sits in a `NON_TRADEABLE_DETECTORS` set with the reason written where the lens decisions live, and it returns when `check_position_age.py` reports rather than on anyone's judgement.

**Free-margin gate landed**, and writing the test found the second half of it. The drop count only appeared in the WAITING line — so a tick that opened four and silently dropped two said *nothing*, which is precisely the case worth seeing. It now reports on every tick it happens, naming the committed percentage and the cap. `_withdrawable()` is a `max()` across the two dex views rather than a sum, and a failed read is treated as unknown rather than as an empty wallet.

**Validator fixed rather than worked around.** You were right that adding an empty `scoring.py` would encode a requirement `CLAUDE.md` denies. The check now asks whether anything actually imports `scoring` — five tests, including that a mention in a comment or docstring is not an import. signals-hunter validates clean now, and the free-margin advisory is gone because the gate is real rather than suppressed.

**Fee drag on the card** — ≈0.50%/day, ≈15%/month at five entries a day, plus the parked detector and the funded-emit cap. The card was honest about sizing and silent about the number someone needs before funding.

18 tests, up from 13. Both new guards mutation-proven: make `whale_open` tradeable again, or remove the cap, and the suite goes red.

On `check_position_age.py` — agreed it answers this for both packages at once, and it needs Jason's box and token, so it is queued rather than blocked on me. Parking the detector is the version that does not wait.

PR #699, opened 2026-09-20
deposit 1.5.0: check for money already in a strategy before asking for a deposit

betashop:
## What the user hit

A user with money already on the platform was told to **deposit more before they could deploy**, and shown the funding card twice. Their balance was sitting inside an idle strategy wallet the whole time. It took eight minutes and a sceptical question from the user before the agent looked in the right place and reclaimed it in a single call.

Three users in the last 48h show the same shape; one of them showed it eleven times in one session.

## Root cause

`account_get_portfolio` can report `total_allocated_in_strategy: 0` while an `ACTIVE` strategy holds the money — and count that same money again under `total_withdrawable`. Read alone, the portfolio says "funded and free" when the money is neither, so the agent concludes the funding wallet is empty and reaches for the funding card.

Nothing in this skill told it to look in a strategy. The skill is precise about perps vs Spot vs EVM, which is where money usually hides — a strategy wallet was simply not on the list.

## What changes

`senpi-deposit-withdraw-transfer` 1.4.0 → 1.5.0, two paragraphs:

- **Deposit / add funds** now begins by ruling out money the user already has: call `strategy_list` before the card; an `ACTIVE` strategy with `netFunded` above zero holds funds `strategy_withdraw_funds` reclaims in one call. The tell is `total_withdrawable` well above `total_in_hyperliquid`.
- The top-up precheck's "nothing covers it" branch points at the same check before it falls through to the card.

No rail changes: money still enters only through the funding card, and the two iron rules are untouched.

## The backend half

The portfolio field itself is wrong and is filed separately for backend review — this PR does not depend on that being fixed. The skill's job is to not strand a funded user even when a balance field lies to it, and the reconcile is one cheap read.

## Verification

The exact CI step covering this skill:

```
python3 -m pytest senpi-strategy-ops/tests … senpi-deposit-withdraw-transfer/tests strategies/tests -q
```

**1565 passed, 9 skipped, 66 subtests passed, exit 0.** Includes the existing `test_existing_rails_survive`, `test_readme_row_matches_the_skill_version` (README row bumped in the same commit) and `test_public_repo_hygiene`, plus a new test pinning the rule.

## Detection rule

Portfolio reads where `total_allocated_in_strategy` is 0 while the user has an `ACTIVE` strategy with `netFunded > 0` — the same query that found these three. Worth an F-rule in the funnel report so the next case pages instead of surfacing as a user complaint.

## Ponytail

Two paragraphs in the skill that already owns this decision, no new section, no new tool, no engine. Rejected: teaching every skill to distrust the portfolio, and a wrapper that reconciles balances — the fix belongs at the one point where the agent decides to ask for money.

Owner: Jason · Reviewer: Ignas


shnoodles:
**Review at `8b8b9ef6` — the incident is real, but the root cause is misread, and the new money-path sentence would do harm. Not approving.**

### Required

1. **The portfolio isn't lying — the fields mean what MCP documents.** `total_allocated_in_strategy` is capital *used as margin* in strategies, and `total_withdrawable` is *"idle capital INSIDE strategy wallets — not in positions, recoverable with strategy_withdraw_funds"* (MCP `src/resources/account-get-portfolio.resource.ts:31-32`). An idle strategy wallet therefore reads `allocated 0, withdrawable X`, which is the field working. Nothing is counted twice: every read I checked sums exactly, e.g. M417755 at 09-20 01:18:56Z `0.624848 + 0 + 67.641263 = 68.266111 = total_balance_usd`, and M417499 at 03:05Z `0.666 + 57.306 + 94.601 = 152.573`. The new sentence (*"count that same money again as `total_withdrawable` — so the portfolio alone reads 'funded and free'"*) teaches the agent to distrust a correct field. Withdraw the backend ticket too. **The tell is simply `total_withdrawable > 0`** in the portfolio the agent already read; no "well above `total_in_hyperliquid`" heuristic needed.

2. **`netFunded` is the wrong field.** It's the funding ledger (`totalFunded − totalWithdrawn`), not what the wallet holds. It stays > 0 on essentially every funded strategy, including ones trading right now, and doesn't shrink with losses. So *"an ACTIVE strategy with `netFunded` above zero holds funds"* fires for every user with a running strategy, and overstates what a losing one holds. MCP's withdraw guide says to read **`withdrawable`** from `strategy_list`, *"the max you can pull"* (`strategy-withdraw-funds.resource.ts`).

3. **"Reclaims in one call" can strand a runtime.** Per the same guide, a withdrawal that drains the wallet to 0 **auto-closes** the strategy (`CLOSING_POSITIONS`). On a runtime-managed strategy that's a raw close by another route, and CLAUDE.md is explicit that teardown is `close.py <id>` because a raw close strands the runtime. The guide also says *"If the user did not state an amount, ASK — never default to the withdrawable balance."*

4. **Version:** main is already 1.5.0 (#744). Rebase and take 1.6.0 (you flagged this).

### The case this rule misses
In the one incident I could reconstruct, **M417755, 09-20 04:09Z**, the agent said *"Deposit $10 (or more) through that card"* while **$67.64** sat in strategy `5d703e7f`. That strategy had been in `CLOSING_POSITIONS` since 23:58Z, and the agent's own withdraw got SERR045 at 00:03Z. It wasn't ACTIVE, so "ACTIVE with `netFunded > 0`" never fires. `total_withdrawable` read 67.64 the whole time, and the right answer was *"it's on its way back — no deposit needed."*

### Suggested replacement (same length)
> **First check they don't already have it.** `total_withdrawable` in the portfolio is money idle inside strategy wallets — not free, not in positions. Above zero, `strategy_list` says where (`withdrawable` per strategy, never `netFunded`) and what to say: `CLOSING_POSITIONS` → it returns on its own; a strategy the runtime manages → `close.py <id>` brings it back; otherwise `strategy_withdraw_funds` for an amount they name (a full drain auto-closes it). Money moves only on their yes.

The top-up "nothing covers it" pointer can stay as-is once it points at this.

### Ponytail (diff only)
Small, and in the right place. The one simplification is item 1: the check needs no new read, since `account_get_portfolio` already carries the answer. CI is green on the head, but the rebase is needed regardless.

shnoodles:
**Post-merge review: this merged at 20:57Z with no review, and all three blocking items from my review above are live on main in `227df766`.** The merged paragraph is the reviewed text unchanged.

1. **The root cause is false.** Per MCP (`account-get-portfolio.resource.ts:31-32`), `total_withdrawable` *is* the idle money inside strategy wallets, and nothing is counted twice: every read sums exactly to `total_balance_usd`. The skill now teaches the agent to distrust a correct field.
2. **`netFunded` is the funding ledger, not the balance.** It's > 0 on nearly every funded strategy, including ones trading right now. MCP's withdraw guide reads `withdrawable` from `strategy_list`.
3. **"Reclaims in one call"**: a full drain **auto-closes** the strategy (`CLOSING_POSITIONS`), which strands a runtime-managed strategy. And the guide says to ask for an amount, never default to it.

Also missing: the deposit-side `$11.50` lines that #731 moved here are still at `SKILL.md:240,242`.

This needs a fix-forward. The replacement paragraph in my review above drops in as-is, and `total_withdrawable > 0` also catches the M417755 case (a strategy stuck in `CLOSING_POSITIONS`) that this text misses.

